Issue #10, opened 2021-11-24
 Improve Elm installation on Gitpod [8h]


AronNovak:
Issue

Gitpod is great for reviewing PRs, and for folks interested in eheza to experiment with it, without having to install anything locally.
When spinning a Gitpod, Elm packages are not installed properly all the time, as documented in the README.

Task

- [x] Try to determine what is the root cause for the download errors
- [x]  Install elm (npm install elm@latest-0.19.1) and elm-format globally. To save time maybe we can do it in parallel to the rest of the installation?
- [x]  When Elm installation is done open the preview browser on Gitpod

AronNovak:
>  When Elm installation is done open the preview browser on Gitpod

Just contacted Gitpod support if we can open two previews.
Only the last opened preview remains visible.

PR #15, opened 2021-11-25
Improve GitPod integration

AronNovak:
#10

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,6 +6,8 @@ image:
 tasks:
   - name: Terminal
     init: |
+      npm install -g elm-format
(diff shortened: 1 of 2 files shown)


AronNovak:
> Thanks for your email! 
> 
> To see another preview, I think you’ll have to start the server in a new terminal. I hope that makes sense!
> 

That's what the support said about opening two previews. Right now the Elm app is visible after launch.
@amitaibu Can we leave it like that then?
The other points were addressed.

amitaibu:
> Can we leave it like that then?

:+1:

amitaibu (approved):

PR #20, opened 2021-11-29
Nutrition rework

AronNovak:


AronNovak:
@adamhstewart confirmed that the logic is correct, he is using this for reports already.

server/hedley/modules/custom/hedley_person/hedley_person.module
```diff
@@ -1026,6 +1043,7 @@ function hedley_person_create_basic_node($type, $author_id = FALSE) {
   // Mark node as imported, to be able to identify it at hook_insert().
   $node->imported = 1;
   $node = node_submit($node);
+  node_save($node);
```
anvmn: @AronNovak 
Saving node is not a job of this function.
Throughout the code, whenever node is created, wrapper is used to populate node data.
Node is saved using the wrapper. 
This functionality prevents problems with `node_insert / node_update` hooks, that expect to see data on node.

Please revert this change (and try-catch clauses in PR), or, lets discuss...

anvmn (changes requested):

server/hedley/modules/custom/hedley_patient/hedley_patient.module
```diff
@@ -41,18 +41,10 @@ function hedley_patient_node_update($node) {
     return;
   }
 
-  $wrapper_orig = entity_metadata_wrapper('node', $node->original);
```
anvmn: @AronNovak

Whty was this changed?

server/hedley/modules/custom/hedley_person/hedley_person.module
AronNovak: I took a closer look, as I didn't remember the root cause, and that's why:
image
It was just a result of bad move(s) in Git, I revert everything what's not related to reporting.

anvmn: @AronNovak 

I see. In this case, please make sure that this PR contains only intended changes, then MWG.

PR #39, opened 2021-12-14
Update Drupal Core version to v7.84

anvmn:


anvmn:
Trivial change. Review not required. Merging...

PR #40, opened 2021-12-14
DB import process

AronNovak:


AronNovak:
image

.ddev/config.local.yaml.example
```diff
@@ -7,6 +7,14 @@ hooks:
         > 0 AND mail LIKE '%~%'"
     # This will be applied to production with a future release.
     - exec: drush sqlq "DELETE FROM system WHERE name='hedley_pusher'"
+    - exec: drush updb -y
+    - exec: drush updb -y
```
anvmn, line 11: @AronNovak really need this twice?

anvmn (approved):

.ddev/config.local.yaml.example
```diff
@@ -7,6 +7,14 @@ hooks:
...
+    # Modules can come and go, production might be behind.
+    - exec: wget https://ftp.drupal.org/files/projects/registry_rebuild-7.x-2.5.tar.gz
+    - exec: tar -xvf registry_rebuild-7.x-2.5.tar.gz --directory ~/.drush
+    - exec: rm registry_rebuild-7.x-2.5.tar.gz
+    - exec: drush rr
```
adamhstewart, line 17: for me this line failed when it is run before the cache clear.

adamhstewart: @AronNovak ^^

.ddev/config.local.yaml.example
AronNovak: Yep, not sure why, but locally I saw that after one `updb`, there was still one item to process.

.ddev/config.local.yaml.example
AronNovak: right, adjusted

PR #42, opened 2021-12-15
Change the date format

Diama1:


client/src/elm/Utils/NominalDate.elm
```diff
@@ -129,7 +129,7 @@ renderDate language date =
     )
         ++ " "
         ++ month
-        ++ ", "
+        ++ " "
```
anvmn, line 132: No need for all these ++. Lets make it `++ " month " ++`.

anvmn (changes requested):

client/src/elm/Utils/NominalDate.elm
anvmn: @Diama1 ^^

Diama1: > No need for all these ++. Lets make it `++ " month " ++`.

@anvmn but month is not a string, it is a variable

anvmn: @Diama1 right, my mistake.
Sorry.

anvmn (approved):

PR #45, opened 2021-12-16
Change from low risk Covid-19 to simple Covid-19

Diama1:


anvmn (changes requested):
@Diama1 

When I search for Low Risk I see 3 occurances:

Selection_417

anvmn (approved):

PR #48, opened 2021-12-16
Naming cleanup

AronNovak:

PR #106, opened 2022-01-20
Update Client's backendUrl

amitaibu:


anvmn (commented):
Actually, this is part of #48. Closing.

Issue #127, opened 2022-02-02
Attendance page stuck on spinner when all group participants have graduated [4h]


anvmn:
![Peek 2022-02-02 20-09](https://user-images.githubusercontent.com/17882245/152212531-2d5837f2-dde8-4114-91ef-63937c0c5951.gif)

anvmn:
After digging into this, found that root cause is at logic for generating data for editable session.
When we have participants (even when all of them have graduated), we expect the data to have at least one person loaded. However, in this case, there're none.

anvmn:
@adamhstewart 
Haven't heard of this problem for 6 months.  I think we can close this issue.

PR #129, opened 2022-02-02
Resolve a usecase when Attendance page gets stuck on spinner

anvmn:
#127


anvmn:
Peek 2022-02-02 20-13

anvmn:
The changes here are trivial. so, I'll go ahead and merge without a review.

PR #166, opened 2022-02-21
Add NULL values to count of HC encounters

adamhstewart:
Prior to May 2021 health center encounters are NULL


server/hedley/modules/custom/hedley_admin/scripts/generate-anc-report.php
```diff
@@ -54,8 +54,8 @@
        field_data_field_prenatal_encounter_type fdfpet ON p.entity_id = fdfpet.entity_id
    WHERE
      p.bundle = 'prenatal_encounter' AND
-     fdfpet.field_prenatal_encounter_type_value='nurse' AND
-     date(edd.field_expected_date_concluded_value) > DATE_ADD(CURDATE(), INTERVAL 30 DAY)
+     (fdfpet.field_prenatal_encounter_type_value='nurse' OR fdfpet.field_prenatal_encounter_type_value is NULL) AND
+     date(edd.field_expected_date_concluded_value) > DATE_ADD('2021-05-30', INTERVAL 30 DAY)
```
adamhstewart: needs to be reverted.

adamhstewart:
https://github.com/Gizra/ihangane/issues/2946

adamhstewart:
@AronNovak could you check the work here, but also give a hand with the code formatting - I ran phpcbf and it didn't help

What we did:
- Counted NULL values for the ANC report
- Provided a limit date as an argument from the robo command (default to today)
- Also provide a limit date for the pregnancy report.

AronNovak:
@adamhstewart I restructured the code to be more clean and it has the side-effect of passing phpcs.
I started to implement `ddev phpcs`, but I don't want to waste more time on it, if you think it worth +1h, I will continue the effort.

adamhstewart:
> if you think it worth +1h, I will continue the effort.

Thanks @AronNovak - let's do that.

adamhstewart:
@AronNovak this restructured code removes the interval limits  so that all the queries yield the same results. See for instance:

image

adamhstewart:
@AronNovak added back the proper limits for active and completed preganancies.

AronNovak:
@adamhstewart 
I know it's not the most serious issue, but I started with an easy task, local `ddev phpcs` works like on other projects:
image

anvmn:
Replaced by #197. Closing...

PR #170, opened 2022-02-24
Add new queries for different types of ANC encounters.

adamhstewart:


AronNovak (approved):

adamhstewart:
@AronNovak I had forgotten that you already reviewed this so I pushed some new work here instead of creating a new PR - sorry for the extra headache.

Here's what's new in the latest commits:
- On the encounters report: ANC CHW totals are correct (we were considering NULL values as CHW when they should have been Health Center)
- On the registration report: Added a limit date to be able to check registration from a moment in time (like ANC report)
- Added a TOTAL row in each of the registration tables.

I have checked these at various dates on the existing metrics table and they all check out (or at least close enough as historical databases weren't take on exactly midnight)

could you give this another review?

anvmn (approved):

PR #179, opened 2022-03-07
Hotfix: Loosen decoding of LMP confirmation field

anvmn:
As per https://github.com/Gizra/ihangane/issues/2949


anvmn:
The change is trivial. Will MWG.

Issue #140, opened 2022-02-07
ANC Logic - Diagnosis and Next Steps: Hypertension [12h]


jeanette-pelizzon:
## Issue

Hypertension is a disease that is diagnosed through monitoring blood pressure. There are two types of hypertension: 

- Gestational = under 20 weeks EGA
- Chronic = over 20 weeks EGA 

The EGA is the only thing that affects whether it is gestational or chronic.

In this flow, the nurse will be able to check **one** of three options for treatment. 

There are two flows for hypertension. 

**Return for Vital Check Flow**
Initial Diagnosis | Next Steps | Case Mgmt   
--|--|--
<img width="272" alt="Screen Shot 2022-02-07 at 4 13 01 PM" src="https://user-images.githubusercontent.com/77505464/152872947-ae6bac77-595d-4e35-8414-6a031722fc85.png"> | <img width="268" alt="Screen Shot 2022-02-07 at 4 13 15 PM" src="https://user-images.githubusercontent.com/77505464/152872971-899496c3-021a-4c2d-9bfa-d2a4dd458a39.png"> | <img width="377" alt="Screen Shot 2022-02-07 at 4 14 55 PM" src="https://user-images.githubusercontent.com/77505464/152873062-34b779f4-93b0-44ba-a697-cab69a39b2de.png">  

Vitals Recheck | Final Diagnosis | Next Steps
--|--|--
<img width="269" alt="Screen Shot 2022-02-07 at 4 13 09 PM" src="https://user-images.githubusercontent.com/77505464/152873283-e9f8cea2-c1fc-4c51-8caa-6ac8709583a7.png"> | <img width="268" alt="Screen Shot 2022-02-07 at 4 13 23 PM" src="https://user-images.githubusercontent.com/77505464/152873141-c5499204-28e3-436a-b4bb-78ed798456e3.png"> |<img width="282" alt="Screen Shot 2022-03-01 at 5 12 33 PM" src="https://user-images.githubusercontent.com/77505464/156257830-a71221bf-94ad-47cd-a80a-2d7e9cfbab92.png">

**No Vital Check Needed**
Diagnosis | Next Steps
--|-- 
[...]

PR #185, opened 2022-03-15
ANC Logic - diagnosis Hypertension

anvmn:
#140


anvmn:
Selection_581

anvmn:
Peek 2022-03-17 09-11

anvmn:
Selection_582

anvmn:
### Vitals form on for recheck:

Selection_584

anvmn:
@mariano-dagostino please review

client/src/elm/Pages/Prenatal/Utils.elm
```diff
@@ -852,19 +877,66 @@ recommendedTreatmentMeasurementTaken allowedSigns measurements =
...
+recommendedTreatmentSignsForHypertension =
+    [ TreatmentMethyldopa2
+    , TreatmentMethyldopa3
+    , TreatmentMethyldopa4
+    ]
+
+
+suspectedHypertensionCondition : Float -> Float -> Bool
+suspectedHypertensionCondition dia sys =
+    (dia >= 90 && dia < 110) || (sys >= 140 && sys < 160)
```
mariano-dagostino, line 942: @anvmn I think is worth to check if we can relax checks here... Specially `< 110` and `< 160 ` which can be potentially false in case higher values.

mariano-dagostino (commented):
@anvmn Looks good. Just one question regarding the logic

nedSaf (approved):

client/src/elm/Pages/Prenatal/Utils.elm
anvmn: @mariano-dagostino 
Indeed, this was changed at #188.
Function renamed to `marginalBloodPressureCondition`, and when using it, we first check if ther's high BP, and if not, if there's marginal BP:
https://github.com/TIP-Global-Health/eheza-app/pull/188/files#diff-9970c5f239b556adc0f4db1546965e4f3f953817d13f7ed917150ace15714051R1153

Issue #189, opened 2022-03-21
General - Add all the missing Translations from different Modules


Diama1:
## Issue

Some translations have been reviewed and some of them needed to be rephrased and some others are missing translations.

## Tasks

- Add the missing translations and rephrase the ones that needed to be rephrased

anvmn:
@Diama1 Translations were deployed to Demo env.
Please request APP upgrade for all devices synced with Demo env.

PR #195, opened 2022-03-22
Add missing translations from different modules

Diama1:
#189

### DIFF
--- a/client/src/elm/Measurement/View.elm
+++ b/client/src/elm/Measurement/View.elm
@@ -1425,8 +1425,8 @@ viewVitalsForm language currentDate config form =
                 (Maybe.map toFloat form.heartRate)
                 (config.setIntInputMsg heartRateUpdateFunc)
                 "heart-rate"
-                Translate.BpmUnitLabel
-            , Pages.Utils.viewPreviousMeasurement language config.heartRatePreviousValue Translate.BpmUnitLabel
+                Translate.HeartBpmUnitLabel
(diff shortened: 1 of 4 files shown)


client/src/elm/Measurement/View.elm
```diff
@@ -1425,8 +1425,8 @@ viewVitalsForm language currentDate config form =
                 (Maybe.map toFloat form.heartRate)
                 (config.setIntInputMsg heartRateUpdateFunc)
                 "heart-rate"
-                Translate.BpmUnitLabel
-            , Pages.Utils.viewPreviousMeasurement language config.heartRatePreviousValue Translate.BpmUnitLabel
+                Translate.HeartBpmUnitLabel
```
anvmn: Lets rename `BpmUnitLabel` to `BreathsPerMinuteUnitLabel` and use it for respiratory rate.
For heart, lest use `BeatsPerMinuteUnitLabel`.

anvmn (changes requested):

client/src/elm/Translate.elm
```diff
@@ -593,7 +593,7 @@ type TranslationId
     | HealthEducationProvidedQuestion
     | HealthInsuranceQuestion
     | Heart
-    | HeartBpmUnitLabel
+    | BeatsPerMinuteUnitLabel
```
anvmn: @Diama1 
Note that we do our best to maintain alphabetcal order here, and when you rename something, you need to move it to proper place. I'll do this myself, so we can deploy the PR.

Issue #201, opened 2022-03-28
Assemble the Activities of a Subsequent Visit [12h]


adamhstewart:
## Issue 

Subsequent visits which were seen previously at the health center have the following activities:
- Danger Signs
- Physical Exam
- Treatment Review

![image](https://user-images.githubusercontent.com/688507/160502848-f4497e80-7f2b-48b3-86a1-a5450ec43c45.png)

- The sub activities for "prior treatment history" and "physical exam" and workflows are the same as they are for an initial visit
- The sub activities for "danger signs" are the same as they are for a CHW.

## Tasks
- Create each of the activities and the workflows.

anvmn:
@adamhstewart 

So, we just fill the activities and end encounter?
No Next steps?

adamhstewart:
No we should already have next step logic from anc health center functionality.

anvmn:
@adamhstewart 
Next steps derrive from activites completed at current encounter.
So, it's not clear to me what should this be.

anvmn:
@adamhstewart 
Do you mean that it should be same Next steps as we have at subsequent encounter for CHW?

anvmn:
If ^^ is correct, I think we said that instead of sending to Health Center, we'll be sending to Hospital.
However, we also have 'Contact HC' subactivity for patients with Covid diagnosis, that are not improving.
Also, should nurse be scheduling follow up and providing health education?

anvmn:
@adamhstewart 

We have Treatment Review and Prior treatment history mentioned at requirements:

![Selection_646](https://user-images.githubusercontent.com/17882245/162638233-42889142-70fc-4a71-acb6-3561885c7fc3.png)

Prior treatment history is an activity that appears on first encounter both for Nurse and CHW.
Treatment Review is an activity that is viewed on subsequent encounter (CHW only), in case medication was prescribed during illness.

Which one do we need here?

anvmn:
@adamhstewart 

> we also have 'Contact HC' subactivity for patients with Covid diagnosis, that are not improving.

As we discussed at ctach up meeting, I made a check if there's an option to get to this path at nurse subsequent encounter.

If we have a COVID suspect, and nurse indicates that test was not performed, (and Malaria test that we give after is negative) we consider this case as Covid positive.
Also, encounter is not closed automatically, meaning that answers to questions will need to be provided.

Therefore, I don't see a way to get to subsequent encounter for nurse with 'Covid suspect' diagnosis.

[...]

PR #213, opened 2022-04-10
AI subsequent encounter for nurses - adjust activities

anvmn:
#201


anvmn:
Changes are trivial. No need to review. Merging.

PR #301, opened 2022-06-20
Demographic areas on reports

adamhstewart:
https://github.com/Gizra/ihangane/issues/2964


adamhstewart:
@anvmn Can you give me an in progress review on JUST the ANC report part of this. I'm trying to refactor the report to have a start and end date and to only consider the diagnosis on a first encounter in this time period.

The steps are:
1. get all the AI encounters and the associated individual participant (for the date range)
2. check each encounter to see that it's the first in the series (add that encounter to "the list")
3. Check the diagnosis for each encounter on "the list" and build the report.

adamhstewart:
@anvmn I notice that this method return more valid encounters during the date range using the date the participant node was created vs the `field_scheduled_date_value`. what is the difference between the two and which should I use to get the proper range?

server/hedley/modules/custom/hedley_admin/scripts/generate-acute-illness-report.php
```diff
@@ -10,37 +10,72 @@
...
+$first_encounters = [];
 foreach ($result as $item) {
-  $diagnosis[] = [
-    $field['settings']['allowed_values'][$item->type],
-    $item->counter,
+  $query = 'SELECT entity_id FROM field_data_field_individual_participant WHERE field_individual_participant_target_id = ' . $item['participant'];
+  $data = db_query($query)->fetchCol();
+
+  // If this enounter is equal to the smallest ID, it's the first.
+  if ($item['encounter'] = min($data)) {
```
anvmn: ==

server/hedley/modules/custom/hedley_admin/scripts/generate-acute-illness-report.php
```diff
@@ -10,37 +10,72 @@
...
+    $first_encounters[] = $item['encounter'];
+  }
+}
+
+// Now get the diagnosis for each first encounter.
+foreach ($first_encounters as $first_encounter) {
+
+  $query = 'SELECT field_acute_illness_diagnosis_value FROM field_data_field_acute_illness_diagnosis di WHERE entity_id = ' .$first_encounter;
```
anvmn: If first encounter was completed, but no diagnosis was made, value of `field_acute_illness_diagnosis_value` is `none`.
So, you want to filter that out.

anvmn (commented):
@adamhstewart 

In theory, after you make 2 fixes I pointed out, the calculation should give accurate results
I do need to mention that this method seems to be a little inefficient.
Say you got 5 encounters pointing to same participant.
With this logic, we, will pull the encounters of that participant 5 times.

What we could use is the fact that participant and first encounter are created on same day.
So, it'd be enough to pull all participants created between dates range.
Then, for each, pull the encounters and identify that with lesser NID as first encounter.

anvmn:
@adamhstewart 

> @anvmn I notice that this method return more valid encounters during the date range using the date the participant node was created vs the field_scheduled_date_value. what is the difference between the two and which should I use to get the proper range?

This is a field on encounter that indicates encounter start date and end dates.
It's a date field where value indicates start date and value2 indicates end date (when set).
Created on the other hand is a timestamp.
Maybe the difference you see got something to do with the way you define dates range in query, because field_scheduled_date_value is set to 00::00::00 of date. For example, `2023-02-21 00:00:00`.

PR #307, opened 2022-06-27
FHIR integration POC

amitaibu:
Added a FHIR integration module, that upon creation or update of data, it uploads it to a FHIR datastore.

1. A child's height is created or updated

image

2. An HTTP request sends the relevant info. No duplicates are created, so only a single patient or observation would be created.

image

3. `Observation `created on Google cloud. It references our `Patient`

image

4. And we have the `Patient` as well

image

Current PR assumes running it locally with DDEV, so one would need to have the following info under `settings.ddev.php` (ping me to get the _real_ values, as the ones here are dummy, for security reasons):

```
$conf['hedley_fhir_private_key'] = "-----BEGIN PRIVATE KEY-----\nMIIEvQIBADA...Ce3McVRrHyM=\n-----END PRIVATE KEY-----\n";
$conf['hedley_fhir_service_account_email'] = 'foo@bar.iam.gserviceaccount.com';
$conf['hedley_fhir_project_id'] = 'bar;
$conf['hedley_fhir_location'] = 'us-central1';
$conf['hedley_fhir_dataset_id'] = 'some-dataset';
$conf['hedley_fhir_fhir_store_id'] = 'some-fhir-store';
```

### DIFF
--- a/server/hedley/modules/custom/hedley_fhir/hedley_fhir.module
+++ b/server/hedley/modules/custom/hedley_fhir/hedley_fhir.module
@@ -15,6 +15,13 @@ function hedley_fhir_node_insert($node) {
 }
 
 /**
+ * Implements hook_node_update().
+ */
+function hedley_fhir_node_update($node) {
+  hedley_fhir_upload_to_datastore($node);
+}
+
+/**
  * Upload measurements to OpenMRS.
  *
  * @param object $node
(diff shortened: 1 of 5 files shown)


server/hedley/modules/custom/hedley_fhir/hedley_fhir.module
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * Implements hook_node_update().
+ */
+function hedley_fhir_node_update($node) {
+  hedley_fhir_upload_to_datastore($node);
+}
+
+/**
+ * Upload measurements to OpenMRS.
```
amitaibu:
```suggestion
 * Upload measurements to FHIR datastore.
```

Issue #244, opened 2022-04-26
ANC Subs Visit - Next Steps - Candidiasis (Yeast Infection) [2h]


jeanette-pelizzon:
## Issue 

When a patient identifies that they have the following symptoms, we provide health education and medication: 
(1) Symptom Review: Burning with Urination + vaginal itching or discharge

Pop-Up | Health Education | Medication
--|--|-- 
<img width="316" alt="Screen Shot 2022-04-26 at 5 35 24 PM" src="https://user-images.githubusercontent.com/77505464/165396449-994f80ae-8613-46d8-ad47-93673ecf06d7.png"> | <img width="320" alt="Screen Shot 2022-04-26 at 5 35 30 PM" src="https://user-images.githubusercontent.com/77505464/165396451-8d896e4c-d4db-43fb-a0a6-13fd307cac08.png"> | <img width="320" alt="Screen Shot 2022-04-26 at 5 35 37 PM" src="https://user-images.githubusercontent.com/77505464/165396452-7807ab86-0050-4f2e-b431-fe11f5aec638.png">

## Supporting Documents: 
You can see the logic for next steps[ here. ](https://docs.google.com/spreadsheets/d/1awAwPTMxRMhb2CSP3ocJvbOJravIaoPb2Q-RERe9KQ8/edit#gid=1762936493)

## Notes: 
Similar to the first ANC encounter, if a patient is receiving health education for multiple things they will stack in one health education sub-activity. If they are being referred to the hospital for multiple things, there is only one referral screen. 

## To Do:  
- Style next steps screens per design 
- Adjust logic per document

jeanette-pelizzon:
updated health education screen to include question: 

<img width="326" alt="Screen Shot 2022-05-03 at 1 04 52 PM" src="https://user-images.githubusercontent.com/77505464/166503531-193f4341-30d0-4923-b0d3-02af162a55f2.png">

anvmn:
@jeanette-pelizzon 

Looking at Figma, seems to me that this diagnosis is also made for `Abnormal Vaginal Discharge +  itching`:

![Selection_731](https://user-images.githubusercontent.com/17882245/167791791-df32808a-cd5f-4939-8230-60a905f1bbc9.png)

However, I don't see it mentioned in Logic file, and there's no issue for that.
Please advice.

jeanette-pelizzon:
[...]

PR #312, opened 2022-06-29
Fix typo and add some styling

Diama1:
[#244 comment](https://github.com/TIP-Global-Health/eheza-app/issues/244#issuecomment-1166611305)

### DIFF
--- a/client/src/assets/scss/_custom.scss
+++ b/client/src/assets/scss/_custom.scss
@@ -1273,7 +1273,6 @@ input::placeholder {
       .popup-heading-wrapper {
         height: 48px;
         text-align: center;
-        margin-right: 50px;
         margin-bottom: 1rem;
 
         img {
--- a/client/src/assets/scss/_custom.scss
+++ b/client/src/assets/scss/_custom.scss
@@ -1273,6 +1273,7 @@ input::placeholder {
       .popup-heading-wrapper {
         height: 48px;
         text-align: center;
+        margin-right: 50px;
         margin-bottom: 1rem;
 
         img {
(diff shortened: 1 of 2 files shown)


Diama1:
Screenshot from 2022-06-29 12-58-34
Screenshot from 2022-06-29 14-56-11

client/src/assets/scss/_custom.scss
```diff
@@ -1273,6 +1273,7 @@ input::placeholder {
       .popup-heading-wrapper {
         height: 48px;
         text-align: center;
+        margin-right: 50px;
```
anvmn: Why do we need this?

anvmn (changes requested):

client/src/assets/scss/_custom.scss
anvmn: I did not see any complaints about margin

Diama1: The issue was about the text that was not well placed and when I looked at the image I found that the text warning was not in the center because of that warning icon that comes first
Screenshot from 2022-06-29 15-22-59

anvmn: @Diama1 

The comment said `Popup text is different:`

Selection_879

It's not about margin. Please revert this change and deal with the text itself.

Diama1: > @Diama1
> 
> The comment said `Popup text is different:`
> 
> Selection_879
> 
> It's not about margin. Please revert this change and deal with the text itself.

Sure

Diama1:
Screenshot from 2022-06-29 21-02-55

anvmn (approved):

PR #314, opened 2022-06-30
Statistics - additional verificaiton when generating acute Illness data

anvmn:
https://github.com/Gizra/ihangane/issues/2965


anvmn:
Changes are trivial. Will MWG.

Issue #292, opened 2022-06-06
Whatsapp Integration [10h]


jeanette-pelizzon:
## Issue 

Nurses and CHWs need a way to share a patient's progress report with them. To do this, we are doing a WhatsApp integration. At the bottom of the progress reports, we are including a new button where they can "send via whatsapp". 

**Flow**: 
Hit "Send Via WhatsApp" Button --> consent Screen --> validate phone number (branches off if new number is needed)--> choose which portions of the report to send --> verify send 

**Video Overview**: 
You can watch a long winded but helpful video [here.](https://www.loom.com/share/abe535011b974ddeaa4c8609e2c1f678)

## Design: 

You can find the design files in Figma [here](https://www.figma.com/file/ar8UMXGjITVnICIAegbtIj/Dashboard-%7C-Progress-Reports-%7C-Etc?node-id=327%3A2239).  This flow is located to the right of the screen under the heading "Whatsapp Send" 

## Report sections for selection:

Source: https://docs.google.com/spreadsheets/d/1gQU_-YdyHKZKDz-3cIpF725aU3QSIxnkw8Hxe7zFx4U/edit#gid=0

![Selection_893](https://user-images.githubusercontent.com/17882245/177209733-c5838565-2ea7-4679-98bd-3c0c886a2389.png)

## To Do: 
- Create backend logic to send to whatapp. 
- Style screens per design

anvmn:
@jeanette-pelizzon @adamhstewart 

Can't open Figma link. It says file does not exist.
Points after reviewing the video:

1. As I see it, country code needs to be reconsidered. If we really need it, there are over 200 countries. Does it mean that we need to add all their flag icons.
Also, what about phone validation? Right now we don't validate anything, beside number characters. Do we need to change that? And, if we need country code, do we need to have validation logic for over 200 countries?

2. I don't think these are the right button labels:

![Selection_873](https://user-images.githubusercontent.com/17882245/175831178-7866b0ad-4580-4197-ae07-3efe7390e9f5.png)

3. At entering phone number flow, I don't see an option of canceling the action. What if we want to get back to previous step, or cancel sending message completly?

4. Selection of what part of report to send need to be well defined for all progress reports we need to support.
Technical implementation of this is taking a screenshot of what we want to send, meaning that we will have to view that section on tablet (lots of effort hiding here), so we need clear definition on all variations.
Also, I don't see how this should be supported for for ANC Labs section, for example.

[...]

PR #316, opened 2022-07-04
[FEATURE] Send Progress Report via WhatsApp

anvmn:
#292


anvmn:
Selection_894

anvmn:
Peek 2022-07-06 21-35

anvmn:
Peek 2022-07-09 00-06

anvmn:
Peek 2022-07-10 14-06

anvmn:
Peek 2022-07-10 21-21

anvmn:
@AronNovak please review.
Since there's another feature that we may need to deliver before this one, please **DO NOT MERGE** this PR,
as this is a base branch for the feature.
Also, there're couple of @todos here, that are handled in follow up PRs.

client/src/elm/Translate.elm
```diff
@@ -12968,6 +13048,78 @@ translationSet trans =
...
+            , kinyarwanda = Nothing
+            }
+
+        SendViaWhatsAppPhoneVerificationHeader ->
+            { english = "The phone number we have on file for this patient is:"
+            , kinyarwanda = Nothing
+            }
+
+        SendViaWhatsAppPhoneVerificationQuestion ->
+            { english = "Is this the correct number for the patients WhatsApp?"
```
AronNovak:
```suggestion
            { english = "Is this the correct number for the patient's WhatsApp?"
```

client/src/elm/Translate.elm
```diff
@@ -12968,6 +13048,78 @@ translationSet trans =
...
+            , kinyarwanda = Nothing
+            }
+
+        SendViaWhatsAppPhoneUpdateAtProfileQuestionPrefix ->
+            { english = "Would you like to update the patient profile for"
+            , kinyarwanda = Nothing
+            }
+
+        SendViaWhatsAppPhoneUpdateAtProfileQuestionSuffix ->
+            { english = "with the number"
```
AronNovak, line 15253: If it would be PHP + Drupal, we'd complain that this is not really translatable. Not sure about Kinyarwanda language, but splitting a sentence into chucks typically cause translatability issues.

anvmn: @AronNovak The thing is that there requiremetn is to have patient name and phone number as part of the sentence, and they should have blue color:

Selection_1462

This means that we need to wrap the sentence with HTML:
```
div [ class "content" ]
        [ translateText language Translate.SendViaWhatsAppPhoneUpdateAtProfileQuestionPrefix
        , span [] [ text person.name ]
        , translateText language Translate.SendViaWhatsAppPhoneUpdateAtProfileQuestionSuffix
        , span [] [ text phoneNumber ]
        , text "?"
        ]
```

Since our Translate works with pure strings right now, we implement it this way.
When translation will be added, we'll review alternatives, if current structure is problematic.

AronNovak (approved):

Issue #343, opened 2022-07-21
Upgrade Drupal core to v7.91 [1h]


anvmn:
https://www.drupal.org/project/drupal/releases/7.91

![Selection_936](https://user-images.githubusercontent.com/17882245/180169642-63deb921-99b3-4b89-8661-762fe5f1e5dc.png)

anvmn:
Deployed to live:
![Selection_938](https://user-images.githubusercontent.com/17882245/180280378-64e3ed84-8477-4574-b0ed-30a290ed91bd.png)

Closing...

PR #344, opened 2022-07-21
Drupal core v7.91

anvmn:
#343


anvmn:
Deployed to live:

Selection_938

Issue #426, opened 2022-08-17
Elevated Web Transactions time during statistics genration may cuase HTTP 504 [4h]


anvmn:
This moring, for several minutes I was getting HTTP 504 (timeout) when trying to log into live backend:

![Selection_1043](https://user-images.githubusercontent.com/17882245/185147158-ca1e4fdb-2391-4061-b8ca-2c3a427aaa4f.png)

Pantheon support suggested that this is probably caused by by slow transactions. From php-slow.log:

```
﻿[0x00007ff6c5417050] execute() /code/includes/database/database.inc:2284

[0x00007ff6c5416f80] execute() /code/includes/database/database.inc:744

[0x00007ff6c5416ea0] query() /code/includes/database/select.inc:1280

[0x00007ff6c5416e20] execute() /code/includes/entity.inc:1395

[0x00007ff6c5416d00] finishQuery() /code/modules/field/modules/field_sql_storage/field_sql_storage.module:760

[0x00007ff6c5416b00] field_sql_storage_field_storage_query() /code/includes/entity.inc:1219

[0x00007ff6c5416a90] execute() /code/profiles/hedley/modules/custom/hedley_chw/hedley_chw.module:310

[0x00007ff6c5416990] hedley_chw_get_village_residents() /code/profiles/hedley/modules/custom/hedley_stats/hedley_stats.module:399

[0x00007ff6c5416810] hedley_stats_get_villages_with_residents() /code/profiles/hedley/modules/custom/hedley_stats/hedley_stats.module:340

[0x00007ff6c54166e0] hedley_stats_calculate_stats_for_health_center() /code/profiles/hedley/modules/custom/hedley_restful/plugins/restful/node/HedleyRestfulSync.class.php:334

[0x00007ff6c5416620] getForHealthCenterStatistics() /code/profiles/hedley/modules/custom/hedley_restful/plugins/restful/node/HedleyRestfulSync.class.php:220

[0x00007ff6c5416490] getForHealthCenter() /code/profiles/hedley/modules/contrib/restful/plugins/restful/RestfulBase.php:818

[0x00007ff6c54163a0] process() /code/profiles/hedley/modules/contrib/restful/plugins/restful/RestfulBase.php:633

[0x00007ff6c5416320] get() /code/profiles/hedley/modules/contrib/restful/restful.module:513

[0x00007ff6c5416170] restful_menu_process_callback() /code/includes/menu.inc:527

[...]

PR #427, opened 2022-08-17
Statistics generation enhancements 

anvmn:
#426


anvmn:
@mariano-dagostino please review

server/hedley/modules/custom/hedley_stats/hedley_stats.module
```diff
@@ -313,69 +319,136 @@ function hedley_stats_calculate_stats_worker($queue_item) {
...
-    $wrapper = entity_metadata_wrapper('node', $health_center_id);
-
-    // Get all clinics of type 'FBF' for the HC.
-    $fbf_clinics = hedley_health_center_get_clinics_by_health_center($health_center_id, 'fbf');
-    // Calculate statistics that are provided for FBF clinics only.
-    $completed_program = $missed_sessions = [];
-    if (!empty($fbf_clinics)) {
-      list($completed_program, $missed_sessions) = hedley_stats_get_session_attendance_stats_by_health_center($health_center_id, $fbf_clinics);
-    }
+function hedley_stats_pull_stats_for_health_center($health_center_id) {
```
mariano-dagostino: @anvmn I'm not sure if the logic defined here is correct. My understanding is if there is no stats for any of the following cache_data the function will return an empty array. Could be possible to have a null value for any of the following `hedley_stats_handle_cache` and still want to treat it as a valid value?

If this is not the case. The PR looks good.

anvmn: @mariano-dagostino 
Any of those returnung NULL means that HC statistics are being recalculated by server.
In this case, we return empty array, indicating that statistics are not available at current moment.

anvmn:
@mariano-dagostino 
Made minor change at [82a1548](https://github.com/TIP-Global-Health/eheza-app/pull/427/commits/82a1548b9ac124b0f6e7a777a3ae38c66d5710f1).

mariano-dagostino (approved):

PR #431, opened 2022-08-22
`admin/content` performance tweak

AronNovak:
#2971

By default, no sorting is applied:
image


anvmn (approved):
@AronNovak anything that needs to be perfomed during deployment ,for this change to take effect?

AronNovak:
@anvmn Just to make sure that `admin_views_node` is in a default state.

PR #434, opened 2022-08-23
README improvement / DDEV fixes

AronNovak:
#2969

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,15 @@ The installation script will perform following steps:
   sites/default/files directory.
 * The database is dropped during the installation.
 
+#### Getting started
+
+First of all, you need to create a Device.
+1. `ddev drush uli`
+1. Fulfill https://eheza-app.ddev.site:4443/node/add/device , note the Pairing code.
+1. Fulfill https://eheza-app.ddev.site:4443/admin/people/create , note the PIN code, assign it to group(s) and health center(s).
+1. `ddev gulp`
+1. Visit , supply the Pairing code and the PIN.
+
 #### Deploy
 
 The default method assumes Pantheon as the hosting service provider.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -68,6 +68,15 @@ The installation script will perform following steps:
...
 * The database is dropped during the installation.
 
+#### Getting started
+
+First of all, you need to create a Device.
+1. `ddev drush uli`
+1. Fulfill https://eheza-app.ddev.site:4443/node/add/device , note the Pairing code.
+1. Fulfill https://eheza-app.ddev.site:4443/admin/people/create , note the PIN code, assign it to group(s) and health center(s).
+1. `ddev gulp`
+1. Visit , supply the Pairing code and the PIN.
```
AronNovak: @anvmn anything missing here? Every time I start working on this project after a break, I find myself not being able to login to the client frontend, so this time I would like to capture every minor details. What I could recall by heart wasn't enough :)
Or is it already described more verbosely somewhere else?

anvmn: @AronNovak 

> https://eheza-app.ddev.site:4443/admin/people/create

This is not the correct URL. For log in, used pincode is that of a nurse, which is a CT, and not user.

However, if we talk about actions that are performed after project installation, including migration of default content, we can use the content that was generated during migration.

There're 2 devices created - pairing code 12345678 and 87654321.
There're 2 nurses created - regular nurse (pin 1234), CHW nurse (pin 2345).

There're also 2 Health centers created.
The one I always work with is `Nyange`.

So, what needs to be done is:
1. Pair using pairing code.
2. Login with pincode.
3. Sync Nyange HC.

AronNovak: Ok, thanks, I added the explanation related to when do we actually need to deal with nurses and devices by hand.

anvmn (approved):

Issue #443, opened 2022-08-26
Failure during statistics sync [2h]


anvmn:
![Selection_1066](https://user-images.githubusercontent.com/17882245/186912035-649426cc-5c06-4368-aab4-40baf5fd26d3.png)

anvmn:
Problem is with HC that got no family planning measurements, and therefore stats generated for it is NULL.
When we return empty result, the response is not as expected:

![Selection_1065](https://user-images.githubusercontent.com/17882245/186912350-b200ee22-f589-4103-a7fa-47a294036c9a.png)

So, there're 2 things to fix:
1. Make sure proper empty result is returned.
2. When a component of statistics object is calculated, and is an empty array, return that object as empty, and not whole statistics.

anvmn:
@adamhstewart Fix for point 1 (#444) deployed to live.

PR #444, opened 2022-08-26
Statistics fix for returned empty result

anvmn:
#443


anvmn:
Changes are trivial. Merging.

Issue #419, opened 2022-08-15
NCD - Labs - Urine Dipstick [8h]


jeanette-pelizzon:
## Issue 

NCD labs will feature the usual Urine Dipstick lab. They will be able to choose if they run the short or the long dip when they order the lab and the results screen will show the appropriate inputs. There is logic associated with the glucose results in these lab results. 

**Labs Order**
<img width="282" alt="Screen Shot 2022-08-15 at 2 46 21 PM" src="https://user-images.githubusercontent.com/77505464/184724015-2c754658-1c71-400d-9b9f-920740ec97f2.png">

**Lab Results - Long**
<img width="198" alt="Screen Shot 2022-08-15 at 2 47 00 PM" src="https://user-images.githubusercontent.com/77505464/184724073-7d247983-6db6-4d54-9181-b0a86c66abe0.png">

**Lab Results - Short**
<img width="198" alt="Screen Shot 2022-08-15 at 2 47 06 PM" src="https://user-images.githubusercontent.com/77505464/184724098-16f6ad0e-e1e8-4e4c-a476-eed7631dedf0.png">

## To Do: 
- Style screens per design
- Adjust necessary logic (if needed)

balagan73:
https://nutrition-ihangane.pantheonsite.io/

### QA Review

### Found issues and questions:

1. Grammar. I think either the 'a' or the plural 's' should be dropped.
![image](https://user-images.githubusercontent.com/3536805/193535599-ac856bcc-da45-4cd4-b6ff-4ab6f7540220.png)

2. The results of the NCD encounter are displayed in the Antenatal progress report. Is this OK? I can find the hypertension diagnoses, but not the Diabetes.
![image](https://user-images.githubusercontent.com/3536805/193537562-6da679e0-8927-4bb8-ab7b-3e08e34926d0.png)

3. Results are still pending, but they display as abnormal in progress report. I selected Long dipstick, but both short and long dipstick displays.

Short dipstick|Results pending
-|-
![image](https://user-images.githubusercontent.com/3536805/193541240-496fe941-075a-47fe-8d80-85eba1e422ef.png)|![image](https://user-images.githubusercontent.com/3536805/193541286-0583e447-afca-46ed-a9e6-4c42dc9fe276.png)

4. The Specific gravity test results are missing from Long dip test results, but I think that was removed in another issue. See image below in point 1. 

### Checked and approved:

1. Panels and questions display as designed.

[...]

Issue #417, opened 2022-08-15
NCD - Labs - Random Blood Sugar [6h]


jeanette-pelizzon:
## Issue 

NCD labs will include Random Blood Sugar tests. The results from this test will be used to diagnose diabetes. 

**Lab Order**
<img width="295" alt="Screen Shot 2022-08-15 at 2 23 36 PM" src="https://user-images.githubusercontent.com/77505464/184721138-1cde0386-f2f7-490f-803d-4f90ea182cab.png">

**Lab Results**
<img width="216" alt="Screen Shot 2022-08-15 at 2 25 14 PM" src="https://user-images.githubusercontent.com/77505464/184721211-42c2e3ea-c421-46cd-892d-2d6e0ae98857.png">

## To Do 
- Style screens per design 
- Create required back end logic (if needed)

balagan73:
https://nutrition-ihangane.pantheonsite.io/

### QA Review

### Found issues and questions:

1. Patient was fasting, blood level of 127, diagnoses displays correctly but not indicated on progress report. Same when patient was not fasting.
![image](https://user-images.githubusercontent.com/3536805/193559937-fc124aa9-77ae-4256-a2fa-30ed413d70d9.png)

2. On progress report Normal range is 74-110, but we diagnose above 126. Is this OK?
![image](https://user-images.githubusercontent.com/3536805/193560015-0c84bd6e-e231-4102-8f45-cf0d1876eb10.png)

3. As I remember the word fasting is not used in Rwanda and 'before meal' should be used.

Fasting|Non-fasting
-|-
![image](https://user-images.githubusercontent.com/3536805/193560219-cf28f193-a9ae-4c1b-81c1-916906efd909.png)|![image](https://user-images.githubusercontent.com/3536805/193561134-3a22ba18-a4da-4b85-9551-090bb26953d3.png)

4. When measurement is taken after meal, on the progress report the before meal normal range is indicated, and the measurement is marked with red even if it is far from the diagnosable limit.
![image](https://user-images.githubusercontent.com/3536805/193561840-bc4eab57-0300-4100-9eaf-332915f31396.png)

### Checked and approved:

1. The limits specified for diagnoses work correctly. See the images with found issues point 3.

jeanette-pelizzon:
1. unsure why its not appearing on the progress report
2. this is accurate - the normal ranges are 74 -110 and the patient had 126 which is well above the normal range and therefore abnormal and an issue. 
[...]

Issue #416, opened 2022-08-15
NCD - Labs - Infrastructure [8h]


jeanette-pelizzon:
## Issue 

We will have several labs activities in the NCD Module. SVG's can be found in Figma in the components file. These labs all require the lab orders (the labs activity in the encounter) AND the lab results screens in the appropriate place. 

- Random Blood Sugar 
<img width="119" alt="Screen Shot 2022-08-15 at 2 20 00 PM" src="https://user-images.githubusercontent.com/77505464/184720539-48067db1-bc6a-44e7-a666-2c05896e0966.png">

- Creatinine (New)  
<img width="98" alt="Screen Shot 2022-08-15 at 2 20 03 PM" src="https://user-images.githubusercontent.com/77505464/184720606-d8007c76-0763-419f-9f51-47db576712f6.png">

- Urine Dipstick 
<img width="109" alt="Screen Shot 2022-08-15 at 2 20 09 PM" src="https://user-images.githubusercontent.com/77505464/184720626-241d7575-32e3-414e-bd44-a3a4f808d991.png">

- HIV 
<img width="90" alt="Screen Shot 2022-08-15 at 2 20 12 PM" src="https://user-images.githubusercontent.com/77505464/184720640-cb7ab8ab-e6f5-4d36-bd9d-e41da50b5780.png">

- Pregnancy 
<img width="119" alt="Screen Shot 2022-08-15 at 2 20 16 PM" src="https://user-images.githubusercontent.com/77505464/184720657-cc9a8035-216d-4a08-b914-13c48a13b7d5.png">

- Liver Function (New) 
<img width="94" alt="Screen Shot 2022-08-15 at 2 20 27 PM" src="https://user-images.githubusercontent.com/77505464/184720675-80169c89-02bf-4988-bb26-6a3ff1f4f381.png">

## To Do
- Create infrastructure for Labs activities

balagan73:
https://nutrition-ihangane.pantheonsite.io/

### QA Review

### Checked and approved:

1. The choices under Laboratory are the same as listed here.
![image](https://user-images.githubusercontent.com/3536805/193310074-0819fb27-67f1-4864-8c70-4c7f16cc91d9.png)

PR #452, opened 2022-09-04
NCD - Labs

anvmn:
#416 
#417 
#419


anvmn:
@mariano-dagostino 

Please review.

For the contect, we're making Labs activity with all it's subactivities (different types of lab tests) available globally, to be able to use at NCD module (so far it was only for Prenatal).
So, there's lots of renaming, code relocation and then appliance within NCD.

client/src/elm/Backend/Update.elm
```diff
@@ -4792,25 +4846,159 @@ generatePrenatalLabsResultsAddedMsgs currentDate after test id =
...
+                                                let
+                                                    updatedCompletedTests =
+                                                        test :: completedTests
+
+                                                    resolutionDate =
+                                                        -- When all performed tests are completed, and Next Steps are either
+                                                        -- completed, or not required, setting today as resolution date.
+                                                        if
+                                                            (List.length updatedCompletedTests == List.length performedTests)
+                                                                && -- @todo
```
mariano-dagostino, line 4965: @anvmn How do you track the pending `@todo` items? Perhaps creating the issue and linking it here?

mariano-dagostino (approved):
@anvmn Looks good. Just a question about how you handle the 5 or so `@todo` items you added in this PR?

client/src/elm/Backend/Update.elm
anvmn: @mariano-dagostino 

When developing a module which 'replicates' the infrastructure of similar module (for example, additional type of individual encounter), there're quire alot of @todos that are not related to current PR development and are completed at follow up PRs.
Before submiting PR for review, I verify none of @todos are needed in this PR.
Once module is competed, I verify there are no @todos left.

Issue #423, opened 2022-08-15
NCD - Next Steps - Infrastructure [6h]


jeanette-pelizzon:
## Issue 

Similar to the ANC module, in NCD we will have labs drawn, the patient will (theoretically) wait for their results to come back later that day and then have next steps associated with those labs then. We are bringing over the exact same flow from ANC with the exception of needing to re-do vitals. The vitals do NOT need to be rechecked in NCD. 

Here is a rough overview of the flow:
Main encounter -> any next steps that can be diagnosed already -> case management -> enter lab results -> next steps diagnosed from lab results

## To Do:
- Create infrastructure 
- Create Logic

balagan73:
### QA Review

### Checked and approved:

1. After finishing the QA for almost all the NCD issues, I haven't encountered any problems with the Next steps infrastructure, it works correctly.

PR #456, opened 2022-09-07
NCD - Next Steps - Activities

anvmn:
#423


anvmn:
Peek 2022-09-11 13-30

anvmn:
@AronNovak 

Please review. Note that this PR defines Next steps activities, and there are several @todos that are logic related, and not implmented here. They're taken care of at #459.

client/src/assets/scss/_new.scss
```diff
@@ -3994,6 +3998,42 @@ div.page-activity.ncd {
...
+
+      .instructions {
+        margin-bottom: 0;
+      }
+    }
+
+    .form.referral {
+
+      .separator:last-of-type {
+        display: none;
```
AronNovak, line 4033: Isn't it something that we could solve in Elm more or less with the same easiness?

anvmn: @AronNovak 

Well, what happens is that we have several blocks, each ending with separator.
We don't know which ones will apear.
So, we could modify code to insert these separators between the blocks, but it's much harder than this CSS.

AronNovak (approved):

Issue #464, opened 2022-09-13
NCDA - Group Nutrition Encounter [7h]


jeanette-pelizzon:
## Issue 

We are adding in an activity to the Nutrition Encounter at the Healthcenter (FBF) called "Questionnaire." This activity will appear in both the individual and group encounter. This issue specifically deals with the group encoutner.

The Questionnaire activity has 14 yes/no boolean questions. The answers will be mapped to the NCDA child scorecard. It will be the last activity a nurse completes with a patient.

<img width="222" alt="Screen Shot 2022-09-15 at 8 53 19 AM" src="https://user-images.githubusercontent.com/77505464/190450381-bf0fda63-b8b6-4ce6-bea2-7640557c5ffd.png">

## To Do:
- add in the Questionnaire activity 
- style screen per design
- any back end logic

jeanette-pelizzon:
Note: updated image in main issue, wording was changed. Please double check with Adam or Jeanette for final wording on the "meal frequency" question as this is still outstanding with the team.

anvmn:
@jeanette-pelizzon 
Can you please send the link to Figma, where I can see the questions form?

jeanette-pelizzon:
[Link is here](https://www.figma.com/file/bP3qdRVzCW6bUrmjS0hIco/Nutrition-%40-HC?node-id=2%3A3). Sorry! (updated)

anvmn:
@adamhstewart Depoyed to Nutrition env for QA

balagan73:
@jeanette-pelizzon Can I have access to the figma file? 

### QA Review

### Found issues and questions:

1. Not related to this issue, but when participant has many children, text overlaps the icon.
![image](https://user-images.githubusercontent.com/3536805/201362730-433fea1e-cbbf-401e-9363-8fde73fe6e66.png)

### Checked and approved:

1. The NCDA activity displays for children under 2 in the FBF group encounters.
![image](https://user-images.githubusercontent.com/3536805/201331874-41dd1b01-b497-4712-b942-40365184d443.png)

jeanette-pelizzon:
@balagan73 You should be able to access [via this link](https://www.figma.com/file/bP3qdRVzCW6bUrmjS0hIco/Nutrition-%40-HC?node-id=2%3A3&t=YVG7sQ8cvUdDNI4R-1). You can slack me your email address and I can add you directly as well if needed.

PR #504, opened 2022-10-12
NCDA activity - Group encounter

anvmn:
#464


anvmn:
Peek 2022-10-12 15-47

anvmn:
@AronNovak please review

client/src/elm/Measurement/Utils.elm
```diff
@@ -3858,3 +3882,241 @@ toNCDAValue form =
...
+
+        -- In how many days (from the session date) will the child be 2 years
+        -- old?
+        daysUntilTwoYearsOld =
+            (365 * 2) - age
+
+        -- In how many days (from the session date) will the child be 1 year
+        -- old?
+        daysUntilOneYearOld =
+            365 - age
```
AronNovak, line 3984: > In the Julian calendar, a year contains either 365 or 366 days, and the average is 365.25 calendar days.

It's this or 366, but I guess it's ok here.

AronNovak (approved):

client/src/elm/Measurement/Utils.elm
anvmn: Well, this is very old code which I had to move.
I don't think it makes much difference, but the code is not in use anyway.

Issue #510, opened 2022-10-19
Resolve PHP warning at logger [1h]


anvmn:
![Selection_1174](https://user-images.githubusercontent.com/17882245/196775686-cba5d034-6fbb-4c47-b9a8-8f08a1aa2e7c.png)

This floods the logger, and overwrites important messages (errors for example).

anvmn:
Deployed to live and verified. Closing.

PR #511, opened 2022-10-19
[Hotfix] Resolve PHP warning at logger

anvmn:
#510


AronNovak:
@anvmn should we merge this, is it on `live`?

anvmn:
@AronNovak 

This is not on live. Seemed not reasonable to perform a deploy just for this.
I'll merge it once we have enough content to deploy.

Issue #519, opened 2022-10-24
NCDA - Data Pulling - Nutrition Behavior 


jeanette-pelizzon:
## Issue 

**Breastfed baby for 6 months without interruption**

_What module is the data being pulled from_: New Questionnaire (this question appears for the first 6 months months of life then switches to the complementary feeding question)

_Green Checkmark on scorecard_: yes to the questionnaire 

_Red X on scorecard:_ no to the questionnaire

_Dash on scorecard_: no data recorded - a missed appointment  

**Appropriate complementary feeding (6-24 mo)**

_What module is the data being pulled from_:  New Questionnaire (this question appears for the after the first 6months months of life - it switches out the breastfeeding question)

_Green Checkmark on scorecard_: yes to the questionnaire 

_Red X on scorecard:_ no to the questionnaire

_Dash on scorecard_: no data recorded - a missed appointment  

**Does the child have a diverse diet?**

_What module is the data being pulled from_:  New Questionnaire  - Question reads as "Does the child receive food items from the 5 food groups?"

_Green Checkmark on scorecard_: yes to the questionnaire 

_Red X on scorecard:_ no to the questionnaire

_Dash on scorecard_: no data recorded - a missed appointment 

**Number of times a child eats a day**

_What Module is the data being pulled from:_ New questionnaire - it is an age dependent question - 1. From 6 up to 9 months: Feed your young child complementary foods 2 to 3 times a day 
2. From 9 up to 12 months: Feed your young child complementary foods 3 to 4 times a day 
3.  From 12 up to 24 months: Feed your young child complementary foods at least 5 times a day

_Green Checkmark on scorecard_: yes to the questionnaire 

_Red X on scorecard:_ no to the questionnaire

_Dash on scorecard_: no data recorded - a missed appointment 

<img width="563" alt="Screen Shot 2022-10-24 at 1 39 33 PM" src="https://user-images.githubusercontent.com/77505464/197625555-d0b82153-b67e-438e-9afe-cc95d369e710.png">

## Related Issues/Docs: 
- Meta Issue: #515 
- [Reference document](https://docs.google.com/spreadsheets/d/1sAYhkh15V2CZcu9OC7pN1yJ3NPShU3q3/edit#gid=1194481634)t. Use the master tab. 
- [Figma ](https://www.figma.com/file/bP3qdRVzCW6bUrmjS0hIco/Nutrition-%40-HC?node-id=316%3A188)

anvmn:
[...]

PR #533, opened 2022-10-30
NCDA Report - Nutrition Behavior pane

anvmn:
#519 
#534

### DIFF
--- a/server/RoboFile.php
+++ b/server/RoboFile.php
@@ -38,7 +38,7 @@ class RoboFile extends Tasks {
 
     if ($result->getMessage()) {
       $this->yell($result->getMessage());
-      // throw new Exception('The GitHub working directory is dirty. Please commit any pending changes or add to .gitignore.');
+      throw new Exception('The GitHub working directory is dirty. Please commit any pending changes or add to .gitignore.');
     }
 
     $result = $this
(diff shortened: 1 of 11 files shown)


anvmn:
Selection_1229

anvmn:
Selection_1231

anvmn:
Peek 2022-11-06 20-34

anvmn:
@AronNovak please review

server/RoboFile.php
```diff
@@ -38,7 +38,7 @@ public function deployPantheon($branchName = 'master') {
 
     if ($result->getMessage()) {
       $this->yell($result->getMessage());
-      throw new Exception('The GitHub working directory is dirty. Please commit any pending changes or add to .gitignore.');
+      // throw new Exception('The GitHub working directory is dirty. Please commit any pending changes or add to .gitignore.');
```
AronNovak: We should solve it in a different way, we should keep the working directory clean simply.

AronNovak (approved):
Looks good, except the deploy code part.

server/RoboFile.php
anvmn: This was never meant to be commited. Reverted.

Issue #574, opened 2022-12-26
Whatsapp Integration - upload screenshots to backend [12h]


anvmn:
### Task: 

Implement point 2 from https://github.com/TIP-Global-Health/eheza-app/issues/339#issuecomment-1363270284.

### Todo:

1. Create content type that will store the screenshot with coresponding data.
2. Extend sync process to upload the images and create nodes for them on backend.

PR #575, opened 2022-12-26
Whatsapp Integration - upload data to backend

anvmn:
#574 

### Deployment instructions:

`drush updb -y`


anvmn:
@AronNovak please review

client/src/js/app.js
```diff
@@ -680,6 +680,103 @@ elmApp.ports.askFromIndexDb.subscribe(function(info) {
...
+        const screenshotsUploadCache = "screenshots-upload";
+        const cache = await caches.open(screenshotsUploadCache);
+
+        result.forEach(async function(row, index) {
+            const cachedResponse = await cache.match(row.screenshot);
+
+            if (cachedResponse) {
+              const blob = await cachedResponse.blob();
+              const formData = new FormData();
+              const imageName = 'whatsapp-upload-' + getRandom8Digits() + '.png';
```
AronNovak, line 709: one note about naming that it's not random, but it's basically the current time in some format, right? Do we need to ensure that filenames need to be unique or not at this point?

AronNovak: (I guess Drupal will do its standard file processing, so not)

AronNovak (approved):

Issue #632, opened 2023-01-25
ANC - Progress Report - Malaria Blood Smear [8h]


jeanette-pelizzon:
## Issue
We had to introduce a blood smear option for malaria testing. This test will not always be run but when it does get used it needs to appear in the lab results section of the progress report. It will only show up on the progress report if it has been run. 

- Name of Test: Malaria Blood Smear
- Normal Result = Negative 
- Result Options = Negative, +, ++, +++ 

## To Do: 
- Add this test to the progress report 
- Let jeanette know if you need a design for this but it should be pretty straightforward.

LudmilaGoichman:
@anvmn @jeanette-pelizzon 

- nutrition env
- tablet

### Questions/Found Issues:
1. The test name should be `Malaria Blood Smear`

> Name of Test: Malaria Blood Smear

<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219959096-f9c8fece-4ff7-4b9a-9f83-4196f17d3099.png"> |<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219960008-282ca8b5-99f4-4c05-9f6e-ffbfc106b699.png">
---|---

2. When the Malaria Blood Smear test result is `negative`, the result doesn't appear in the report (in all encounters)

<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219959298-d46df29b-a337-4c71-ab9f-2744665dbdcd.png">|<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219959351-dd95734e-783b-45ec-a157-f631b29dffeb.png">
---|---

3. when the test wasn't be performed, the record appears in the report

>  It will only show up on the progress report if it has been run.

first encounter:

<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219960224-1c2bce1d-d409-41ee-9bb6-9cf84209333e.png">|<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219959547-06bd61a6-f462-4d55-9935-ce41501dd252.png">
---|---

subsequent encounter:

<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219960095-e9df286d-3a47-41ed-8157-4e17a07902fe.png">| <img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219960124-7c7d682c-ca7b-4e33-9f9b-ce65d4bcf1d4.png">

[...]

Issue #631, opened 2023-01-25
ANC - Labs - Malaria [6h]


jeanette-pelizzon:
## Issue 
We need to create a new pathway if Malaria RDT was not completed. Apparently, the supplies to run an RDT are sometimes out of stock and they need a different way to diagnose malaria. 

If RDT was not completed the following conditional questions should appear: 
 - check list of why not 
 - Was a blood smear performed 
 - If yes to the blood smear question, then What was the result.
 
<img width="244" alt="Screen Shot 2023-01-25 at 8 37 37 AM" src="https://user-images.githubusercontent.com/77505464/214623347-8b8c2efa-2da1-4894-bb37-bd2675846671.png">

## Figma
- You can see Figma design[ here](https://www.figma.com/file/0rIDfLlgZ3KHSt0VAOP2za/ANC-%7C-Health-Center?node-id=1640%3A39869&t=RIqn4CFdJGu3M2Io-0)

## Note 
- This will need to update to the progress report. I'll make a separate issue for that. 
- The options for results may change - waiting to hear back from the team. 

## To Do
- update the conditional questions for if no RDT was performed.

anvmn:
@jeanette-pelizzon 

You say `different way to diagnose Malaria`. Does this mean that answers to follow up question should somehow result in Malaria Diagnosis?

adamhstewart:
@anvmn any non-negative result on the blood smear (+, ++, +++) all result in a malaria diagnosis (same as a positive RDT)

LudmilaGoichman:
@jeanette-pelizzon @anvmn 

- nutrition env
- tablet

### Question:
Conditional questions appear not when the second question is answered No, but when the first question is answered No. I think this is correct, but I want to make sure

<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219956862-1164c8b7-f41b-4c96-95c5-2982ebd6c3a8.png">|<img width="365" alt="image" src="https://user-images.githubusercontent.com/35059802/219957071-456f5e26-4324-474a-b0fc-657af0306d7a.png">
---|---

### Checked and Approved:
 the conditional questions have been updated

[...]

PR #656, opened 2023-02-12
ANC Changes - Malaria lab

anvmn:
#631 
#632


anvmn:
Peek 2023-02-12 21-49

anvmn:
Selection_1571

client/src/elm/Backend/Measurement/Model.elm
```diff
@@ -1016,6 +1017,14 @@ type TestResult
     | TestIndeterminate
 
 
+type BloodSmearResult
```
mariano-dagostino, line 1020: @anvmn Do you think is worth to refactor this to something more specific:

```
type BloodSmearResult = Maybe BloodSmearTestResult

type BloodSmearTestResult = Negative 
                          | Positive NumberOfPlus
```

It seems a bit error prone to have everything inside one type. But up to you.

mariano-dagostino (approved):

client/src/elm/Backend/Measurement/Model.elm
anvmn: @mariano-dagostino 
We list all possible results here, so, no need to refactor.
We have same structure at numerous other places, and there were no issues with it.

Issue #654, opened 2023-02-07
Stock Mgmt - Correct Entry [8h]


jeanette-pelizzon:
## Issue

As we are manually entering in data, we need a way for users to go in and correct any errors. We also need these to be tracked separately, rather than just updating the main entry. 

To do this, we have created the following screen: 
<img width="416" alt="Screen Shot 2023-02-07 at 3 08 49 PM" src="https://user-images.githubusercontent.com/77505464/217387822-3c179de8-6309-45a8-b5ff-2766f188a496.png">

We will log: 
- Date (can be in the past or the current day)
- How much stock was added on that date 
- How much stock was issued on that date 
- and a reason as to why the correction is needed. 

This will log as it's own line entry in the month details (for the month that = the date). Example: It's Feb 2 and I need to adjust an entry for Jan 30. That adjustment will appear on the Jan month details.

anvmn:
@adamhstewart 

I believe following format will better suite our needs:

- Selection - stock added / deducted.
- Number input 
- Date
- Reason selection - should be dynamic, according to added/deducted selection.

Also, this should follow same pattern of Receive Stock:
1. Make sure person who's logged in, is the one performing the operation.
2. Signature.

adamhstewart:
@anvmn agree to the above, except:

> How much stock was issued on that date

This should not be an adjustment... issued stock is only through FBF.

anvmn:
@adamhstewart
should be an option to add to stock here, or not?

Issue #644, opened 2023-01-30
Stock Mgmt - Receive Stock [10h]


jeanette-pelizzon:
- On hold until we get answers from team 

## Issue
When a user clicks "Receive Stock", they will be taken to a new screen where they can input all the necessary details.  

- Date - calendar will pop up 
- Where is the stock from - drop down (options in red on the screenshot)
- Expiration Date - Calendar will pop up 
- How much stock is being received - numerical input 
- Months of Stock - this should be automatically calculated (Months of stock=Total current stock(balance)/monthly consumption average) 
- Log In verification - make sure its the correct user - if no, they need to select the correct user OR @adamhstewart do we want them to log out and log in as the correct user - forces them to be diligent about who they are logged in as. 
- Signature 

<img width="628" alt="Screen Shot 2023-02-08 at 8 11 10 AM" src="https://user-images.githubusercontent.com/77505464/217585942-625b6a56-7135-41e7-af1e-21d9c86c4fa6.png">

jeanette-pelizzon:
Main issue updated. @adamhstewart need your input on the user verification bit.

anvmn:
@adamhstewart 

Is this still on hold?

Comments:

1. Log In verification should be first question. Also, need to define what happens if the answer is NO.
2. Please define formula for Months of Stock.

jeanette-pelizzon:
**Updated Designs for Receiving Stock:** 

**The first screen** will verify the user -
<img width="308" alt="Screen Shot 2023-02-15 at 8 20 40 AM" src="https://user-images.githubusercontent.com/77505464/219088203-fa1c20ae-8236-4216-b1d5-6c3e9626b926.png">

**If answer is no**, this screen will appear and should take them back to the home screen.
<img width="350" alt="Screen Shot 2023-02-15 at 8 22 16 AM" src="https://user-images.githubusercontent.com/77505464/219088477-879e1ae0-7439-4fa6-9da0-2b38ea316212.png">

**If the answer is yes**, then they can proceed with the rest of the receive stock inputs. *NOTE*: we have added in a freeform field for users to input observations so that it matches with the form they use in real life. This is the first time we are having a free form input field. @adamhstewart do we need a character limit? Just remember kinyarwanda words are super long
[...]

PR #660, opened 2023-02-16
Stock Management - Receive stock + Correct Entry

anvmn:
#644 
#654


anvmn:
Peek 2023-02-16 23-17

client/src/assets/scss/_new.scss
```diff
@@ -4837,4 +4836,48 @@ div.stock-management {
...
+      border: 2px solid $color-light-gray;
+      color: $color-text;
+      padding: 10px;
+    }
+  }
+
+  // .overlay .ui.modal.diagnosis-popup {
+  //   margin-left: -250px;
+  //   width: 500px;
+  // }
```
AronNovak, line 4882:
```suggestion
```

perhaps

client/src/elm/Pages/Utils.elm
```diff
@@ -1185,3 +1185,28 @@ customSaveButton language active msg label =
...
         [ text <| translate language label ]
+
+
+customPopup : Language -> Bool -> TranslationId -> ( Html msg, Html msg, msg ) -> Html msg
+customPopup language showWarning actionLabel ( topMessage, bottomMessage, action ) =
+    div [ class "ui active modal diagnosis-popup" ]
+        [ div [ class "content" ] <|
+            [ div [ class "popup-heading-wrapper" ]
+                [ img [ src "assets/images/exclamation-red.png" ] []
+                , div [ class "popup-heading" ] [ text <| translate language Translate.Warning ++ "!" ]
```
AronNovak, line 1196: Shouldn't the exclamation mark be part of the `translate.Warning`?

AronNovak:
@anvmn I restarted CircleCI and now it's green.

client/src/elm/Pages/Utils.elm
anvmn: Well,  we can make it whatever we like, but as I see it, we translate a word.
So, no punctuation marks need to be part of it.

AronNovak (approved):

client/src/assets/scss/_new.scss
anvmn: This is removed at follow up PR (#664), so no need to handle it here.

Issue #665, opened 2023-03-06
Resilience Messaging - Well Being Icon [4hr]


jeanette-pelizzon:
## Issue 

Since we have changed the module name from messaging to well-being, we also need to change the icon. We talked about doing a Lotus Flower but Angele said it does not resonate with Rwandans. 

This is an alternative: 
<img width="180" alt="Screen Shot 2023-03-05 at 4 53 07 PM" src="https://user-images.githubusercontent.com/77505464/222996891-1e7c4a24-d7ef-4732-af8a-5efbbfe93c9e.png">

[- Figma Link ](https://www.figma.com/file/Z5IoKolLaRe1NGuRe6MozG/Resilience-Messaging?node-id=775%3A1896&t=fESs3JnRQ2bk50Rt-0)

To Do: 

- Update Icon

PR #667, opened 2023-03-08
Change the well being icon

Diama1:
#665


Diama1:
Screenshot from 2023-03-08 13-40-19

client/src/elm/Pages/MessagingCenter/View.elm
```diff
@@ -938,7 +938,7 @@ messageOptionsDialog language currentTime currentDate nurseId tab state =
                         [ class "ui fluid button cyan"
                         , onClick <| ToggleMessageRead nurseId messageId message isRead
                         ]
-                        [ img [ src "assets/images/envelope.svg" ] []
+                        [ img [ src "assets/images/wellbeing.svg" ] []
```
anvmn: Why?

anvmn (changes requested):

client/src/assets/images/envelope.svg
anvmn, line 1: @Diama1 Can you revert this please. There should be no changes here.

Diama1: @anvmn  I deleted it in the previous commit then when you gave a comment, I realized that it was being used some where else then I put it back

anvmn (approved):

PR #672, opened 2023-03-14
Restore APP update functionality

anvmn:
https://github.com/Gizra/ihangane/issues/3037

### DIFF
--- a/.ddev/.gitignore
+++ b/.ddev/.gitignore
@@ -23,11 +23,14 @@
 /db_snapshots
 /import-db
 /import.yaml
-/mutagen
+/mutagen/mutagen.yml
+/mutagen/.start-synced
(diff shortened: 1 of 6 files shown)


.ddev/.gitignore
```diff
@@ -23,11 +23,14 @@
 /db_snapshots
 /import-db
 /import.yaml
-/mutagen
+/mutagen/mutagen.yml
```
mariano-dagostino, line 26: @anvmn This file is managed by ddev (as you can see in the first lines of the file), but it seems it is committed to github in early version. I think the proper fix is to 

```
git rm .ddev/.gitignore
git push ...
```

mariano-dagostino (approved):
@anvmn This looks good as is, you will have the same issue in next versions of ddev.

Issue #681, opened 2023-04-02
Stock Mgmt - design a process for signatures upload to backend + sync to devices [8h]


anvmn:
After #680 is completed

isapisa:
@LudmilaGoichman see https://github.com/TIP-Global-Health/eheza-app/pull/685

PR #685, opened 2023-04-10
Stock Management - store and sync signature

anvmn:
#681


anvmn:
Selection_1756

client/src/elm/Backend/NutritionEncounter/Model.elm
```diff
@@ -67,7 +67,7 @@ type Msg
     | HandleSavedMuac (WebData ())
     | SaveNutrition PersonId (Maybe NutritionNutritionId) NutritionValue
     | HandleSavedNutrition (WebData ())
-    | SavePhoto PersonId (Maybe NutritionPhotoId) PhotoUrl
+    | SavePhoto PersonId (Maybe NutritionPhotoId) ImageUrl
```
AronNovak, line 70: Shouldn't we rename it to `SaveImage` to be coherent?

client/src/elm/Backend/PrenatalEncounter/Model.elm
```diff
@@ -194,7 +194,7 @@ type Msg
     | HandleSavedSocialHistory (WebData ())
     | SaveVitals PersonId (Maybe VitalsId) VitalsValue
     | HandleSavedVitals (WebData ())
-    | SavePrenatalPhoto PersonId (Maybe PrenatalPhotoId) PhotoUrl
+    | SavePrenatalPhoto PersonId (Maybe PrenatalPhotoId) ImageUrl
```
AronNovak, line 197: same here, if we want to get rid of Photo, instead of Image, likely the function names should not be an exception.

client/src/elm/Backend/NutritionEncounter/Model.elm
AronNovak: Now I see, either the Image can be used as a Signature or as a Photo.

client/src/js/app.js
```diff
@@ -1253,43 +1263,59 @@ function attachSignaturePad() {
...
 
+        if (response.ok) {
+         var json = await response.json();
+         reportSignaturePadResult(json.url);
+        }
+        else {
+          // If something goes wrong while storing signature in cache,
+          // currently we do nothing.
+          // This situation is very rare, and if it does happen, user
+          // will most likely repeat the action.
```
AronNovak, line 1298: Why not `console.warn` the event?

client/src/js/app.js
```diff
@@ -1253,43 +1263,59 @@ function attachSignaturePad() {
...
+        }
+      }
+      catch (e) {
+        // Something was wrong with storing signature in cache.
+        // Take no action (for now).
```
AronNovak, line 1303: Why not `console.warn` the event?

AronNovak (approved):

client/src/js/app.js
anvmn: We don't use console.warn.
If we want to have deeper handlng, we need to notify Elm App and show proper warning.

Issue #698, opened 2023-05-02
NCDA - Aggregated Scorecard: Initial Elm App + hosting custom page on Drupal 7 [10h]


anvmn:
**Tasks:** 

1. Create new custom page on backend - `/ncda-aggregated-scorecard`.
This page will contain the Elm application.
As first stage, it will attach JS and CSS, and pass no data.
Access is alowed only for users with `administrator` role.

2. Create Inital ELM app (Elm v0.19.1). The app will present simple greeting message.

anvmn:
NCDA is on live for few months. Closing.

PR #699, opened 2023-05-02
[Feature] Complete NCDA

anvmn:
#698


anvmn:
Selection_1823

anvmn:
@AronNovak please review

server/elm/src/Pages/Scoreboard/Utils.elm
```diff
@@ -0,0 +1,5 @@
+module Pages.Scoreboard.Utils exposing (..)
+
+
+tmp =
+    0
```
AronNovak: What's the purpose of this? Do we use it somewhere?

server/hedley/modules/custom/hedley_aggregated_ncda/js/elm.js
```diff
@@ -0,0 +1,35 @@
...
+   * Adds Elm application.
+   */
+  Drupal.behaviors.ElmApps = {
+    attach: function (context, settings) {
+      var elmApps = settings.elm_apps;
+      // Iterate over the apps.
+      Object.keys(elmApps).forEach(function (appName) {
+        // appName with unique css ID, e.g. `elm-app--2`.
+        var node = document.getElementById(appName);
+        if (!node) {
```
AronNovak: is it on purpose that we do not use jQuery for this? Also `context` and `once`?

server/elm/elm-watch.sh
```diff
@@ -0,0 +1,35 @@
...
+  # --> all the errors you see are the current ones, not stale
+  clear;
+  tput reset;
+
+  echo -en "${DIM}";
+  date -R;
+  echo -en "${COLOR_OFF}";
+
+  elm make src/Main.elm --output ../hedley/modules/custom/hedley_aggregated_ncda/js/elm-main.js
+  # on Linux optionally prepend for better performance: sysconfcpus -n 1
```
AronNovak, line 24: We are able to conditionally invoke `sysconfcpus` if it's present.
Do you plan to use it inside DDEV or outside?

AronNovak:
```bash
if command -v sysconfcpus >/dev/null; then
    sysconfcpus -n 1 elm make src/Main.elm --output ../hedley/modules/custom/hedley_aggregated_ncda/js/elm-main.js
else
    elm make src/Main.elm --output ../hedley/modules/custom/hedley_aggregated_ncda/js/elm-main.js
fi

```

AronNovak: Perhaps we might want to install that tool in `.ddev/web-build/Dockerfile`

server/elm/src/Pages/Scoreboard/Utils.elm
anvmn: @AronNovak this PR lays infrastructure for the feature.
Right now, there're no utility funcitons, but very likely that we'll  need this file when developing the feature.
Elm does not allow empty file, so I put there something temporary for now.

server/hedley/modules/custom/hedley_aggregated_ncda/js/elm.js
anvmn: @AronNovak 

Whole Elm app was ported from WFP-INCE.
Checked several other projects where we attach Elm apps, and we have this code at all.
Do you see a problem with current implementation?

AronNovak: the only thing is that `attached` have the potential to be called multiple times during the lifetime of a page. Could it be problematic? If not, I do not see any issues.

server/elm/elm-watch.sh
anvmn: @AronNovak again, we got this from WFP-INCE.
On my local station current way of compiling works just fine.
It's very fast as well, so, I don't see a need for CPU optimizations right now.

server/hedley/modules/custom/hedley_aggregated_ncda/js/elm.js
anvmn: Never seen functionality issue here, so, I don't think this method causes any problems.

AronNovak (approved):

server/hedley/modules/custom/hedley_ncda/scripts/generate-data-for-all.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+/**
+ * @file
+ * Generates new revisions for all entities of certain content type.
```
anvmn:
```suggestion
 * Generates aggregated NCDA data for residents from all villages.
```

Issue #727, opened 2023-05-28
NCDA - Aggregated Scorecard: Universal Intervention - Immunisation [10h]


anvmn:
#713 Follow up.

Immunisation info:
![Selection_1892](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/91e60297-e4e8-44b3-a60d-b6760901d655)

PR #728, opened 2023-05-28
NCDA - Aggregated Scorecard: Universal Intervention pane - Immunisation

anvmn:
#727


anvmn:
Selection_1894

server/elm/src/Backend/Scoreboard/Model.elm
```diff
@@ -102,7 +103,46 @@ type alias UniversalInterventionData =
...
+type alias VaccinationValue =
+    { administeredDoses : EverySet VaccineDose
+    , administrationDates : EverySet NominalDate
+    }
+
+
+type VaccineType
+    = VaccineBCG
+    | VaccineOPV
+    | VaccineDTP
```
AronNovak, line 133: The list of vaccines seems to be a thing that will change in the future. Wouldn't it be better to make it editable?
The list of required vaccines typically change in every country from time to time.

anvmn: @AronNovak Every vaccine got it's own administration logic.
Number of doses, period between doses, and others, which are even more complex.
So, this whole logic is too complex to be configurable. Not impossible, but not worth the effort.

We may revise in future, if needed.

AronNovak (approved):

PR #732, opened 2023-05-31
Nutrition script by district

AronNovak:
https://github.com/Gizra/ihangane/issues/3046

## Status
```
 aaron   3046-nutrition-script-by-district  ~  gizra  eheza-app  ddev robo report:nutrition Dummy
 [Exec] Running cd /var/www/html/server/www &&
 drush scr profiles/hedley/modules/custom/hedley_admin/scripts/generate-nutrition-report.php --district=Dummy
28174 children with age below 6 years located in Dummy district.
```

### DIFF
--- a/server/RoboFile.php
+++ b/server/RoboFile.php
@@ -426,8 +426,13 @@ class RoboFile extends Tasks {
   /**
    * Generates the nutrition report.
    */
-  public function reportNutrition() {
-    $this->_exec('cd /var/www/html/server/www && drush scr profiles/hedley/modules/custom/hedley_admin/scripts/generate-nutrition-report.php');
+  public function reportNutrition($district = NULL) {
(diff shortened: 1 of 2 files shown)


anvmn (approved):
@AronNovak 

Per geofields hierarchy, top level is Province, followed by District.
Right now, there are 3 provinces, and total of 5 districts.
Since all districts names are different, this will work.
In threory, there can be multiple districts that will share same name, but I suppose we 
can address this latter, if required.

//cc @adamhstewart

AronNovak:
@anvmn Thanks, I leave it open until @adamhstewart decides if we should do the quickest way or not here.

adamhstewart:
@AronNovak By text Name is fine - the likelihood of two same-named districts is nil.

Issue #780, opened 2023-07-11
NCDA CHW Questionnaire - immunisation history


anvmn:
~~Todo: Define the architecture using which CHW will have a way to update vactinations thate were performed.~~

**Tasks:**

1. Turn Child Scoreboard encounter into multiactivity encounter (Encounter page with activities + Activity page for each activity), with 2 activites - Child Scorecard (displayed always) and Vaccination History (optional).
2. If child is behind on vaccinations, and responds that there are vaccines to record, display Vaccination History activity. It will have up to 7 tasks (for each of vaccines child is behind on).
![Selection_2065](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/5677d1ef-c962-46cd-afa0-b12a7c75c22f)

3. Vaccination History task form will behave similar to vaccination form at SPV encounter. Only difference is that it will not present a question about giving a shot today.
4. Vaccination History activity is to be accounted for at logic that generates vaccination progress data structure for child.

anvmn:
@adamhstewart ^^

adamhstewart:
@anvmn please check child Adeo Emmy on the nutrition environment. We did an encounter with him and update his vaccin records to be up to date. The scorecard shows that he is behind and there are no vaccines shown in the progress report.

anvmn:
@adamhstewart 

> @anvmn please check child Adeo Emmy on the nutrition environment. We did an encounter with him and update his vaccin records to be up to date. The scorecard shows that he is behind and there are no vaccines shown in the progress report.

Managed to reproduce the issue locally.
The problem here is that current implementation relies on existence of at least one SPV encounter.
If there's such an encounter, it's able t generate vaccination data from both sources - SPV and Child Scoreboard data.
However, if there're no SPV encounters, it assumes there were no vaccinations, which is not correct.

I'll modify the logic to take into account a scenario where there were no SPV encounters, but Child Scoreboard took place.

anvmn:
NCDA is on live for few months. Closing.

PR #782, opened 2023-07-12
NCDA for CHW - vaccination history activity

anvmn:
#780


anvmn:
@AronNovak This PR is big because it's about single activity that got 7 tasks of same type (differrent vaccinations), so it was not wort to break it into 7 PRs.
There're also some code relocations (to allow multiple modules to use same utility functions), so many files are different only  at the `import` section.

client/src/elm/Pages/ChildScoreboard/Activity/Model.elm
```diff
@@ -2,29 +2,91 @@ module Pages.ChildScoreboard.Activity.Model exposing (..)
...
+    , opvForm = emptyVaccinationForm
+    , pcv13Form = emptyVaccinationForm
+    , rotarixForm = emptyVaccinationForm
+    , activeTask = Nothing
     }
 
 
 type Msg
-    = SetActivePage Page
+    = NoOp
```
AronNovak, line 66: Where do we use this `NoOp`? It seems to be an outlier compared to the other messages. Do we really need it?

AronNovak: Ah, I see, is it temporary then?

AronNovak (approved):

client/src/elm/Pages/ChildScoreboard/Activity/Model.elm
anvmn: @AronNovak 

Well, no, it's not temporary. This is 'empty' message - similar to concept of `emptyNode`.
There are rare cases where we need the `NoOp` (no operation), as some of logical pathes should result in action while others are not. For example:

Selection_2100

PR #807, opened 2023-07-30
Sanitize patient name

anvmn:
https://github.com/Gizra/ihangane/issues/3060


anvmn:
Trivial change. Review is not mandatory. Will MWG.

Issue #850, opened 2023-08-23
Allow admins to monitor devices that may have a sync issue [4h]


anvmn:
**Task:** Create custom page that would show a table with devices that are suspected to have a sync problem.
Table columns: `Device Name`, `Changes to Upload`, `Last Sync Time`.
Access allowed for superuser and administrators.

anvmn:
Deployed to live

PR #851, opened 2023-08-23
Table to monitor devices that may have a sync issue

anvmn:
#850

### DIFF
--- a/server/hedley/modules/custom/hedley_admin/hedley_admin.module
+++ b/server/hedley/modules/custom/hedley_admin/hedley_admin.module
@@ -22,7 +22,84 @@ function hedley_admin_menu() {
     'access arguments' => ['hedley reporting admin'],
   ];
 
+  $items['admin/reports/devices-sync'] = array(
+    'title' => 'Devices sync state',
+    'description' => 'Presents devices that are suspected to have sync issue',
+    'page callback' => 'drupal_get_form',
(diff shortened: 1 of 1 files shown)


anvmn:
Peek 2023-08-24 19-38

AronNovak (approved):

server/hedley/modules/custom/hedley_admin/hedley_admin.module
```diff
@@ -308,3 +317,95 @@ function hedley_admin_health_center_access($account = FALSE) {
...
+  $query->groupBy('u.name');
+  $query->orderBy('last_sync', 'DESC');
+
+  $result = $query->execute();
+
+  $rows = [];
+  foreach ($result as $key => $row) {
+    $words = explode(' ', $row->name);
+    if (end($words) == 'Robot') {
+      array_splice($words, -1);
```
AronNovak, line 363: I am not sure why does it make sense to eliminate the "Robot" word, if it's for admin and super-admin, but surely it makes no harm.

anvmn: When device is paired, user is created for the device.
When created, user title is set to '$device_title Robot", to indicate that user was created automatically for the device.
So, "Robot" word is dropped to present the actual name of the device.

Issue #862, opened 2023-09-10
Improve code using elm-review [8h]


anvmn:
[Elm review](https://github.com/jfmengels/elm-review/tree/2.13.1) is helpfull tool for improving Elm code

anvmn:
Done

PR #864, opened 2023-09-11
Integrate elm-review 

anvmn:
#862


AronNovak:
@anvmn We should integrate it into CircleCI, right?

AronNovak (approved):

anvmn:
@AronNovak 

> @anvmn We should integrate it into CircleCI, right?

Well, we may at some point, but we better start on some other project, as here we get very large code base.
Also, with current setting, I get errors that I wouldn't consider as such. For example:
Selection_2256

The problem here is that this logic is part of `Simplify` rule, which suggests many good changes, but as far as I've seen, it's not possible to select what we want to aplly, and what not.
We can probably copy it to be a custom rule, and make the changes we want, but at current moment, I don't think it justifies the effort.

Another issue we got is that it also removes all unused code, which is not always correct, as we may want to save it for future use.

Issue #865, opened 2023-09-12
Kirundi Translations


Diama1:
### Todo
Add Kirundi translations into the App.

anvmn:
Deployed to live. Closing.

PR #875, opened 2023-09-14
Kirundi Translations

Diama1:
#865


adamhstewart:
@anvmn what is this test and why does it fail? ^^

anvmn:
@adamhstewart 

> @anvmn what is this test and why does it fail? ^^

It's simpletets, and very often, it gets stuck.

Selection_2264

If we rerun it, it usually passes.

Issue #890, opened 2023-09-27
Resolve slowness on registation form for CHW, at Rwanda site [3h]


anvmn:
Currently, it take several seconds to see a response for click on form.

anvmn:
Deployed to live. Closing.

PR #894, opened 2023-09-27
'Cahce' Geo structure to avoid thier constant recalculations (which causes slowness)

anvmn:
#890


client/src/elm/SyncManager/Model.elm
```diff
@@ -349,6 +350,11 @@ type alias Model =
     -- `idle` - 50; which is the minimum we will allow.
     -- `sync` - 10000. The means that sync will sit idle for 10 seconds.
     , syncSpeed : Editable SyncSpeed
+
+    -- We genereate and store Geo structure, to avoid pereated recalculations
```
anvmn:
```suggestion
    -- We genereate and store Geo structure, to avoid repeated recalculations
```

AronNovak (approved):

Issue #901, opened 2023-10-04
Corrupted data saved at IndexedDB while storing Prenatal Specialty Care [2h]


anvmn:
Corrupted data at IndexedDB  as a result of Save operation was seen twice, and both times it was the PersonId of Prenatal Specialty Care content type: 

- https://github.com/Gizra/ihangane/issues/3064
- https://github.com/Gizra/ihangane/issues/3079

**Task:**  investigate what may be a root cause for this error at a specific CT.

anvmn:
For the context:

Overall, at live DB we have 23 nodes of type Prenatal Specialty Care (created since October 2022).
2 of them had corruptions.

Prenatal Specialty Care, **as any other prenatal measurement** got standard method for performing save:

`SaveSpecialityCare PersonId (Maybe ( PrenatalSpecialityCareId, PrenatalSpecialityCare ))` at `Pages.Prenatal.Activity.Model`, and at `Pages.Prenatal.Activity.Update`:

```
        SaveSpecialityCare personId saved ->
            let
                measurementId =
                    Maybe.map Tuple.first saved

                measurement =
                    getMeasurementValueFunc saved

                appMsgs =
                    model.specialityCareData.form
                        |> toSpecialityCareValueWithDefault measurement
                        |> unwrap
                            []
                            (\value ->
                                [ Backend.PrenatalEncounter.Model.SaveSpecialityCare personId measurementId value
                                    |> Backend.Model.MsgPrenatalEncounter id
                                    |> App.Model.MsgIndexedDb
                                , App.Model.SetActivePage <| UserPage <| PrenatalEncounterPage id
                                ]
                            )
            in
            ( model
            , Cmd.none
            , appMsgs
            )
```

**Note that corruption is at `personId`.**


At `Backend.PrenatalEncounter.Update`:

```
        SaveSpecialityCare personId valueId value ->
            ( { model | saveSpecialityCare = Loading }
            , saveMeasurementCmd currentDate encounterId personId nurseId healthCenterId valueId value prenatalSpecialityCareEndpoint HandleSavedSpecialityCare
            , []
            )
```

anvmn:
2 things to note:

1. Prenatal Specialty Care save code is last of all measurements. Theoretically, this should not be a problem.
2. At `Pages.Prenatal.Activity.View`, we got following invokation:

```
        , viewSaveAction language
            (SaveSpecialityCare assembled.participant.person assembled.measurements.specialityCare)
[...]

PR #902, opened 2023-10-04
Avoid having data corruption at IndexedDB while storing Prenatal Specialty Care

anvmn:
#901


anvmn:
Trivial change. Merging.

Issue #926, opened 2023-10-25
Localization - Burundi - DTP - HepB - Hib is called Pentavalent, extra dose of DTP [8h]


adamhstewart:
## Issue
Following #920 

In Burundi context, the combo vaccine for DTP - HepB - Hib is named Pentavalent

In addition there is an extra (4th dose of DTP only) at 18 months

## Tasks
- Rename the combo vaccine in Burundi context (keep the icon)
- Add new vaccine called DTP for one dose at 18 months. It should only appear after 3 doses of Pentavalent and with a minimum 28 day interval from the last Pentavalent dose.

anvmn:
@adamhstewart 

> In Burundi context, the combo vaccine for DTP - HepB - Hib is named Pentavalent

Ok, that makes sence. These are the details for Pentavalent (5 diseases):

![Selection_2393](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/b872693a-5c03-4eda-9a39-648e1021f76d)

Looking at current implementation however, seems that `Hib` disease is missing:

![Selection_2394](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/1ecb08fb-6a9a-4190-b73e-85837a4007f6)

Why is that? Should we fix it?

> In addition there is an extra (4th dose of DTP only) at 18 months

I suppose that DTC4 from #920 is the (4th dose of DTP only).
This is new type of vaccine we'll have to introduce. I suppose it handles 3 diseases - Diptheria, Tetanus, Pertusis. I'll addapt the forms accordingly.
Note that we need to add it for 2 types of encounters - Child Scorecard and SPV.
So, only the activities infra will require ~6h. In total, I suppose it will at least require 12h.

adamhstewart:
> Why is that? Should we fix it?

I'll ask - but let's not deal with that here.

> I suppose that DTC4 from https://github.com/TIP-Global-Health/eheza-app/issues/920 is the (4th dose of DTP only).

Yes, it's a Typo - it should be DTP (the 4 is for the fourth dose , as DTP is included in Pentavalent).

> So, only the activities infra will require ~6h. In total, I suppose it will at least require 12h.

OK to bump

balagan73:
https://qa-vhw.pantheonsite.io/app/

## QA Review

### Found issues:

1. Typo in Diphteria
[...]

Issue #920, opened 2023-10-24
Regionalization - Burundi Immunization Calendar [12h]


adamhstewart:
## Issue

Adapt immunization schedule for Burundi:

BCG/HepBo/Polio: at birth
Polio1, Pentavale1, Rotavirus1, PCV13(Pneumonia):6 weeks
Polio2, Pentavale2, Rotavirus2, PCV13(Pneumonia):10 weeks
Polio3, Pentavalent 3, PCV13(Pneumonia) and IPV:14 weeks
Vitamin A Every 6 months: During campaign
Measle/Rubeola and mosquito net:9 months
VitA and albendazole: 12 months
DTC4, MR, vitA, Albendazole and mosquito net: 18 months 
VitA and Albendazole every 6 months until 59 months

NO HPV immunization

## Task
- Adapt schedule

# Immunisations:

Age | Immunisation | Notes
--|--|--
Birth |BCG Bacilius Calmette - Guérin (BCG)
|| Oral Polio Vaccine
6 Weeks (1.5 months) |Oral Polio Vaccine
||Pentavale
||Pneumococcal Vaccine (PCV 13)
||Rotavirus (Rotarix)
10 Weeks (2.5 months) |Oral Polio Vaccine
||Pentavale
||Pneumococcal Vaccine (PCV 13)
||Rotavirus (Rotarix)
14 Weeks (3.5 months) |Oral Polio Vaccine
||Pentavale
||Pneumococcal Vaccine (PCV 13)
||Inactivated Polio Vaccine (IPV)
36 Weeks (9 Months) |Measles - Rubella (MR)
18 Months (1.5 years) |Measles - Rubella (MR)
||DTP (4-th dose)

Vaccine | Dose | Interval
--|--|--
BCG Bacilius Calmette - Guérin (BCG) | 1 | 
Oral Polio Vaccine | 4 | 28 days min
Pentavale | 3 | 28 days min
Pneumococcal Vaccine (PCV 13) | 3 | 28 days min
Rotavirus (Rotarix) | 2 | 28 days min
Inactivated Polio Vaccine (IPV) | 1 |
Measles - Rubella (MR) | 2 | 9 months
DTP (4-th dose) |  1 |

anvmn:
@adamhstewart 

This issue mixes different subjects together.
It should be divided into 3 - Immunization, Medication and Resources (mosquito net).

**Immunization**:
[...]

PR #927, opened 2023-10-25
Regionalization - Burundi Immunisation changes

anvmn:
#926 
#920


anvmn:
Selection_2405

anvmn:
Selection_2407

AronNovak (approved):

server/RoboFile.php
```diff
@@ -43,7 +43,7 @@ public function deployPantheon($branchName = 'master') {
 
     if ($result->getMessage()) {
       $this->yell($result->getMessage());
-      throw new Exception('The GitHub working directory is dirty. Please commit any pending changes or add to .gitignore.');
+      // throw new Exception('The GitHub working directory is dirty. Please commit any pending changes or add to .gitignore.');
```
AronNovak: May I ask what was the problem here? Normally it's a safeguard that helps us to prevent broken deployments from local machines.
On Travis, sometimes we need to tweak gitignore for that.

anvmn: @AronNovak 

I will revert this change. The problem is with `.gitignore`:

Selection_2418

Issue #935, opened 2023-11-02
Installation failure - File og-7.x-2.x-dev.tar.gz_date=1698701047 is corrupt (wrong md5 checksum) [0.5h]


anvmn:
![Selection_2420](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/8b2c6847-ab05-4ed2-9887-25106150115d)

anvmn:
New OG version was issued a couple of days ago, so we'll make it official for installation.

![Selection_2421](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/ff98b4b7-706b-40f1-aabe-d2580d53093d)

PR #936, opened 2023-11-02
FIx installation failure - set proper version for OG module

anvmn:
#935


anvmn:
Trivial change. Merging.

Issue #945, opened 2023-11-12
Display Immunization Followups in Casemanagment for CHW [8h]


adamhstewart:
## Issue
<img width="656" alt="E-Heza_System" src="https://github.com/TIP-Global-Health/eheza-app/assets/688507/100327dd-502c-43eb-b994-dea5d89e20a9">

## Tasks
- Create new section for Immunization
- Show followups that are past the date for the immunization visit
    - for first 30 days past the visit date, the tags should read "this month"
    - for after 30 days past the visit date, the tag should read "past due"
    - Text should read "Check for immunization."
    - link to a new CHW SPV encounter

anvmn:
@adamhstewart 

This issue is missing the 'reminder resolution' part.
For example, starting new SPV for CHW encounter, should resolve reminders from previous encounters.
Also, maybe same should be done if nurse runs standard SPV?

adamhstewart:
> This issue is missing the 'reminder resolution' part.

It should only be resolved if:
- the patient visits the HC AND receives the required vaccines; OR
- the patient receives follow up from the CHW AND the required vaccines are recorded.

anvmn:
@adamhstewart 

> > This issue is missing the 'reminder resolution' part.
> 
> It should only be resolved if:
> 
> * the patient visits the HC AND receives the required vaccines; OR
> * the patient receives follow up from the CHW AND the required vaccines are recorded.

In both cases, SPV encounter is started, where Next Visit activity is mandatory (in case future vaccinations are required).
So, each such encounter creates new 'follow up', which means we can resolve all previous ones.
Does that make sense?

adamhstewart:
@anvmn ^^ above makes sense, yes.

balagan73:
https://nutrition-ihangane.pantheonsite.io/app/

## QA Review

### Found issues and questions:

1 When a patient is behind with immunization, 
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/2d65271e-aea1-4ebf-bbdf-411edb37c062)

souldn't it be indicated under case management as Past due similar to other follow ups? I am not sure how this is checked. I have adjusted the date back on the Well Child followup, but maybe this way some cron task were skipped or something. I need help, how to test it. @anvmn @adamhstewart 
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/14aa11de-0ce5-42d7-9f9b-8f8ddec28de0)


[...]

Issue #944, opened 2023-11-12
Create an Immunization Followup Content Type [8h]


adamhstewart:
## Issue
A content type to hold the follow-up date that a child should be rechecked for immunization by a CHW.

In this case, the follow-up option should be a date that is equal to the date in the SPV that indicates when a child's next immunization visit.  should have been

## Tasks
- Create the content type
- Create infrastructure to sync with frontend app.

anvmn:
@adamhstewart 

We have Next visit content type, so I'd first consider expanding it for our needs.

adamhstewart:
> We have Next visit content type, so I'd first consider expanding it for our needs.

sure, if it makes most sense.

balagan73:
https://nutrition-ihangane.pantheonsite.io/app

## QA Review

### Found issues:
- None

### Checked and approved
- Date of next immunization visit is stored and displayed:

![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/4d192d0e-2e6a-4be9-873b-853c4702dce5)

Issue #943, opened 2023-11-12
Add Followup notification for Immunization Case Managemengt - CHW [4h]


adamhstewart:
## Issue 

Allow a CHW to indicate that there should be a follow-up created for the next time a patient is supposed to receive vaccination.

<img width="1031" alt="Cursor_and_E-Heza_System" src="https://github.com/TIP-Global-Health/eheza-app/assets/688507/cd36a9eb-d415-4a4c-bcc2-a9917c5b9990">

## Task
- Create the checkbox boolean field on the form

anvmn:
@adamhstewart 

Note that Next Visit is a Next Steps task.
Currently, Next Visit is not displayed for CHW. I suppose that will change after https://github.com/TIP-Global-Health/eheza-app/issues/934#issuecomment-1807798825 is answered.

Also, it's not clear to me why CHW should indicate anything with checkbox. Don't we want the case management always?

adamhstewart:
> Don't we want the case management always?

confirmed we will automate this when the encounter is ended.

balagan73:
This will be checked under https://github.com/TIP-Global-Health/eheza-app/issues/945

PR #947, opened 2023-11-20
SPV for CHW - follow ups at Case Management

anvmn:
#943
#944
#945

### DIFF
--- a/client/src/elm/Backend/Fetch.elm
+++ b/client/src/elm/Backend/Fetch.elm
@@ -275,6 +275,13 @@ shouldFetch currentTime model msg =
                 |> Maybe.withDefault NotAsked
                 |> isNotAsked
 
+        FetchWellChildEncountersForParticipants ids ->
+            if List.isEmpty ids then
+                False
+
+            else
+                List.any (\id -> not (Dict.member id model.wellChildEncountersByParticipant)) ids
+
         FetchWellChildMeasurements id ->
             Dict.get id model.wellChildMeasurements
                 |> Maybe.withDefault NotAsked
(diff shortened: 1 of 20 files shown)


client/src/elm/Backend/Fetch.elm
```diff
@@ -275,6 +275,13 @@ shouldFetch currentTime model msg =
                 |> Maybe.withDefault NotAsked
                 |> isNotAsked
 
+        FetchWellChildEncountersForParticipants ids ->
+            if List.isEmpty ids then
+                False
```
AronNovak, line 280: should not we return an empty list in this case?

server/hedley/modules/custom/hedley_case_management/hedley_case_management.module
```diff
@@ -256,6 +261,66 @@ function hedley_case_management_resolve_prenatal_follow_ups($person_id, $to_nid
...
+    // than 2 months old.
+    // This type of exam does not have Next Visit activity, so
+    // it should not resolve immunization follow-ups.
+    return;
+  }
+
+  $person_id = $wrapper->field_individual_participant->field_person->getIdentifier();
+
+  if (empty($person_id)) {
+    // Failed to load required data.
```
AronNovak, line 285: shall we log it as an error somewhere?

client/src/elm/Backend/Fetch.elm
anvmn: @AronNovak 

It's `shouldFetch` function, and it checks if issued fetch command really needs to be executed, as it's possible that we have the data already in Model. So, it returns `Bool`.

server/hedley/modules/custom/hedley_case_management/hedley_case_management.module
anvmn: Well, no, I don't see a need for that.
We have numerous places throughout the code with similar checks.
The idea is to prevent trying to access properties on null (which should not happen, as we should have the data always), etc.

PR #950, opened 2023-11-22
Regionalization, Burundi - MUAC in mm fix

### LINKED ISSUE #916: Regionalization - Burundi: Unit of measurement for MUAC [4h]
## Issue 

In Burundi MUAC measurements are taken in mm rather than cm. This is true in case of both children and pregnant women

## Task
- Allow input to be taken in mm (also indicate mm on the interface) and store the value in cm
- Show previous measurements in mm

Note: as far as I can see, MUAC is not displayed in raw numbers on the progress report or on the child scorecard.

anvmn:
#916


anvmn:
Selection_2470
Selection_2469

AronNovak (approved):

PR #965, opened 2023-11-29
Fake travis_wait for simpletest

AronNovak:


anvmn:
@AronNovak should we increase the timeout here?

AronNovak:
@anvmn Yep, good point, sorry that this PR got lost essentially.

anvmn (approved):

Issue #989, opened 2023-12-31
Allow Lab Tecnicians to enter labs for "paused" encounter [4h]


adamhstewart:
## Issue

When a nurse indicates a test was able to be performed that day a case management followup should be created that indicates that the lab has been ordered. The lab technician should be able to click on the patient's name and open an encounter to enter the missing lab results.

<img width="757" alt="E-Heza_System" src="https://github.com/TIP-Global-Health/eheza-app/assets/688507/c0b761ba-d36b-4d61-aa3d-cd041c3e2023">

## Tasks
- Allow lab tech to be able to access the second part of an ANC encounter for a patient who is awaiting lab results.
- Change the lab of the "pending" labs to "ordered"

anvmn:
@adamhstewart 

Would it be correct to say that 'ordered' entries should be seen only by Lab Techs, and nurses should see an entry only when Lab tech has filled all of the results?

adamhstewart:
> Would it be correct to say that 'ordered' entries should be seen only by Lab Techs, and nurses should see an entry only when Lab tech has filled all of the results?

No, because we don't want to break the existing workflow. The nurse should be able to see AND perform the operation.

anvmn:
@adamhstewart 

> Change the lab of the "pending" labs to "ordered"

When we create an entry, it has expiration period of 35 days.
The entry says 'Pending' until 7 days before expiration.
After, it says 'Closing soon'.

So, for lab tech:
'Pending' => 'Ordered'
Closing soon' => ?

adamhstewart:
@anvmn i think also “closing soon”

balagan73:
## QA Review

### Found issues
- None
 
### Checked and approved:
1. Lab Technician sees Ordered items for paused encounters, where they need to enter lab results
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/7589ed99-b971-4ca2-be09-da7047ae9516)

2. When changing concluded date on the Lab results, Closing soon pill displays.
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/2cc11a31-91b3-4478-9daf-ab2ba0140bf1)

Issue #988, opened 2023-12-31
Create Lab Technician user type [6h]


adamhstewart:
## Issue
A pregnant woman comes to the lab and I look for her name in the Lab case management section. I see that labs were ordered for her (“ordered”). I open case management and enter all lab results that I performed or I document if labs were not performed. I complete the section and receive a review of lab results similar to the progress report (like IZs). I review and accept this. 

## Tasks
- Create a user (nurse) type that is a Lab Technician, this user type should be able to see only the case management section and the lab sections with in it.
- Create the entry screen for a lab tech - should contain only device status and case management.

balagan73:
https://nutrition-ihangane.pantheonsite.io/app/#device

## QA Review

### Found issues

### Checked and approved

Nurse created with Lab Technician role. Entry screen contains only the Case management and Device Status activities. Under case management only the ANC Labs display. 

Entry screen|Case management
-|-
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/f07df5be-61df-4d2d-ac1b-3a3acb0bd38b)|![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/c589aac2-cdb8-40c2-b52c-13cf64a54988)

PR #991, opened 2024-01-08
[Feature] Lab technician

anvmn:
#988 
#989


anvmn:
Peek 2024-01-08 21-55

Issue #1002, opened 2024-01-24
TB - Medications


Diama1:
## Issue
Medication will have 3 sub activities.

Note: Prescribed medications only appears in the first TB Management encountert

Prescribed Medication | DOT | Treatment  Review
--|--|--
![Screenshot from 2024-01-24 16-24-40](https://github.com/TIP-Global-Health/eheza-app/assets/32199791/af3d75b8-f3ea-4b89-8dbb-55d8cbbd20ec) | ![Screenshot from 2024-01-24 16-39-15](https://github.com/TIP-Global-Health/eheza-app/assets/32199791/7db53b90-9d01-47a4-94c1-c4f5706541b8) | ![Screenshot from 2024-01-24 16-39-55](https://github.com/TIP-Global-Health/eheza-app/assets/32199791/d8f6a2d6-aca7-40d5-86f6-d77e97b0413c)

For the DOT activity if the answer is No this will be a sub question for both the :
Why not:
- meds already taken today
- [not indicated] only for "Will you provide DOT/TDO today?"
- med unavailable
- advised to hold b/c of side effects
- patient refused

## Todo

- Create the activities per design

anvmn:
@adamhstewart 

**Prescribed Medication**:
 - Should we allow multiple selection?
 - Third option looks strange to me. Can you confirm the wording please?
 - Shouldn't we add another option for 'no medication prescribed'? If we do, how would this affect the DOT activity?	

**DOT/DTO**:
- Issue description says something about additional question if No answer is given. It's not clear for which question is it (description says both but it does not make sense to me).



Also, I'll need the **icons for the tasks**.

Diama1:
@anvmn 
> **DOT/DTO**:
> 
> * Issue description says something about additional question if No answer is given. It's not clear for which question is it (description says both but it does not make sense to me).

The additional question will come when we say No to both of the questions. But the difference will be only for "Will you provide DOT/TDO today?" we will add [not indicated] .
For example: for the first question 
Why not?
- meds already taken today
- not indicated
- med unavailable
- advised to hold b/c of side effects
- patient refused

for the second we remove  [not indicated] option

adamhstewart:
> Should we allow multiple selection?

Yes

> Third option looks strange to me. Can you confirm the wording please?

Wording is correct

> Shouldn't we add another option for 'no medication prescribed'? If we do, how would this affect the DOT activity?

Medication is always prescribed for TB. The "exit ramp" here is "other"

[...]

PR #1015, opened 2024-02-12
Tuberculosis Management - Medication

anvmn:
#1002


anvmn:
Peek 2024-02-13 15-09

client/src/elm/Measurement/Model.elm
```diff
@@ -1868,3 +1868,32 @@ type alias NutritionCaringForm =
...
+    { takenAsPrescribed = Nothing
+    , missedDoses = Nothing
+    , feelingBetter = Nothing
+    , sideEffects = Nothing
+    , reasonForNotTaking = Nothing
+    , reasonForNotTakingDirty = False
+    , totalMissedDoses = Nothing
+    , totalMissedDosesDirty = False
+    , adverseEvents = Nothing
+    , adverseEventsDirty = False
```
AronNovak, line 1898: How does this `Dirty` concept work? As I see, empty data is not dirty, upon input, it becomes dirty? But does this become non-dirty afterwards at some point?

AronNovak (approved):

client/src/elm/Measurement/Model.elm
anvmn: @AronNovak 

We have following situation:
Form is the UI, and on refresh all it's fields loose set values.
When we store the form, Value is created - representation of stored data for back-end.
To allow editing saved form, we need to combine form with Value, and present the data.
For correct presentation we have `form <=> value` functions.

Dirty fields are used when we have a conditional input, which is edited and needs to get empty value.
For example:
1. Question: 'Did you administer medication?'
2. If 1 answer is No, we ask for the reason and list several options.

So, we have a form with fields for q1 and q2.
On first run, user answers q1 No, and fills the reason for Q2.
On save, value is created with 2 fields - v1 and v2.
Now page is refreshed, and user decides to edit the answer to q1 - enters Yes.
If we do nothing, on save `form => value` function will result in same v2 we had previously, while it needs to be empty.
To overcome this, we use dirty field for q2. So, when q1 is set to Yes, we set q2 to empty value and q2Dirty set to true, which will indicate for `form => value` function to force use of q2 from form overriding saved v2.

PR #1054, opened 2024-03-05
Aggregated NCDA - additional tasks

anvmn:
https://github.com/Gizra/ihangane/issues/3107

Per https://github.com/Gizra/ihangane/issues/3107#issuecomment-1980364044


anvmn:
###Before:
Selection_266

Selection_267


###After:

Selection_270
Selection_271

Issue #1055, opened 2024-03-06
TB Management - Add AI Diagnoses TB Suspect [4h]


adamhstewart:
## Issue
A cough of two weeks or longer is considered a TB Suspect, and it overrides any diagnosis.

The popup for diagnosis should read Tuberculosis Suspect:
<img width="844" alt="Cursor_and_E-Heza_System" src="https://github.com/TIP-Global-Health/eheza-app/assets/688507/275c8346-77cb-4a50-94d3-81430eb025e2">

Next steps items are:
- Refer to health center
- Followup

anvmn:
@adamhstewart 

If I remember this right, COVID was most urgent diagnosis, where we would show next steps right after diagnosis was made.
Other diagnoses required 3 or 4 activities to be completed, to resolve the diagnosis.

How should we treat DB? Given more than 2 weeks cough, make the diagnosis right away + show next steps?

adamhstewart:
> How should we treat DB? Given more than 2 weeks cough, make the diagnosis right away + show next steps?

We should complete the full activities - For COVID it was a matter of isolation. Here it's not a concern.

anvmn:
@adamhstewart 

> > How should we treat DB? Given more than 2 weeks cough, make the diagnosis right away + show next steps?
> 
> We should complete the full activities - For COVID it was a matter of isolation. Here it's not a concern.

In this case, it can override all diagnoses except for COVID.

anvmn:
@adamhstewart 

> Next steps items are:
> - Refer to health center
> - Followup 

We have AI both for nurse and CHW.
If it's CHW that's running the encounter, it makes sense to refer to HC.
What do we do for nurse? Refer to hospital?
Or maybe, TB suspect should be determined only by CHW?

anvmn:
@adamhstewart 

> Next steps items are:
> - Refer to health center
> - Followup 

For the context, at AI first encounter we determine (possible) diagnose.
Subsequent encounter will be available (in other words illness will remain open) is diagnose was made.
Subsequent encounter is used to track illness progress and reaction for treatment.
As a result, next steps at these encounters are different.
For example, at subsequent encounter, we send to HC if there was no improvement.

So the question is what should be next steps at subsequent encounter?

adamhstewart:
> What do we do for nurse? Refer to hospital?

[...]

Issue #1051, opened 2024-03-04
TB Management - Update cough duration [6h]


Diama1:
## Issue

As for TB management, an individual becomes a TB suspect if, through acute illness, cough persists as a symptom for a period of 2 weeks or more. 

As such we need to change the cough duration field to reflect a diagnosis for TB:
~~Less than 2 weeks~~
~~2 Weeks or more~~
- 2 Weeks or less
- More than 2 weeks

## Task
- Add the new field and deprecate the old field
- Create infrastructure to handle the new field
- Update the UI on the forntend

anvmn:
@adamhstewart what do we do about progress report then?

anvmn:
@adamhstewart 

> @adamhstewart what do we do about progress report then?

This is what I've implemented, so we can make use existing if fields and get all needed info:


![Peek 2024-03-11 11-57](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/57e39caa-4b10-498d-b99e-062bde383069)

Will this do?

adamhstewart:
@anvmn ^^ Unfortunately, no - we need to convert to the two options as described, I assume this means a new field and infrastructure.

For the purposes of the progress report, we can assume less than 2 weeks to be 7 days and  more than 2 weeks to be 15 days.

anvmn:
@adamhstewart 

> For the purposes of the progress report, we can assume less than 2 weeks to be 7 days and more than 2 weeks to be 15 days.

It's 2 weeks or more, so I'll make it 14. This way we maintain consistency with max duration of other signs.

anvmn:
As discussed at catch up meeting:

Periods are:

- 2 Weeks or less
- More than 2 weeks

anvmn:
@adamhstewart deployed to Nutrition env

balagan73:
### QA Review

## Found issues and questions

- None

## Checked and approved

1. Options on UI are updated
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/4b731b76-0edf-4149-972c-96346b638454)

PR #1063, opened 2024-03-07
TB Management -  Tuberculosis Suspect diagnosis at Acute Illness

anvmn:
#1051
#1055

### DIFF
--- a/client/src/elm/Translate.elm
+++ b/client/src/elm/Translate.elm
@@ -1702,6 +1702,7 @@ type TranslationId
     | TuberculosisProvideDOTTodayQuestion
     | TuberculosisReasonNotProvidedToday TuberculosisDOTSign
     | TuberculosisReasonMedicationsNotDistributed TuberculosisDOTSign
+    | TuberculosisSuspect
     | TuberculosisSymptomQuestion TuberculosisSymptom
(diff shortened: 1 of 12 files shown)


anvmn:
Peek 2024-03-11 23-52

client/src/elm/Translate.elm
```diff
@@ -21162,6 +21183,12 @@ translationSet trans =
         TuberculosisReasonMedicationsNotDistributed reason ->
             translationSet <| TuberculosisReasonNotProvidedToday reason
 
+        TuberculosisSuspect ->
+            { english = "Tuberculosis Suspect"
+            , kinyarwanda = Nothing
+            , kirundi = Just "Nta Gupima/gusuzuma"
```
AronNovak: This is not the equivalent of the English string. This means: `No Testing/Assessment`

anvmn:
```suggestion
            , kirundi = Nothing
```

AronNovak (approved):

client/src/elm/Translate.elm
anvmn: Thanks @AronNovak. Looks like a result of copy/paste and forgetting to replace values with Nothing.

Issue #1052, opened 2024-03-04
TB Management - Case management for TB Suspect [4h]


Diama1:
## Issue

After the individual has been found as a TB Suspect through Acute illness, a case management followup should be created in TB Management rather than in Acute Illness.

![Screenshot from 2024-03-04 22-43-52](https://github.com/TIP-Global-Health/eheza-app/assets/32199791/5e5696d1-fe7f-4b8d-84d0-f803e5c634a9)

## Task
- Create the Case management followup

anvmn:
@adamhstewart deployed to Nutrition env

balagan73:
### QA Review

## Found issues and questions

- None

## Checked and approved

1. Case management follow up is properly created under Tuberculosis
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/4f897362-4c04-4a97-bf23-64881a8ae533)

PR #1067, opened 2024-03-13
TB Management - Case management, TB Suspect adjustments

anvmn:
#1052


client/src/elm/Pages/GlobalCaseManagement/Utils.elm
```diff
@@ -232,13 +237,67 @@ generateImmunizationFollowUps limitDate followUps =
...
+One is Acute illness follow up, in case Tuberculosis suspect is diagnosed.
+Other is the follow ups from Tuberculosis management encounter.
+So, in order to combine these 2, we'll have to include AI follows for person,
+as part of Tuberculosis follow ups.
+Currently, all Tuberculosis encounters are derived from single participant (as
+there's no option to 'end' Tuberculosis illness), so this solution is valid.
+In case we'll have multiple Tuberculosis illnesses in future, this wil need to
+be revised.
+There's also an option that patient did not attend Tuberculosis encounter, and
+there's no participant. This is why we return second dictionary derrived from
```
AronNovak:
```suggestion
there's no participant. This is why we return second dictionary derived from
```

server/hedley/modules/custom/hedley_case_management/hedley_case_management.module
```diff
@@ -361,14 +368,17 @@ function hedley_case_management_handle_tuberculosis_insert_trigger($node) {
...
+ *
  * @param int $person_id
  *   Person ID.
  * @param int $to_nid
  *   Optional; IDs of loaded follow-ups should be below this node ID.
  * @param int $limit
  *   Maximal number of follow-ups to load. Defaults to 200.
  */
-function hedley_case_management_resolve_tuberculosis_follow_ups($person_id, $to_nid = NULL, $limit = 200) {
+function hedley_case_management_resolve_tuberculosis_follow_ups($person_id, $to_nid = NULL, $limit = 1000) {
```
AronNovak: Why did we need to bump the limit from 200 to 1000? Were there such cases with more than 200 follow-ups? It sounds suspicious at first sight.

anvmn: Well, no. Person should not have more than 200 follow  ups, specially, that follow ups are getting resolved, and excluded from query results. So I'll revert the change.

anvmn:
```suggestion
function hedley_case_management_resolve_tuberculosis_follow_ups($person_id, $to_nid = NULL, $limit = 200) {
```

AronNovak (approved):

PR #1070, opened 2024-03-14
Duplicates report

adamhstewart:
This enables a report of likely duplicate records. It identifies likely duplicates by matching birthdates within a village (while stripping out birthdates on XXXX-01-01 (as this is the date that is used for estimated birthdates).


AronNovak (approved):

AronNovak:
Do we need to include this in the Jenkins-based reports?

adamhstewart:
> Do we need to include this in the Jenkins-based reports?

@AronNovak  Not necessary, thanks.

Issue #1089, opened 2024-03-28
HIV Management - Encounter Infrastructure [6h]


anvmn:
Tasks:
- Create infrastructure for Participant and Encounter pages
- Start encounter logic

anvmn:
@adamhstewart

Deployed to Nutrition env. Ready for QA.

balagan73:
## QA Review

### Found issues and questions:


### Checked and approved:

1. HIV encounter is implemented
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/e55530c8-ee6d-42f7-b401-43a8b1892b3d)

balagan73:
@anvmn What are the Participant pages? I have checked the participant directory, but did not see any HIV related data.

anvmn:
@balagan73 

> @anvmn What are the Participant pages? I have checked the participant directory, but did not see any HIV related data.

Participant page is the one you added at screenshot.
That's the page where we have decision logic of to start encounter and create new participant, or subsequent encounter for existing participant.

PR #1090, opened 2024-03-28
HIV management - encounter infra

anvmn:
#1089


anvmn:
Peek 2024-03-28 22-41

Issue #1112, opened 2024-04-13
HIV Management - Case Management [10h]


adamhstewart:
### Creating a followup
- In a HIV Management encounter, the CHW has the option for follow-up in one day, one week, or one month. Any option should create a follow-up in a new "TB Management" section of the Case Management screen.

- In any place where there is an HIV Lab test (e.g. ANC Labs) that gives a positive result a followup should be scheduled for one week from the positive test.

Logic for display on the case management "pills" should follow AI case management, that is 
- This week: follow-up due within 7 days
- This month: follow-up due in 7 days or more
- Due today: follow-up due today.
- Past due: followup was due in the past

Resolution of a case management follow-up should occur if the patient has an HIV Management encounter when that follow-up is active (whether or not they entered through the case management screen). That is, any case management encounter should clear previous follow-ups, and a patient can only have one HIV Management follow-up at a time.

## Tasks
- Create the follow-up infrastructure
- Create the logic for the creation of a follow-up and its resolution in both HIV management and in Lab Results
- Render the page for HIV Management on the Case Management screen

anvmn:
@adamhstewart

Deployed to Nutrition env. Ready for QA.

balagan73:
https://nutrition-ihangane.pantheonsite.io/app

## QA Review

### Found issues:

- None

### Checked and approved:

1. Status of follow up displays correctly:

Past due|Due today
-|-
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/bb10616d-2dd4-4827-b487-75317b00c551)|![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/8a992eb5-a071-450a-9854-3be041dd8281)

This week|This month
-|-
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/b581bfaf-4002-4e51-a037-8f9a145a47d0)|![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/c7d45f6d-2857-42c5-8cb6-29522c59aee9)

PR #1113, opened 2024-04-14
HIV Management - Case management 

anvmn:
#1112


anvmn:
Peek 2024-04-15 23-35

AronNovak (approved):

Issue #1117, opened 2024-04-21
DDEV installation fails [1h]


anvmn:
![Selection_427](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/52a5cf68-c18e-422f-90d3-b0f01d0a82a1)

anvmn:
**Analysis:**

Failure is cased by error when trying to install Python 2.7.
Reason for this is Python 2 becoming obsolete at DDEV 1.23.0:

![Selection_433](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/9e860078-597f-4241-9328-dc62a2c969c5)


Tried going by with Python 3 (#1118), but this created dependency issues when running `ddev gulp`.

To be able to proceed development, downgraded DDEV to 1.22.7 locally.

anvmn:
Deployed to live

PR #1118, opened 2024-04-21
WIP: Fix DDEV installation failure

anvmn:
#1117

### DIFF
--- a/.ddev/web-build/Dockerfile
+++ b/.ddev/web-build/Dockerfile
@@ -1,8 +1,7 @@
 ARG BASE_IMAGE
 FROM $BASE_IMAGE
 
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confold" --no-install-recommends --no-install-suggests inotify-tools python2.7 build-essential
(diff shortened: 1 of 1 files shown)


anvmn:
@AronNovak any idea why this failure started? Do we need restore python 2.7 installation (checked release notes script, and it seems to work)? With this change, we have python 3.11.2 on DDEV:

Selection_428

--- UPDTATE;

Seems like this is not a good solution, as we now get dependnecy issues when running `ddev gulp`. See https://github.com/TIP-Global-Health/eheza-app/issues/1117#issuecomment-2068050037.

On my local, downgraded DDEV to 1.22.7 to be able to proceed development.

AronNovak:
On my local, seemingly, I have an independent issue, for the record: https://stackoverflow.com/questions/54085943/npm-err-invalid-dependency-type-requested-alias - with `npm`. But the suggestion in the thread does not work, as we use an older NodeJS.
So for now, indeed we could use older DDEV, but keeping in mind to upgrade, when we have time to do so.

anvmn:
The fix is to freeze DDEV version for Travis. Implemented at #1121. Closing.

Issue #1093, opened 2024-03-31
HIV Management - Medications


anvmn:
![image](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/7b9a3d5b-e65f-4728-b179-89a064be6406)
![Selection_353](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/6bcbebd0-87fe-4362-b2f1-813c834d065e)

anvmn:
@adamhstewart 

Option 1 and Option 3 for medications is identical. How do we fix it?

![Selection_364](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/7c1754f5-5a26-4134-95fc-71861bf1e291)

Diama1:
@anvmn Based on the HIV doc, here is the third option 
![Screenshot from 2024-04-03 10-53-20](https://github.com/TIP-Global-Health/eheza-app/assets/32199791/4362f044-d838-43f6-984d-1f01978b239a)

anvmn:
@adamhstewart

Deployed to Nutrition env. Ready for QA.

balagan73:
https://nutrition-ihangane.pantheonsite.io/app

## QA Review

### Found issues and questions:

1. Instead of adverse effect and effects, adverse events is displayed.
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/dcb6806d-6d09-412f-b7f5-f1902cded50c)

### Checked and approved:

1. Prescribed medication screen looks and works per design
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/79afbae9-554b-468a-a4d7-c3bd9dc86c0b)

2. Treatment review screen looks and works per design except the difference in adverse effect pointed out in issues point 1. 
![image](https://github.com/TIP-Global-Health/eheza-app/assets/3536805/489de9f5-9fa6-413a-9b16-86047e1977a5)

adamhstewart:
@anvm changes for medications:

There are three categories of medication to be added, they should be labeled with the list headers on the screen exactly as they appear below

Most Common Anti-Retroviral Medications: 
- Dolutegravir + Lamivudine + Tenofovir (DTG-3TC-TDF) 
- Atazanavir + Ritonavir (ATZ/r) 
- Dolutegravir (DTG) 
- Abacavir + Lamivudine (ABC-3TC) called Kivexa in Burundi 
- Lamivudine + Tenofovir (3TC-TDF)
[...]

PR #1132, opened 2024-04-25
HIV management - update medication set

anvmn:
#1093 

Per https://github.com/TIP-Global-Health/eheza-app/issues/1093#issuecomment-2077019400


anvmn:
Peek 2024-04-25 16-24

Issue #1133, opened 2024-04-25
HIV management - conditional Prescribed Medication sub-activity at subsequent encounter [3h]


anvmn:
During subsequent encounter, we need to know if patient is taking different medications from the ones that were prescribed, and if so, record them.

### Tasks:

- At Treatment Review sub-activity, add question `Are you taking different medications from the ones prescribed?`
- If answer is positive, show Prescribed Medication sub-activity to record the medications being taken.

anvmn:
@adamhstewart 

Worth QAing this. Deployed to Nutrition env.

adamhstewart:
@anvmn After QA-ing requesting some substantial changes: 
- This question should be asked on the "Prescribed Medications" sub activity. That means that the subactivity will always be present both on first encounters and subsequent encounters.
- The question should be phrased: "Please check the medications that are prescribed. Is the patient still taking the following medications?" The currently prescribed medications should be listed.
    - If yes, continue to allow to close the subactivity
    - If no, present the list of medications to update the prescription.

Bump 4h

adamhstewart:
@adamhstewart 

## Checked and approved
- This question should be asked on the "Prescribed Medications" sub activity.
- If yes, continue to allow to close the subactivity
- If no, present the list of medications to update the prescription.

## Please Fix

- CHANGE: "Please check above the medications that were prescribed. Is the patient still taking the medications listed?" TO "Please check the medications that were prescribed. Is the patient still taking the medications listed above?"

PR #1135, opened 2024-04-25
HIV management - Prescribed Medication sub-activity at subsequent encounter

anvmn:
#1133


anvmn:
Peek 2024-04-25 18-06

Issue #1137, opened 2024-04-30
WhatsApp - add basic number validations to identify invalid numbers and prevent sending messages there [2h] 


anvmn:
### Validations:
- Number should not be empty
- Number should be digits only
- Minimal length of 6 digits

Example of message sent to empty number:
![Selection_456](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/57ce9596-c234-459f-82ed-18f38b30d45f)

anvmn:
Deployed to live

PR #1138, opened 2024-04-30
WhatsApp - destination number validations

anvmn:
#1137


anvmn:
Peek 2024-04-30 12-33

Issue #1128, opened 2024-04-25
When a nurse selects “Lab” for Hepatitis B test, the diagnosis for Hepatitis B and Referral activity don't appear. (and other scenarios)


LudmilaGoichman:
Nutrition env

1. When a nurse selects “Point of Care” and the tests are positive for Hepatitis B and Malaria, two activities appear: Medical Distribution and Referral. I think only Referral should appear

<img width="355" alt="image" src="https://github.com/TIP-Global-Health/eheza-app/assets/35059802/4316066c-ebe6-49f0-bec9-ffb1d345d0d1">|<img width="355" alt="image" src="https://github.com/TIP-Global-Health/eheza-app/assets/35059802/64e323f0-7ac5-491a-b4b2-29de03128447">
---|---

2. When a nurse selects “Point of Care” and the tests are positive for Hepatitis B and Malaria, the diagnosis appears twice. And when Malaria is negative

<img width="355" alt="image" src="https://github.com/TIP-Global-Health/eheza-app/assets/35059802/4316066c-ebe6-49f0-bec9-ffb1d345d0d1">| <img width="355" alt="image" src="https://github.com/TIP-Global-Health/eheza-app/assets/35059802/68d65c99-6140-42b5-be00-e90371be4c9e">
---|---

<img width="353" alt="image" src="https://github.com/TIP-Global-Health/eheza-app/assets/35059802/4593bb4b-e54d-4881-aae6-33cea2e6deb6">

3. Patient Akuzwe Gossy

When a nurse selects “Lab” for Hepatitis B test and the test is positive (and Malaria is negative - “Point of Care”), the diagnosis for Hepatitis B and Referral activity don't appear. 

<img width="353" alt="image" src="https://github.com/TIP-Global-Health/eheza-app/assets/35059802/503f666f-dea2-4a37-b27c-9c104be6b39a">|<img width="353" alt="image" src="https://github.com/TIP-Global-Health/eheza-app/assets/35059802/4626ed4a-4866-484b-8a33-ad8de57fadc0">
---|---

4. Patient Akuzwe Ketsia

When a nurse selects “Lab” for Hepatitis B and the test is positive (and Malaria is positive - “Point of Care”), the diagnosis for Hepatitis B and Referral activity don't appear. The Malaria diagnosis and Medical Distribution action are displayed before the lab test, and after the Medical Distribution action is completed, a Pause Encounter button appears.

[...]

PR #1140, opened 2024-05-01
ANC Labs - fix diagnostics for recurrent encounter phase

anvmn:
#1128


anvmn:
Peek 2024-05-01 14-00
Peek 2024-05-01 13-58

Issue #1141, opened 2024-05-01
HIV Management - Add testing option for patients without a diagnosis [4h]


adamhstewart:
## Issue
If a patient indicates they do not have an HIV diagnosis, the CHW should be presented with the option to perform a rapid HIV test.

### Logic:
Question: "Was this person diagnosed with HIV?"
- If yes, continue as currently implemented.
- If, no ask "Would you like to perform an HIV test Today?

Question: "Would you like to perform an HIV test Today?"
- If yes, ask for test results (same question and responses to existing Rapid HIV Testing)
- If no, continue to close encounter

## Tasks
- Create 2 new questions and logic
- Update backend infra

adamhstewart:
## Checked and Approved
- If yes, continue as currently implemented.
- If, no ask "Would you like to perform an HIV test Today?
- If yes, ask for test results (same question and responses to existing Rapid HIV Testing)
- If no, continue to close encounter

PR #1145, opened 2024-05-02
HIV Management - option to record HIV test result at Diagnostics activity

anvmn:
#1141


anvmn:
Peek 2024-05-04 23-48

Issue #1146, opened 2024-05-04
HIV Management - Add Option for Not taking medication as prescribed [1h]


adamhstewart:
## Issue
In some cases, a patient may have not yet have started treatment after having just received a prescription. Under the treatment review subactivity, there should be an option for "Treatment not yet started" as an answer to "Why not?" (the conditional question after a user indicates that "No" to "Did you take the medication as prescribed?"

<img width="1111" alt="E-Heza_System" src="https://github.com/TIP-Global-Health/eheza-app/assets/688507/af68f441-b812-41db-890b-c00ede8000f6">

## Task
- Add option "Treatment not yet started"

adamhstewart:
## Checked and Approved
- Add option "Treatment not yet started"

PR #1147, opened 2024-05-05
HIV Management - 'Treatment not started' option at 'Reason for not taking' medication

anvmn:
#1146


anvmn:
Peek 2024-05-05 10-08

Issue #1154, opened 2024-05-07
Add icon for case management that was created today [2h]


adamhstewart:
## Issue

When a followup that was created today appears in case management, it appears without an arrow (since and encounter can't be created in the same day):

<img width="879" alt="E-Heza_System" src="https://github.com/TIP-Global-Health/eheza-app/assets/688507/8d5ea836-fd52-4a46-a180-9b4d511f13eb">

Rather we should use a grayed out arrow

## Task
- substitute blank space with a greyed out arrow.

anvmn:
Deployed to live

PR #1157, opened 2024-05-08
Case Management - view gray icon when start encounter action is disabled

anvmn:
#1154


anvmn:
Selection_498

mariano-dagostino (approved):
@anvmn Looks good. I wonder if instead of a new icon you could de-saturate the image using css. But I'm not sure how other disabled icons are managed in the app.

PR #1169, opened 2024-05-16
Dashboard stats - Make sure only single value is sent for muac and zscores

anvmn:
https://github.com/Gizra/ihangane/issues/3144

### DIFF
--- a/server/hedley/modules/custom/hedley_stats/hedley_stats.module
+++ b/server/hedley/modules/custom/hedley_stats/hedley_stats.module
@@ -1625,10 +1625,10 @@ function hedley_stats_get_spv_data($health_center_id) {
       $item['encounter_type'] = $row->field_well_child_encounter_type;
     }
     if (!empty($row->muac)) {
-      $item['muac'] = $row->muac;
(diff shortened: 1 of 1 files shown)


server/hedley/modules/custom/hedley_stats/hedley_stats.module
```diff
@@ -3802,3 +3809,16 @@ function hedley_stats_add_patient_details($node, array &$patients_details, $add_
...
+ * Extracts the first value from a comma-separated string of values.
+ *
+ * @param string $values_string
+ *   The values string.
+ *
+ * @return string
+ *   First value.
+ */
+function hedley_stats_extract_first_value($values_string) {
+  return explode(',', $values_string)[0];
```
AronNovak, line 3823: Should we play safely? So I am not sure for instance `explode` will return an array in every case.

anvmn: Selection_529
Selection_527
Selection_528

AronNovak (approved):

Issue #1022, opened 2024-02-17
TB Management - Translations


adamhstewart:
## Issue
Translation for Kinyarwanda and Kirubndi for TB Management module - most strings should be available now for translation

## Tasks
- Translation in Kinyarwandsa
- Translation in Kirundi

adamhstewart:
@Diama1 Kinyarwanda handled in #1053 - leaving this open for Kirundi.

anvmn:
Deployed to live

PR #1189, opened 2024-05-31
TB Management - Kirundi translation

Diama1:
#1022


anvmn:
@Diama1 is this completed?

PR #1220, opened 2024-06-24
WIP: Install tools on-the-fly, use standard DDEV image

AronNovak:


AronNovak:
@amitaibu @anvmn This is a no-go for the sake of local development.
If we install it on-the-fly, `ddev restart` means that we need to re-download the tools again. We'd shift the instability more to local development, than to CI, which is even worse for the day-to-day work.

Issue #1247, opened 2024-07-11
Eheza Reports - data anomalies


anvmn:
### Problems:

1. Initial, there was no Person CT. Instead there were Mother and Child CTs. Person CT has replaced them after ~1.5 years. All Mother and Child nodes were migrated to Person nodes.
The problem is that creation date of Person node did not original creation of Child/Mother node.
This means that we got 6327 patients with wrong creation date.

2. Some patients got invalid Geo location fields. This causes patients to be excluded at reports.

 - While there are 3 provinces, there are 10 different values in DB:
![Selection_706](https://github.com/user-attachments/assets/c84b849a-4ebb-4b67-93da-3b4d7322be8b)

 - While there are 5 districts, there are 8 different values in DB:
 
![Selection_709](https://github.com/user-attachments/assets/62e93579-a7fa-41bf-88bf-637496d605e6)

- 152 patients that do not have cell field set:

![Selection_720](https://github.com/user-attachments/assets/3a42cd25-c917-4ddb-8af4-38061e22c850)

anvmn:
### Invalid Province - 'Parent':

Child with all Geo fields set to Parent':

![Selection_707](https://github.com/user-attachments/assets/ade94796-5d6b-473f-9263-48b8ee9d4c3d)

Got a relationship with mother:

![Selection_708](https://github.com/user-attachments/assets/cd28e384-d345-4836-ab0d-388e0b948a58)

Mother got Geo fields set properly.

### Resolution - setting fields per mother.

anvmn:
### Other provinces mapping:

```
    'AMAJYARUGU' => 'Amajyaruguru',
    'North' => 'Amajyaruguru',
    'North Province' => 'Amajyaruguru',
    'Northern Province' => 'Amajyaruguru',
    'UBURASIRAZUBA' => 'Iburasirazuba',
    'Kigali City' => 'Umujyi wa kigali',
```

### Resolution - handled at dedicated script.

anvmn:
### Patients with empty district:

![Selection_713](https://github.com/user-attachments/assets/262563e5-214b-4f2f-9acb-d92341468e08)

For KIVURUGA sector, district must be Gakenke. Setting manually.

![Selection_715](https://github.com/user-attachments/assets/6efa58a5-f1ba-44e2-b9be-ddbe43dd747e)

For RUSHASHI sector, district must be Gakenke. Setting manually.

anvmn:
[...]

PR #1248, opened 2024-07-11
 Eheza Reports - fix data anomalies

anvmn:
#1247

### DIFF
--- a/server/elm/src/Pages/Reports/View.elm
+++ b/server/elm/src/Pages/Reports/View.elm
@@ -81,14 +81,16 @@ viewReportsData language currentDate themePath data model =
                                 let
                                     dateSelectorConfig =
                                         let
-                                            sixYearsAgo =
+                                            -- The date system became live, and
+                                            -- first content was uploaded.
+                                            launchDate =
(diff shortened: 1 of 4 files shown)


server/elm/src/Pages/Reports/View.elm
```diff
@@ -81,7 +82,7 @@ viewReportsData language currentDate themePath data model =
                                     dateSelectorConfig =
                                         let
                                             sixYearsAgo =
-                                                Date.add Years -6 currentDate
+                                                Date.fromCalendarDate 2018 Jan 1
```
mariano-dagostino: This doesn't look right. Eight the variable name is incorrect or the calculation is wrong.

mariano-dagostino (changes requested):
@anvmn the php code looks good. the elm code seems incorrect.

server/elm/src/Pages/Reports/View.elm
anvmn: @mariano-dagostino  Indeed, wrong variable name. Fixed.

mariano-dagostino (approved):

PR #1250, opened 2024-07-17
Fix for auto creation of CHW group participant when enrolling newborn

anvmn:
https://github.com/Gizra/ihangane/issues/3150


anvmn:
Trivial change. Merging.

Issue #1255, opened 2024-07-18
TB management - display the progress report after the save of the last to do activity


Diama1:
## Issue

When the user completes the final activity, they need to be taken directly to the progress report, rather than seeing an empty page and having to click on the progress report tab. This should be consistent with the behavior of other modules.

## Todo
- Implement as for the other modules

anvmn:
@LudmilaGoichman 

Deployed to Nutrition env.
Please test for first and subsequent encounters.

LudmilaGoichman:
@anvmn 

- nutrition env
- android tablet

### Questions:
1. Should I check all modules? 

>Implement as for the other modules

2. patient Ntwari Claude. I had a few (first and subsequent) encounters, but the diagnosis didn't appear in the report

subsequent encounter:

https://github.com/user-attachments/assets/d6bc847b-1293-4397-98ea-4c9ab7a7b827

### Checked and Approved:
The progress report is displayed after the save of the last activity 

https://github.com/user-attachments/assets/68b74e3c-fe50-4908-a97d-a16f61099641

anvmn:
@LudmilaGoichman 

> patient Ntwari Claude. I had a few (first and subsequent) encounters, but the diagnosis didn't appear in the report

What you show is first encounter. I know this, because only first encounter got diagnostics activity. For subsequent encounter to be possible, at first encounter, Tuberculosis must be diagnosed.

LudmilaGoichman:
@anvmn no, it's the third encounter:

<img width="720" alt="image" src="https://github.com/user-attachments/assets/f6cf0aa5-1ea9-4416-9c8b-b7e537d704a2">

<img width="720" alt="image" src="https://github.com/user-attachments/assets/71593458-2ae7-4b36-bcbb-7e7c82c6320a">

<img width="718" alt="image" src="https://github.com/user-attachments/assets/2325500f-8f7d-489d-8df1-27fd2161a69a">

anvmn:
@LudmilaGoichman 

Are those all share same participant?

LudmilaGoichman:
@anvmn 
yes 

<img width="1071" alt="image" src="https://github.com/user-attachments/assets/52d18e42-d168-4a10-a47b-19c5e7aa651f">

anvmn:
@LudmilaGoichman 

Actually, each encounter got it's own participant:

[...]

PR #1256, opened 2024-07-21
TB management - display the progress report when all activities are completed

anvmn:
#1255

### DIFF
--- a/client/src/elm/Backend/Update.elm
+++ b/client/src/elm/Backend/Update.elm
@@ -2048,7 +2048,7 @@ updateIndexedDb language currentDate currentTime zscores site features nurseId h
                         in
                         ( newModel, extraMsgs )
 
-                processVitalsRevisionAndUpdateLabsResults participantId encounterId value =
+                processVitalsRevisionAndUpdatePrenatalLabsResults participantId encounterId value =
                     if downloadingContent then
                         ( model, [] )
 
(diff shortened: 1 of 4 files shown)


anvmn:
Peek 2024-07-21 11-39

client/src/elm/Backend/Update.elm
```diff
@@ -3569,6 +3589,76 @@ updateIndexedDb language currentDate currentTime zscores site features nurseId h
...
+
+                [ TuberculosisTreatmentReviewRevision _ data ] ->
+                    let
+                        ( newModel, extraMsgs ) =
+                            processTuberculosisRevisionAndNavigateToProgressReport data.encounterId
+                    in
+                    ( newModel
+                    , Cmd.none
+                    , extraMsgs
+                    )
```
mariano-dagostino, line 3660: @anvmn Any reason to use a more compact approach?

```elm
[ TuberculosisDOTRevision _ data
, TuberculosisFollowUpRevision _ data
, TuberculosisHealthEducationRevision _ data
, TuberculosisMedicationRevision _ data
, TuberculosisReferralRevision _ data
] ->
    let
        ( newModel, extraMsgs ) =
            processTuberculosisRevisionAndNavigateToProgressReport data.encounterId
    in
    ( newModel
    , Cmd.none
    , extraMsgs
    )
```

mariano-dagostino (approved):

client/src/elm/Backend/Update.elm
anvmn: @mariano-dagostino 

We're telling here that we expect a list with single revision of specific type.
This is what we get when we save a measurement on client.
There is no better way to make this more compact, without over complicating the code, making it unreadable.

Issue #1257, opened 2024-07-22
Eheza Reports - calculate and store Nutrition report results at backend (for large data sets)


anvmn:
When we need to generate Nutrition report for large data sets, it takes several minutes to complete, and browser is constantly asking if to continue waiting, or force close the page.
This is poor user experience.

Since Nutrition report is always provided to current date, for each large data set (Global, HC, Province and District), we can calculate results using Drush script and store them as content (during night time).
Then, instead of generating results on front-end, we can simply present results pulled from backend.

anvmn:
Deployed to live

PR #1258, opened 2024-07-22
Eheza Reports - calculate and store Nutrition report results at back-end

anvmn:
#1257


anvmn:
Need to merge for delivery. If needed, review fixes will be deployed separately.

Issue #1265, opened 2024-07-31
RM - Correct translations 


Diama1:
## Issue 

Some of the texts are not translated. they need to be fixed and add the missing translations.

## Todo

- Correct the translation and add the missing one

anvmn:
Deployed to live

PR #1268, opened 2024-08-01
Resilience Message - Correct Translations

Diama1:
#1265

### DIFF
--- a/client/src/elm/Translate.elm
+++ b/client/src/elm/Translate.elm
@@ -413,7 +413,7 @@ type TranslationId
     | AILaboratoryTask AILaboratoryTask
     | All
     | AllowedValuesRangeHelper FloatInputConstraints
-    | AlmostEverydayOption
+    | AlmostEveryday
     | AmbulancArrivalPeriodQuestion
     | ANCEncountersNotRecordedQuestion
     | ANCIndicateVisitsMonthsPhrase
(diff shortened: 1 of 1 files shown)


Diama1:
Screenshot from 2024-08-01 14-55-21
Screenshot from 2024-08-01 14-55-38
Screenshot from 2024-08-01 14-56-22
Screenshot from 2024-08-01 14-56-35

anvmn (changes requested):
@Diama1 

To prevent duplication, define standalone translation and use `translationSet` (see `Never` as example):
Selection_763

Make this for following:

Selection_755
Selection_757

Selection_756

Selection_758

Selection_759
Selection_760

Selection_761


Selection_762

client/src/elm/Translate.elm
```diff
@@ -413,6 +413,7 @@ type TranslationId
     | AILaboratoryTask AILaboratoryTask
     | All
     | AllowedValuesRangeHelper FloatInputConstraints
+    | AlmostEverydayOption
```
anvmn, line 416: Drop the `Option` suffix, here, and everywhere bellow.

General  note - if you decide on standard, it should be implemented everywhere.
So, if you decided to have `Option` suffix, there should not be sets without it.

anvmn (changes requested):

anvmn (approved):

Issue #1274, opened 2024-08-11
Update contrib modules 


anvmn:
### Tasks:

- Update all contrib modules that got version for upgrade
- ~~Adapt code for PHP 8.2~~

anvmn:
PHP version upgrade is halted. See https://github.com/TIP-Global-Health/eheza-app/pull/1275#issuecomment-2284353052.

PR #1275, opened 2024-08-11
Contrib modules upgrade

anvmn:
#1274

### Deployment instructions:

- `drush updb -y`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ workflows:
 jobs:
   lint_phpcs:
     docker:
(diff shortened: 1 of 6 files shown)


anvmn:
2 types of errors when changing date:

Selection_790

For first (Ramsey), need to update version to most recent.

For second (restful), will create a patch.

anvmn:
Decided to halt PHP version upgrade. Found several contrib modules that are incompatible.
Also, simpletest produces hundreds of exceptions:
Selection_795

AronNovak (approved):

Issue #1279, opened 2024-08-14
Statistical queries, Demographics report - incorrect counting of Nutrition group encounters


anvmn:
In group encounter, participant  is adult + child pair.
Currently, in case adult recorded a measurement (family planning), encounter is counted bot for child and adult, so we get 2 instead of 1.

adamhstewart:
@anvmn we intentionally count this as two encounters (as it's two people) - change should be reverted.

PR #1280, opened 2024-08-14
Statistical queries, Demographics report - fix counting of Nutrition group encounters

anvmn:
#1279


anvmn:
Trivial change. Merging.

Issue #1282, opened 2024-08-14
META: Encounters completion stats


anvmn:
For each type of encounter we want to know:
-  For each of possible activities:
    1. How many activities in total were shown.
    2. How many activities from those shown were completed.

anvmn:
### High level design:

1. Each type of encounter will be added a field where it's completion data will be stored.
2. For each type of encounter, we'll have a dedicated script that will locate encounters that don't have completion data set, and will compute and set it.
3. The scripts will be executed once a day, during night time.
4. We'll add new front-end APP that will allow selection of desired report:
  - Encounter type
  - Scope (Global / HC)
  - Date range (?)
 5. App will forward request to back-end, where encounters matching the criteria will be loaded, and their completion metrics will be aggregated and send to front-end.
 6. App will process raw data and generate completion statistics.

anvmn:
Notes:

- Encounters had new logic and activities added with time. Each new content type that was added requires additional check for launch date (date when new version that included that change was deployed).
- There are special scenarios where activities that are usually considered mandatory are skipped. For example COVID suspect diagnosis on CHW app will lead straight away to Next steps.

PR #1283, opened 2024-08-14
[Feature] Completion reports

anvmn:
#1282


anvmn:
First results for Nutrition Individual encounters:

Selection_802

anvmn:
Selection_813

anvmn:
@AronNovak  please review. Docs are added at #1288.

server/hedley/modules/custom/hedley_reports/scripts/generate-completion-data.php
```diff
@@ -0,0 +1,93 @@
...
+    // No more items left.
+    break;
+  }
+
+  $ids = array_keys($result['node']);
+  $nodes = node_load_multiple($ids);
+  foreach ($nodes as $node) {
+    $completion_data = hedley_reports_generate_completion_data_for_nutrition_encounter($node);
+    $node->field_reports_data[LANGUAGE_NONE][0]['value'] = json_encode($completion_data);
+    node_save($node);
```
AronNovak: should we wrap in a try-catch and add error handling upon failure? Or ok to let it fail as-is?

server/hedley/modules/custom/hedley_reports/scripts/generate-data-for-all.php
```diff
@@ -34,7 +34,7 @@
   ->addTag('exclude_deleted');
 
 if ($exclude_set) {
-  $base_query->addTag('exclude_set');
+  $base_query->addTag('exclude_set_reports_data');
```
AronNovak, line 37: Where do we use this tag?

AronNovak (approved):

server/hedley/modules/custom/hedley_reports/scripts/generate-completion-data.php
anvmn: @AronNovak 

At https://github.com/TIP-Global-Health/eheza-app/pull/1283/files#diff-576b4cc9568e1d723be865511b32f6fcb9147fdf926905d82686213e1435548dR23.

The context is that every encounter has to calculate it's completion data only once.
From that point on, it never changes.

So, the plan is to run the script once initially, and then, have overnight Jenkins task that will fill completion data for only those encounters, that don't have it set.

server/hedley/modules/custom/hedley_reports/scripts/generate-data-for-all.php
anvmn: @AronNovak 

At https://github.com/TIP-Global-Health/eheza-app/pull/1283/files#diff-576b4cc9568e1d723be865511b32f6fcb9147fdf926905d82686213e1435548dR23.

The idea is that every encounter has to calculate it's completion data only once.
From that point on, it never changes.

So, the plan is to run the script once initially, and then, have overnight Jenkins task that will fill completion data for only those encounters, that don't have it set.

server/hedley/modules/custom/hedley_reports/scripts/generate-completion-data.php
anvmn: @AronNovak 
Don't see a real need for this. `node_save()` failures are very rare.
If it will happen, the calculation will be attempted again the next night.
As we deal with thousands of encounters, final completion results will not be affected.

Issue #1287, opened 2024-08-19
Encounters completion stats - Acute Illness encounters


anvmn:
For each of possible activities:
    1. How many activities in total were shown.
    2. How many activities from those shown were completed.

anvmn:
- Feature launch date  - September 30, 2020.
- First encounter created on October 01, 2020.
- Activities released after launch:
  - acute_illness_contacts_tracing (2021-12-22)
  - acute_illness_core_exam (2021-12-22)
  - acute_illness_danger_signs (Wednesday, 24 March 2021)
  - acute_illness_follow_up (Tuesday, 8 June 2021)
  - acute_illness_muac (Thursday, 6 December 2020)
  - acute_illness_nutrition (Thursday, 6 December 2020)
  - covid_testing (2021-12-22)
  - health_education (Wednesday, 24 March 2021)
  - treatment_ongoing (Wednesday, 24 March 2021)

anvmn:
**Note to self:**

`acute_illness_trace_contact` CT is never used - consider deleting it.

PR #1288, opened 2024-08-19
Encounters completion stats - Acute Illness

anvmn:
#1287

### DIFF
--- a/server/elm/src/Backend/Completion/Utils.elm
+++ b/server/elm/src/Backend/Completion/Utils.elm
@@ -59,6 +59,82 @@ nutritionMotherActivityFromMapping mapped =
             Nothing
 
 
+acuteIllnessActivityFromMapping : String -> Maybe AcuteIllnessActivity
+acuteIllnessActivityFromMapping mapped =
+    case mapped of
+        "a" ->
+            Just AcuteIllnessAcuteFindings
+
+        "b" ->
+            Just AcuteIllnessContactsTracing
+
+        "c" ->
+            Just AcuteIllnessCoreExam
+
+        "d" ->
+            Just AcuteIllnessDangerSigns
+
+        "e" ->
+            Just AcuteIllnessFollowUp
+
+        "f" ->
+            Just AcuteIllnessMUAC
+
+        "g" ->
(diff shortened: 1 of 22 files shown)


anvmn:
**Initial results:**

Selection_815

anvmn:
Results after fixes:

Selection_820

server/elm/src/Backend/Completion/Utils.elm
```diff
@@ -59,6 +59,82 @@ nutritionMotherActivityFromMapping mapped =
             Nothing
 
 
+acuteIllnessActivityFromMapping : String -> Maybe AcuteIllnessActivity
+acuteIllnessActivityFromMapping mapped =
+    case mapped of
+        "a" ->
+            Just AcuteIllnessAcuteFindings
+
+        "b" ->
```
AronNovak, line 68: For my curiosity, where do we use these letters? Would it make debugging (like via network tab, PHP side, etc) easier if we use more expressive strings on the left side?

AronNovak (approved):

server/elm/src/Backend/Completion/Utils.elm
anvmn: @AronNovak 
For the context:
- Data we need to present is side wide (currently, we got ~170K patients and ~500K encounters, and growing). We also need to support presentation of smaller chunks of data (HCs, Districts, Villages, etc...)
- Data needs to be filterable by period ( start date to end date).
- Data needs to be presented on fly, within reasonable period of time (up to 30 sec).
- RAM limit at Pantheon site is currently 512MB.

To support these requirements, the idea is that every patient/encounter will store it's own data locally (data is updated based on uploaded content / Jenkins tasks).
When queried for, data is generated on back-end and then attached as script for Elm APP to decode and operate with.
Also, large data sets (Global, Provinces, Districts) can not be loaded within reasonable time.
As a result, we have dedicated  CT that stores file of encoded data (updated daily).

So, the mapping we do, reduces the size of the file by factor of ~ 10.
Also, it reduces memory size we consume when loading that data before passing it to Elm App.
Mapping is 'decoded' on client side, there's no issue with debugging complexity.

Issue #1289, opened 2024-08-26
Encounters completion stats - Well Child


anvmn:
For each of possible activities:
    1. How many activities in total were shown.
    2. How many activities from those shown were completed.

anvmn:
**First encounter uploaded:   29 December 2021** 
**Feature launch: 22 December 2021**

**30 activities in total:**

well_child_albendazole - 25 May 2022
well_child_bcg_immunisation -  9 February 2022 
well_child_caring - 4 April 2024
well_child_contributing_factors - 15 March 2022
well_child_dtp_immunisation - 9 February 2022
well_child_ecd - 15 March 2022
well_child_feeding - 4 April 2024
well_child_follow_up - 5 March 2022
well_child_food_security - 4 April 2024
well_child_head_circumference - 10 February 2022
well_child_health_education - 10 February 2022
well_child_height - 9 February 2022
well_child_hygiene - 4 April 2024
well_child_ipv_immunisation - 9 February 2022
well_child_mebendezole - 9 February 2022
well_child_mr_immunisation - 23 March 2022
well_child_muac - 9 February 2022
well_child_ncda - 21 November 2023
well_child_next_visit - 15 March 2022
well_child_nutrition - 9 February 2022
well_child_opv_immunisation -  9 February 2022
well_child_pcv13_immunisation - 17 March 2022
well_child_photo -  9 February 2022
well_child_pregnancy_summary -  9 February 2022
well_child_rotarix_immunisation - 9 February 2022
well_child_send_to_hc - 9 February 2022
well_child_symptoms_review -  9 February 2022
well_child_vitals - 9 February 2022
well_child_vitamin_a - 9 February 2022
well_child_weight - 9 February 2022

PR #1290, opened 2024-08-26
Encounters completion stats - Well Child

anvmn:
#1289


anvmn:
Selection_833
Selection_834

server/elm/src/Pages/Completion/Utils.elm
```diff
@@ -93,3 +107,102 @@ allNutritionMotherGroupActivities =
...
+    , WellChildHeight
+    , WellChildHygiene
+    , WellChildIPVImmunisation
+    , WellChildMebendezole
+    , WellChildMRImmunisation
+    , WellChildMUAC
+    , WellChildNCDA
+
+    -- This is just an indication that next visit notice
+    -- was presented. Not showign it.
```
AronNovak:
```suggestion
    -- was presented. Not showing it.
```

server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -3416,6 +3407,969 @@ function hedley_reports_count_acute_illness_signs(array $values) {
...
+      'measurements_by_type' => $measurements_by_type,
+    ];
+  }
+
+  foreach ($encounters_data as $index => $encounter_data) {
+    $encounter = $encounter_data['encounter'];
+    $measurements_by_type = $encounter_data['measurements_by_type'];
+
+    // Resolve encounter start date.
+    $start_date = explode(' ', $encounter->field_scheduled_date[LANGUAGE_NONE][0]['value'])[0];
```
AronNovak, line 3520: what do we do here? Extracting year or dropping time component? `DateTime` would likely do a same. Just thinking.

server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -3416,6 +3407,969 @@ function hedley_reports_count_acute_illness_signs(array $values) {
...
+  $query
+    ->entityCondition('entity_type', 'node')
+    ->entityCondition('bundle', 'individual_participant')
+    ->fieldCondition('field_encounter_type', 'value', 'child-scoreboard')
+    ->fieldCondition('field_person', 'target_id', $person_id)
+    ->propertyCondition('status', NODE_PUBLISHED)
+    ->addTag('exclude_deleted');
+
+  $result = $query
+    ->range(0, 1)
```
AronNovak, line 3732: Is it true that one child has always one scoreboard? Or would we need to sort the query?

server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -3416,6 +3407,969 @@ function hedley_reports_count_acute_illness_signs(array $values) {
...
+      continue;
+    }
+
+    // Loading all immunisation activities recorded during the encounter.
+    $query = new EntityFieldQuery();
+    $result = $query
+      ->entityCondition('entity_type', 'node')
+      ->entityCondition('bundle', $immunisation_types, 'IN')
+      ->propertyCondition('status', NODE_PUBLISHED)
+      ->fieldCondition('field_child_scoreboard_encounter', 'target_id', $encounter->nid)
```
AronNovak, line 3788: Do we need the `exclude_deleted` tag here as well? If not, why do we need it for one of the above `EntityFieldQuery`s?

server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -3416,6 +3407,969 @@ function hedley_reports_count_acute_illness_signs(array $values) {
...
+    if ($age_in_days >= (14 * 7)) {
+      $possible_immunisations[] = 'ipv';
+    }
+    if ($age_in_days >= (36 * 7)) {
+      $possible_immunisations[] = 'mr';
+    }
+
+    $site = variable_get('hedley_general_site_name', '');
+    if ($site == 'burundi') {
+      // All 3 dosed of DTP were given, it has passed at least 28 days since
```
AronNovak:
```suggestion
      // All 3 doses of DTP were given, it has passed at least 28 days since
```

server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -3416,6 +3407,969 @@ function hedley_reports_count_acute_illness_signs(array $values) {
...
+ * @return string|bool
+ *   A string representing the interval between doses (e.g., '+4 weeks',
+ *   '+9 months'), or FALSE if the immunization type is not recognized.
+ */
+function hedley_reports_well_child_get_interval_for_vaccine($immunisation_type, $site) {
+  switch ($immunisation_type) {
+    case 'bcg':
+    case 'ipv':
+    case 'dtp_sa':
+      return '+0 days';
```
AronNovak, line 4193: Does this mean that it's a one-time shot?

server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -3532,3 +4538,34 @@ function hedley_reports_ends_with($haystack, $needle) {
...
+ * @param string $pref
+ *   The prefix to be removed from the start of the string.
+ * @param string $suff
+ *   The suffix to be removed from the end of the string.
+ *
+ * @return string
+ *   The string without the specified prefix and suffix, or the original
+ *   string if the prefix or suffix is not found.
+ */
+function hedley_reports_drop_preffix_and_suffix(string $string, string $pref, string $suff) {
```
AronNovak:
```suggestion
function hedley_reports_drop_prefix_and_suffix(string $string, string $pref, string $suff) {
```

server/hedley/modules/custom/hedley_reports/scripts/generate-well-child-completion-data.php
```diff
@@ -0,0 +1,85 @@
...
+
+// Get allowed memory limit.
+$memory_limit = drush_get_option('memory_limit', 500);
+
+$type = 'individual_participant';
+$base_query = new EntityFieldQuery();
+$base_query
+  ->entityCondition('entity_type', 'node')
+  ->entityCondition('bundle', $type)
+  ->fieldCondition('field_encounter_type', 'value', 'well-child')
```
AronNovak, line 33: Do we need to exclude deleted here?

server/hedley/modules/custom/hedley_reports/scripts/generate-well-child-completion-data.php
```diff
@@ -0,0 +1,85 @@
...
+  ->propertyCondition('status', NODE_PUBLISHED);
+
+$count_query = clone $base_query;
+$count_query->propertyCondition('nid', $nid, '>');
+$count = $count_query->count()->execute();
+
+if ($count == 0) {
+  drush_print("There are no nodes of type $type for well child encounters in DB.");
+  exit;
```
AronNovak:
```suggestion
  exit 1;
```

Is it kind of an error?

server/hedley/modules/custom/hedley_reports/hedley_reports.module
anvmn: Yep

server/hedley/modules/custom/hedley_reports/hedley_reports.module
anvmn: Extracting YYYY-MM-DD and dropping HH-MM.

server/hedley/modules/custom/hedley_reports/hedley_reports.module
anvmn: Well, no, we don't need it here, as we query for measurements, and they don't have `field_deleted`.
This flag was introduces while we were developing merges of duplicate patients.
So, for example, patient was created twice, and each instances had one nutrition encounter.
So, when we merge them, we want to transfer all encounters and measurements from patient 1 to patient 2. The participant itself is not needed, as there should be single participant for patient in Nutrition (to which all encounters are pointing).
Therefore, participant should have `field_deleted` to mark patient 1 participant as deleted, but encounters and measurements should not, as they always get transferred.

server/hedley/modules/custom/hedley_reports/hedley_reports.module
anvmn: Similar to Nutrition, Child scorecard should have single participant, to which all encounters are pointing.

server/hedley/modules/custom/hedley_reports/scripts/generate-well-child-completion-data.php
anvmn: As we load participants here, indeed we need to add the tag. Nice catch!

server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -3687,7 +3687,7 @@ function hedley_reports_well_child_generate_vaccination_progress_data_regular(ar
     }
 
     foreach ($immunisations_by_type[$type]['doses'] as $index => $dose) {
-      $common_type = hedley_reports_drop_preffix_and_suffix($type, 'well_child_', '_immunisation');
+      $common_type = hedley_reports_drop_prefix_and_sufix($type, 'well_child_', '_immunisation');
```
AronNovak: This is tricky.
https://ies.ed.gov/ncee/rel/regions/southeast/pdf/prefix_base_word_suffix.pdf
`prefix`
and `suffix` are the correct forms AFAIK.

AronNovak (approved):

Issue #1306, opened 2024-09-13
Pregnancy Dating - Remove calendar restrictions and the range question


Diama1:
## Issue

For pregnancy dating we need to:
- Remove the When was the Patient's Last Menstrual Period?” Drop down question
- Unrestict the calendar for the user to choose any date they want defaulting at the current date.

<img width="1027" alt="E-Heza_System" src="https://github.com/user-attachments/assets/2218efe4-8ed6-491d-aaf3-50979e242c85">

## Todo
- Remove the range question
- Unrestrict the calendar

anvmn:
@LudmilaGoichman please test on Nutrition env

LudmilaGoichman:
@anvmn 

- nutrition env
- Android tablet

### Checked and Approved:
1. the first question was removed:

- Nurse:

<img width="352" alt="image" src="https://github.com/user-attachments/assets/41300e77-0d8b-465f-adcf-deb4770032f5">|<img width="352" alt="image" src="https://github.com/user-attachments/assets/7dcd8e83-75f6-4913-8a47-c1c7682df3c6">
---|---

- CHW:

<img width="352" alt="image" src="https://github.com/user-attachments/assets/c1653f94-4372-4157-ab3e-d10896715dd8">|<img width="352" alt="image" src="https://github.com/user-attachments/assets/1fcc9dea-1e29-4ac6-966d-2d231e0a58a7">
---|---

2. calendar: 
<img width="352" alt="image" src="https://github.com/user-attachments/assets/1f3a50c0-4f4d-4347-b66c-80c23d8eaf0e">

PR #1319, opened 2024-09-17
Pregnancy Dating - drop range question and adjust calendar restrictions

anvmn:
#1306


anvmn:
Peek 2024-09-17 12-07

AronNovak (approved):

Issue #1307, opened 2024-09-13
Family Ubudehe - Make the question optional and remove ubudehe on the display


Diama1:
## Issue

The question about Family Ubudehe should be optional since it is not used by everyone, and where ubudehe is displayed it should be removed. 

| Patient Registration | Patient Profile |
| -------- | -------- |
| <img width="1170" alt="E-Heza_System_🔊" src="https://github.com/user-attachments/assets/1c276502-65ab-4d39-abfc-be98b8daaaa8"> |  <img width="1071" alt="Screenshot_9_16_24__9_57_AM" src="https://github.com/user-attachments/assets/533765dd-1d6c-4553-897b-2dbd9060ecbe"> |

## Todo
- Make the question Family Ubudehe optional
- Remove the Ubudehe where it is displayed

anvmn:
@adamhstewart 

Note that Ubudehe is used when deciding if Mother FBF should be distributed or not.
Also, I suggest that instead  of removing Ubudehe display, we show it only if it got a value.

//cc @Diama1

PR #1328, opened 2024-09-19
Make Family Ubudehe Question Optional

Diama1:
#1307

### DIFF
--- a/client/src/elm/Activity/Utils.elm
+++ b/client/src/elm/Activity/Utils.elm
@@ -467,18 +467,7 @@ expectMotherActivity currentDate offlineSession motherId activity =
 
                                         entitledByUbudehe =
                                             Dict.get motherId offlineSession.mothers
-                                                |> Maybe.map
-                                                    (\mother ->
-                                                        case mother.ubudehe of
-                                                            Just Ubudehe1 ->
-                                                                True
-
-                                                            Just Ubudehe2 ->
-                                                                True
-
-                                                            _ ->
-                                                                False
-                                                    )
+                                                |> Maybe.map (\_ -> True)
                                                 |> Maybe.withDefault False
                                     in
                                     isBreastfeeding && entitledByUbudehe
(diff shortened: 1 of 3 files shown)


Diama1:
[ubudehe.webm](https://github.com/user-attachments/assets/f87f5b4c-bf6e-4e5b-a122-61c30187460d)

client/src/elm/Activity/Utils.elm
```diff
@@ -467,18 +467,7 @@ expectMotherActivity currentDate offlineSession motherId activity =
...
-                                                            Just Ubudehe1 ->
-                                                                True
-
-                                                            Just Ubudehe2 ->
-                                                                True
-
-                                                            _ ->
-                                                                False
-                                                    )
+                                                |> Maybe.map (\_ -> True)
```
anvmn: If there's no more dependency on Ubudehe, and every mother should get it, drop the condition.

anvmn (changes requested):
@Diama1 drop the condition

client/src/elm/Activity/Utils.elm
Diama1: @anvmn there is no dependency on Ubudehe

anvmn: @Diama1 then drop the condition

client/src/elm/Activity/Utils.elm
```diff
@@ -464,24 +464,8 @@ expectMotherActivity currentDate offlineSession motherId activity =
...
-                                                            Just Ubudehe2 ->
-                                                                True
-
-                                                            _ ->
-                                                                False
-                                                    )
-                                                |> Maybe.withDefault False
                                     in
-                                    isBreastfeeding && entitledByUbudehe
+                                    isBreastfeeding
```
anvmn: @Diama1 now we don't need to have it as a param.
Remove the `let`, and apply logic directly.

anvmn (approved):

Issue #1325, opened 2024-09-18
Skip Obstetric History page 2 on first pregnancy [3h]


adamhstewart:
## Issue 
On page 1 of Obstetric HIstory, If it is determined that this is the first pregnancy (Gravisa = 1 Para = 0000 ) then skip page 2 of Obstetric History and mark the subactivity complete. This occurs when the person checks currently pregnant but zeros on all other sections of page 1.

## Tasks
- Create the logic to skip the second part of Obstetric history when appropriate.

anvmn:
@LudmilaGoichman please test on Nutrition env

LudmilaGoichman:
@anvmn 

- Nutrition env
- Android tablet

### Checked and Approved:
1. When Gravida = 1 Para = 0000, the second page doesn't appear in Obstetric History. When Gravida and Para are higher, the second page appears.


https://github.com/user-attachments/assets/56eb44ae-fc29-44ce-8ae0-ed2d0f289d3e

PR #1331, opened 2024-09-19
ANC - skip Obstetric History step 2 on first pregnancy

anvmn:
#1325


anvmn:
Peek 2024-09-19 23-08

ci-scripts/circleci_wait.sh
```diff
@@ -3,7 +3,7 @@
 
 CMD="$*"
 
-TIMEOUT=3700
+TIMEOUT=4300
```
AronNovak, line 6: Was it a one-time hiccup that the timeout wasn't enough or do you think it's a frequent problem?

anvmn: @AronNovak 

It seems that whenever we exceed 1h execution, we get timeout anyways.
So, this change is not effective. Saw tasks fail with this increased timeout.
All failing tasks state 1h execution time. For example: 

https://app.circleci.com/pipelines/github/TIP-Global-Health/eheza-app/4742/workflows/022cde6a-0985-44b5-9143-90f92554c68b/jobs/16322

Selection_875

AronNovak: @anvmn if needed, let's open an issue for me and I can investigate a bit. Thanks for the details!

AronNovak (approved):

ci-scripts/circleci_wait.sh
anvmn: @AronNovak 

https://github.com/TIP-Global-Health/eheza-app/issues/1335

Issue #1335, opened 2024-09-23
Circle CI tasks timeout after 1h execution [3h]


anvmn:
All failing tasks state 1h execution time. For example:

https://app.circleci.com/pipelines/github/TIP-Global-Health/eheza-app/4742/workflows/022cde6a-0985-44b5-9143-90f92554c68b/jobs/16322

![Selection_875](https://github.com/user-attachments/assets/47a43281-378b-4441-ac77-f29141e65fa8)

AronNovak:
@anvmn With #1342 , do you consider the tests are stable enough according to the past two weeks?

anvmn:
@AronNovak Seem that tests now run under 40 minutes, so it's fine.

PR #1342, opened 2024-09-24
CircleCI optimization

AronNovak:
#1335


AronNovak:
@anvmn It seems to be a quick win. The execution time is well below 1h immediately with this. Sooner or later we must adapt the infrastructure according this too, but that's also a trivial YAML change, plus the QA processes of course.

anvmn (approved):

PR #1343, opened 2024-09-24
Parallel test execution

AronNovak:
#1335 

Based on https://circleci.com/docs/parallelism-faster-jobs/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,10 +37,14 @@ jobs:
       - run: ci-scripts/install_shell.sh
       - run: ci-scripts/test_shell.sh
   test_simpletest_linux:
+    parallelism: 2
     machine:
       image: ubuntu-2204:current
     steps:
       - checkout
       - run: ci-scripts/install_ddev.sh
       - run: ci-scripts/install_drupal.sh
-      - run: ci-scripts/circleci_wait.sh ddev simpletest
+      - run:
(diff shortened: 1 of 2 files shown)


AronNovak:
@anvmn On CircleCI, this helps, even in ~30 minutes, we have results (with more resource usage overall, so not really green, 2 build processes, 2 installation processes are executed, etc). But not sure if it interferes with how you execute tests locally.

anvmn:
@AronNovak 

Not sure if this improves considerably the speed right now.
The thing is that out of 4 tasks, 3 run for less than one minute.
The other one runs for 30 - 40 min.
So, as far as I understand, running 2 in parallel doesn't produce big improvement.
What we could do I suppose, is split simpletest task into to, where each will have half of tests,
and have these 2 run in parallel.

> so not really green

Is this what you refer to?
Selection_898

AronNovak:
Likely the whole thing does not worth, as we already specify concurrency at a lower level:
```
for TEST_FILE in $TEST_FILES; do
    php ./www/scripts/run-tests.sh --php "$(which php)" --concurrency 4 --verbose --color --url http://eheza-app.ddev.site:8081 --file "../../$TEST_FILE" 2>&1 | tee -a /tmp/simpletest-result.txt
done
```

That's why we see no improvement essentially, closing this PR then.

Issue #1271, opened 2024-08-07
Duplication Solution - Add Search by ID to Prevent Duplicates


Diama1:
## Issue

Currently, our search functionality is limited to name-based searches. To prevent duplicate entries, we propose adding the option to search by ID, as IDs are unique identifiers. The search field will have a placeholder informing users that they can search by either name or ID.

![Screenshot from 2024-08-07 09-35-38](https://github.com/user-attachments/assets/8ec0bdca-5577-4e8b-b701-c2eab0c4304b)

## Todo

- Implement the logic to search by ID number
- Create the screen as per design

PR #1375, opened 2024-11-17
Search by ID

anvmn:
#1271


anvmn:
Peek 2024-11-18 23-27

client/src/assets/scss/_new.scss
```diff
@@ -712,6 +712,7 @@ div.page-people {
         display: flex;
         flex-direction: column;
         height: 100%;
+        overflow-x: hidden;
```
AronNovak, line 715: What do we solve with this? Do we crop some parts?

client/src/elm/Backend/Model.elm
```diff
@@ -119,7 +119,11 @@ type alias ModelIndexedDb =
 
     -- Tracks searchs for participants by name. The key is the phrase we are
     -- searching for.
-    , personSearches : Dict String (WebData (Dict PersonId Person))
+    , personSearchesByName : Dict String (WebData (Dict PersonId Person))
+
+    -- Tracks searchs for participants by name. The key is the phrase we are
```
AronNovak:
```suggestion
    -- Tracks searches for participants by name. The key is the phrase we are
```

client/src/assets/scss/_new.scss
anvmn: Well, what happens is that a div that has input inside is few pixels wider that the screen, which case the horizontal overflow to appear.
There's not visual change when it's hidden, so, it's the easiest solution.

AronNovak (approved):

Issue #1376, opened 2024-11-24
POC:  pull GPS coordinates from the device and store them as a field during registration [12h]


anvmn:
### Objective:

The objective of this Proof of Concept (POC) is to demonstrate the ability of a mobile application to acquire accurate GPS coordinates of the user’s location. Additionally, the app will provide alternative methods for location input, including manual text entry and a map interface, in scenarios where GPS acquisition is unavailable due to bandwidth constraints or technical issues.

---
### Key Features

**GPS Coordinate Acquisition** 
Automatically acquire the user's current location using the device's built-in GPS capabilities.
Display acquired coordinates in real-time within the app interface.

**Manual Location Entry** 
Allow users to input latitude and longitude manually through a dedicated text input field.
Validate input to ensure proper format (e.g., decimal degrees or degrees/minutes/seconds).

**Map-Based Location Entry** 
Integrate a map interface allowing users to select a location by:
- Dropping a pin on the map.
- Searching for an address or place name.
- Display the selected location’s coordinates in the app.

**Data Storage** 
- Store acquired or entered coordinates on the backend in a standardized format.
- Provide an option to export or share coordinates for further use in a visual map format.

### Assumptions
- Bandwith cannot be totally relied upon to work at all or to produce accurate results - alternate input/confirmation is required
- Physical addresses are not available, nor would they be contained in any lookup reference (e.g., Google Maps) to product coordinates based on a text address.
- Users would be familiar with the location where they are providing the input coordinates (i.e., they could find it on a map).

---
This POC will validate the technical feasibility of the proposed functionality and serve as a foundation for future development.

anvmn:
@adamhstewart 

Since our APP runs in browser, we'll have to rely on Chrome to provides access to GPS coordinates (through the Geolocation API, a part of the browser's Web API).
Note that Chrome requests permission from the user to access the location.

For mobile devices (with built-in GPS sensors), Geolocation API fetches coordinates from the sensor, and IP network (Mobile/Wifi).
For desktops, IP network is used.

Since mobile devices have GPS sensors enabled as a default, and desktops got internet, so it seems to me that in majority of cases, we'll be able to get coordinates from Chrome.

> Manual Location Entry

There are 2 coordinates, 9 digits each. For example:

```
Registraiton longitude:  34.8987359
Registraiton laitude: 32.0785488
```

I don't think it's something that should be filled manually.


> Map-Based Location Entry

[...]

PR #1377, opened 2024-11-25
GPS coordinates during registration

anvmn:
#1376


anvmn:
Peek 2024-11-26 10-11


Selection_1031

Selection_1032

adamhstewart:
@anvmn This is good - let's move this from PoC to a "beta" feature that we can test I the field in a live environment (meaning we will deploy this).

Requesting the following changes:

1. Rather than storing the Lat/Long data invisibly to the user, display the coordinates on the registration page along with a boolean field to save the data (this would allow then to NOT save the data at an alternate location).

E-Heza_System


2. Make this feature configurable (on/off) by region/installation.

anvmn:
Peek 2024-12-15 20-33

client/src/js/app.js
```diff
@@ -517,6 +517,23 @@ elmApp.ports.scrollToElement.subscribe(function(elementId) {
   waitForElement(elementId, scrollToElement, null);
 });
 
+elmApp.ports.getCoordinates.subscribe(function() {
+  if ("geolocation" in navigator) {
+    navigator.geolocation.getCurrentPosition(
```
AronNovak, line 528: https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition#options

See the options here.
It would be meaningful to specify a timeout.
Also consider the others, `maximumAge` and `enableHighAccuracy`.

AronNovak (approved):

Issue #1270, opened 2024-08-07
Duplication Solution - Duplicate Patient Registration Warning


Diama1:
## Issue

When a user is registering a patient who has the same date of birth or ID number and resides in the same village as an existing patient, they should be alerted about the potential duplicate. The warning message will include the name and parent of the existing patient, along with the reason for the suspicion (matching date of birth or ID number and same village).

![Screenshot from 2024-08-07 09-36-37](https://github.com/user-attachments/assets/12c60638-a7e9-4128-a730-0bfe91583cd6)

## Todo

- Create the screen as per designs.
- Implement the logic for detecting potential duplicate patients.

anvmn:
@Diama1 

>  a patient who has the same date of birth or ID number and resides in the same village as an existing patient

Since national ID number is unique (it is, right?), I don't see the need to make a check for village. If we find existing patient with same national ID, we will present the warning.
Also, I don't really see how date of birth can be used here, since theoretically, there can be several patients from same village born on same day (twins, for example).

So, as I see it, only parameter checked here needs to be national ID.

anvmn:
@adamhstewart 

Implemented #1380 based on https://github.com/TIP-Global-Health/eheza-app/issues/1270#issuecomment-2510024219.
Tell me if changes are required.

PR #1380, opened 2024-11-28
Duplicate Patient Registration Warning

anvmn:
#1270


anvmn:
Peek 2024-11-28 20-29

client/src/elm/Pages/Person/Update.elm
```diff
@@ -95,35 +117,77 @@ update currentDate site reverseGeoInfo selectedHealthCenter maybeVillageId isChw
...
+                                                        , []
+                                                        )
                                             )
                                         -- We should never get here, because when editing,
                                         -- we always have the ID of person being edited.
-                                        |> Maybe.withDefault []
+                                        |> Maybe.withDefault ( [], [] )
+
+                        Form.Input nationalIdNumber _ (Form.Field.String value) ->
+                            if String.length value > 13 then
```
AronNovak, line 143: This is odd. If I read correctly, we talk about the national ID number here. In most countries, it has a fixed length. Can we rely on that here? Or is the situation different here for any reasons?

AronNovak (approved):
There's one question about data validation, otherwise looks good.

client/src/elm/Pages/Person/Update.elm
anvmn: Well, there are several countries we support and number of characters may vary.
The decision here is if to send fetch message to IndexedDB, with the results for national ID if input reached 14 characters, so in worst case, we'll have some extra search results in the model.

Issue #1383, opened 2024-12-15
Extend GPS PoC [5h]


adamhstewart:
# Issue

Baesed on #1376 , let's move this from PoC to a "beta" feature that we can test in the field in a live environment (meaning we will deploy this).

Requesting the following changes:

1. Rather than storing the Lat/Long data invisibly to the user, display the coordinates on the registration page along with a boolean field to save the data (this would allow then to NOT save the data at an alternate location).

<img width="743" alt="E-Heza_System" src="https://github.com/user-attachments/assets/fd33bcbf-43a0-4cbc-add0-ab76181b241f" />

2. Make this feature configurable (on/off) by region/installation.

PR #1384, opened 2024-12-15
GPS coordinates during registration - betta

anvmn:
#1383


anvmn:
Peek 2024-12-15 20-33

PR #1395, opened 2025-01-15
[Hotfix] Fix logic for decision if to display value in cell at Aggregated NCDA

anvmn:
https://github.com/Gizra/ihangane/issues/3185


anvmn:
Peek 2025-01-15 23-31

Issue #1410, opened 2025-01-29
Show that patient is unsure of LMP on progress report [2h]


adamhstewart:
## Issue 
When a patient is unsure of their LMP, it should be shown on the progress report

<img width="889" alt="Image" src="https://github.com/user-attachments/assets/928b5f8f-c750-4f97-a52f-3dc1fd273fe5" />

<img width="869" alt="Image" src="https://github.com/user-attachments/assets/3cd38e26-ec37-46b2-9307-92f5d9f83852" />

## Tasks
- add the labels when the LMP is uncertain.

adamhstewart:
@anvmn replace language from "Patient unsure of LMP" to "LMP not confident"

adamhstewart:
@Diama1 Adjust label to LMP: Not Confident and add tranlsation.

PR #1420, opened 2025-02-11
Adjust label to LMP: Not Confident and add Translations

Diama1:
[#1410 Comment](https://github.com/TIP-Global-Health/eheza-app/issues/1410#issuecomment-2648430624)

### DIFF
--- a/client/src/elm/Translate.elm
+++ b/client/src/elm/Translate.elm
@@ -926,7 +926,6 @@ type TranslationId
     | LevelOfEducationForResilience EducationLevel
     | LipidPanel
     | LiveChildren
-    | Lmp
     | LmpDateConfirmationLabel
     | LmpDateConfirmationQuestion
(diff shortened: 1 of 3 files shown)


client/src/elm/Translate.elm
```diff
@@ -24163,9 +24170,9 @@ translationSet trans =
             }
 
         UnsureOfLmp ->
-            { english = "LMP not confident"
-            , kinyarwanda = Nothing
-            , kirundi = Nothing
+            { english = "Not confident"
```
anvmn: => `LMP: Not Confident`

anvmn (changes requested):
@Diama1 

From issue requirements, only change need is to update and translate `UnsureOfLmp`.
I don't see why we need anything else.

anvmn (approved):

Issue #1426, opened 2025-02-20
Add 2nd IPV dose at 9 months - Rwanda Regionalization [2h]


adamhstewart:
## Issue

Rwanda's protocol for immunization recently changed after the development of the module in 2021; There is now a second dose of IPV at 9 months.

The protocol states that the IPV 2nd dose should be administered at 9 months with MR1.

## Task
- Add logic for the second IPV dose at 9 months.

anvmn:
@adamhstewart 

So far, all definitions for doses had 2 parameters:
- First dose (counting from birth date).
- Minimal interval from previous dose.

Per https://github.com/Gizra/ihangane/wiki/Module---Pediatric-Visit, first dose of IPV is given at 14 weeks.
But what is the minimal interval from previous dose?
Also,  what happens if first dose was not given  until child was 9 months old?

adamhstewart:
> But what is the minimal interval from previous dose?

4 weeks

> Also, what happens if first dose was not given until child was 9 months old?

They should be prompted for a second dose after 4 weeks and if not given, should be considered behind in the IPV vaccination.

anvmn:
@adamhstewart 

There may be a children that got first IPV dose on time, and still attending SPV encounters.
Those children can be 5 years old already, or even more than that. 

So, is there a maximal age after which we do not want to administer second dose?

adamhstewart:
@anvmn no max

anvmn:
Deployed to live

PR #1428, opened 2025-02-23
Acute Illness - second IPV dose for Rwanda site

anvmn:
#1426


anvmn:
Peek 2025-02-23 16-50

Issue #1439, opened 2025-03-11
Errors on `ddev restart` [3h]


amitaibu:
![Image](https://github.com/user-attachments/assets/9d470e28-fe42-4b0a-bf6a-108ce15d0dd9)

```
ddev version                            
 ITEM             VALUE                                    
 DDEV version     v1.24.3                                  
 architecture     amd64                                    
 cgo_enabled      0                                        
 db               ddev/ddev-dbserver-mariadb-10.11:v1.24.3 
 ddev-ssh-agent   ddev/ddev-ssh-agent:v1.24.3              
 docker           27.3.1                                   
 docker-api       1.47                                     
 docker-compose   v2.33.1                                  
 docker-platform  linux-docker                             
 global-ddev-dir  /home/amitaibu/.ddev                     
 mutagen          0.18.1                                   
 os               linux                                    
 router           ddev/ddev-traefik-router:v1.24.3         
 web              ddev/ddev-webserver:v1.24.3 
```

PR #1440, opened 2025-03-12
Update DDEV inside CI

AronNovak:
#1439

### DIFF
--- a/.ddev/web-build/Dockerfile
+++ b/.ddev/web-build/Dockerfile
@@ -32,7 +32,7 @@ RUN chmod +x /usr/bin/elm
 # gulp for the client compilation.
 RUN npm install -g gulp
 # For Elm Unit tests
-RUN npm -g config set user root
+ENV NPM_CONFIG_UNSAFE_PERM=true
 RUN npm install -g elm-test@0.19.1-revision6
 # Install svg2elm.
 RUN npm install -g svg2elm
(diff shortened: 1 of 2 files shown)


AronNovak:
@amitaibu Now the project loads with latest DDEV inside CircleCI. Your error message was a different one, likely it was a temporary problem. Or are you on an `aarch64` architecture?

amitaibu (approved):
Thanks.

(I'm on Ubuntu)

amitaibu:
@AronNovak I've seen this error before, not related to this PR. But maybe you have an idea why I get it on `ddev start`

```
Failed to restart eheza-app: unable to start project eheza-app because the configured database type does not match the current actual database. Please change your database type back to mariadb:10.2 and start again, export, delete, and then change configuration and start. To get back to existing type use 'ddev config --database=mariadb:10.2' and then you might want to try 'ddev debug migrate-database mariadb:10.5', see docs at https://ddev.readthedocs.io/en/stable/users/extend/database-types/ 
```

amitaibu:
(Running `ddev config --database=mariadb:10.2` does solve it for me. But not sure why we get this error at all)

AronNovak:
@amitaibu 
See https://github.com/TIP-Global-Health/eheza-app/pull/1342

You can do `ddev delete --omit-snapshot` (it will discard all the data in the DB of course) and start it over, it will never complain about DB version mismatch, well, until the next upgrade.

PR #1441, opened 2025-03-16
WIP: IHP integration

amitaibu:


client/src/js/app.js
```diff
@@ -740,7 +740,6 @@ elmApp.ports.askFromIndexDb.subscribe(function(info) {
               }
 
               const changes = {
-                'uuid': json.uuid,
```
anvmn: @amitaibu 

As far as I can tell from code, there is no need for `uuid` to be set here.
Is there a reason you added it?

amitaibu: Can be removed

Issue #1442, opened 2025-03-17
Update Resilience Survey Schedule


Diama1:
### Issue

Currently, surveys are scheduled to appear:  
-  At the start of the program.  
- 3 Months After the first survey.  
- 6 Months Into the program.  

During the pilot, we discovered that the program duration is **29 weeks and 5 days**, rather than just 6 months.  

**New Survey Schedule**
We need to adjust the survey schedule as follows:  
1. **1st Survey:** At the beginning of the program.  
2. **2nd Survey:** At **week 16, day 1** from the starting date.  
3. **3rd Survey (Final):** At **week 29, day 5** from the starting date.  

#### To-Do
- Update the logic used for survey schedule to reflect the new timing.

anvmn:
Deployed to resilience env

PR #1444, opened 2025-03-27
RM - Update the Resilience Survey Schedules

Diama1:
#1442


anvmn (changes requested):
This looks wrong.
The idea should be to look back to date of last survey, and decide, based on time that has passed since then.

anvmn:
Replaced by #1445. Closing.

Issue #1451, opened 2025-04-17
Create an ANC Diagnoses Report [4h]


adamhstewart:
## Issue

We need an additional report, exactly like the AI Diagnoses report, but for diagnoses made during ANC. 

The generated report for this is here: https://github.com/TIP-Global-Health/eheza-app/blob/main/server/hedley/modules/custom/hedley_admin/scripts/generate-acute-illness-report.php#L99

## Task
- Create the report

anvmn:
@adamhstewart 

So, it should look like this:

![Image](https://github.com/user-attachments/assets/670b724b-f232-41bb-85bb-c1b2e18a77f2)

Should we make this a separate report, or add as part on ANC we got right now?

adamhstewart:
@anvmn 

> So, it should look like this

Yes

> Should we make this a separate report

Separate: ANC Diagnoses

anvmn:
@adamhstewart deployed to [CHW env](https://chw-ihangane.pantheonsite.io/)

PR #1455, opened 2025-04-20
Statistical queries -  ANC Diagnoses report

anvmn:
#1451


anvmn:
Peek 2025-04-21 18-20

server/hedley/modules/custom/hedley_person/scripts/migrate-children.php
```diff
@@ -22,7 +22,7 @@
 $batch = drush_get_option('batch', 50);
 
 // Get allowed memory limit.
-$memory_limit = drush_get_option('memory_limit', 500);
+$memory_limit = drush_get_option('memory_limit', 800);
```
AronNovak, line 25: Let's verify if DDEV has the same (elevated) memory limit like the new Pantheon limit.
https://github.com/Gizra/drupal-starter/blob/main/.ddev/php/memory-limit.php.ini

AronNovak (approved):

server/hedley/modules/custom/hedley_person/scripts/migrate-children.php
anvmn: Done

Issue #1473, opened 2025-05-07
Show message to provide nutritional supplement for patients with MUAC less than 23 [2h]


adamhstewart:
## Issue

In cases where MUAC is less than 23, ass an additional nmessage in red to provide nutritional supplement

<img width="771" alt="Image" src="https://github.com/user-attachments/assets/6d02f510-a03b-40f1-9950-bf6a0958695c" />

## Task
Add new message

anvmn:
@adamhstewart 

This got some inconsistency with the color alert.
22 is still an ok value. Yellow alert is displayed when muac < 22.

anvmn:
@adamhstewart 

Deployed to Nutrition env

adamhstewart:
> 22 is still an ok value. Yellow alert is displayed when muac < 22.

Adjust the nutrition supplement message to appear at only less than 22 (matching the yellow/red message)

anvmn:
@adamhstewart 

> > 22 is still an ok value. Yellow alert is displayed when muac < 22.
> 
> Adjust the nutrition supplement message to appear at only less than 22 (matching the yellow/red message)

Implemented and deployed to Nutrition env

PR #1474, opened 2025-05-07
ANC -  nutritional supplement alert for MUAC

anvmn:
#1473

### DIFF
--- a/client/src/assets/scss/_new.scss
+++ b/client/src/assets/scss/_new.scss
@@ -2367,6 +2367,13 @@ div.page-activity.prenatal {
             color: $color-green;
           }
         }
+
+        .nutritional-supplement-alert {
+          color: $color-red-alert;
+          font-size: 16px;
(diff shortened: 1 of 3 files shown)


anvmn:
Selection_1422

client/src/assets/scss/_new.scss
```diff
@@ -2367,6 +2367,13 @@ div.page-activity.prenatal {
             color: $color-green;
           }
         }
+
+        .nutritional-supplement-alert {
+          color: $color-red-alert;
```
AronNovak, line 2372: Do we have `color-red` and `color-red-alert` separately?

AronNovak (approved):

client/src/assets/scss/_new.scss
anvmn: @AronNovak  Yes
Selection_1424

Issue #1461, opened 2025-04-29
Calculate Pre-preganancy Baseline BMI and display on nutrition assessment screen [2h]


adamhstewart:
## Issue

As a followup to #1456 - also calculate the baseline BMI and show it on the nutrition assessment page:

<img width="810" alt="Image" src="https://github.com/user-attachments/assets/e061356b-a793-4e1d-8384-8dbac93fee06" />

Show both the baseline BMI and the category (underweight, normal, overweight, obese) 

## Task
- Do the calculation
- Display underneath the BMI field - `Pre-pregnancy Baseline BMI: 20.2 - normal`

anvmn:
@adamhstewart 

Pregnancy Dating activity is usually first activity to be filled.
BMI calculation requires height. Height it taken as part of Nutrition assessment task at Examination activity.

So, as I see it, in common case scenario, there won't be enough data to display BMI, unless we add Height input on Pregnancy Dating activity.

Also, examination activity only exists for nurses. So, CHW might not have any info about height.

adamhstewart:
@anvmn, this has been updated to go on the nutrition assessment page.

anvmn:
@adamhstewart 

> [@anvmn](https://github.com/anvmn), this has been updated to go on the nutrition assessment page.

Implemented and deployed to Nutrition env

Issue #1456, opened 2025-04-23
Add Baseline pre-pregnancy weight [6h]


adamhstewart:
## Issue

We need to determine a pregnant woman's baseline. prepregnancy weight in order to track weight gained throughout the pregnancy.

This question will be asked on the Pregnancy Dating activity:

<img width="1166" alt="Image" src="https://github.com/user-attachments/assets/58d1d0a2-701b-47fb-b3a3-c0a68a8bb193" />

It should be required and only asked once in the first encounter and disappear for future encounters

## Tasks
- Create the field on the front end
- Create the infrastructure and syncing for the field on the backend.

anvmn:
@adamhstewart 

This issue is a bit more complex, since LMP activity has also a confirmation option.
That is, when LMP activity is filled first by CHW, and then patient arrives to Nurse and first nurse encounter is performed, we display LMP date recorded by CHW, and offer an ability to change it.
So now that we have additional fields, we'll need to support updating that one as well.

adamhstewart:
@anvmn bumped 2h per ^^

anvmn:
@adamhstewart 

Deployed to Nutrition env

adamhstewart:
@anvmn There something odd about this input field. When you try to remove the entry, it fills back in again.

https://www.loom.com/share/3c10d30c94914858b973ebfc47e52d76?sid=e80c1fac-b417-4e0a-a8d1-2543bb4c988c

anvmn:
@adamhstewart 

> [@anvmn](https://github.com/anvmn) There something odd about this input field. When you try to remove the entry, it fills back in again.

Fixed and deployed to Nutrition env

Diama1:
@adamhstewart @anvmn
As discussed, we need to add an option for cases where the mother does not know her pre-pregnancy weight.

We agreed that the first question should be:
“Does the mother know her pre-pregnancy weight?”

<img width="1204" height="508" alt="Image" src="https://github.com/user-attachments/assets/85125ae3-9c3e-4371-ac61-dcde3cea709f" />

If Yes, we proceed as usual and provide the field to enter the pre-pregnancy weight.

If No, we continue the workflow and use the weight collected during Examination → Nutrition Assessment as the baseline weight.

anvmn:
@adamhstewart @Diama1 

[...]

PR #1497, opened 2025-05-28
ANC-RCT fixes 01

anvmn:
#1456
#1461 
#1473

### DIFF
--- a/client/src/elm/Translate.elm
+++ b/client/src/elm/Translate.elm
@@ -119,6 +119,7 @@ import Pages.Prenatal.Activity.Types
         , GWGClassification(..)
         , HistoryTask(..)
         , MedicationTask(..)
+        , PrePregnancyClassification(..)
         , TreatmentReviewTask(..)
         )
 import Pages.Prenatal.Model exposing (HypertensionTreatementUpdateOption(..))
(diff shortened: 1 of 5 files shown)


anvmn:
Selection_1463

client/src/elm/Translate.elm
```diff
@@ -17548,6 +17569,23 @@ translationSet trans =
...
+                PrePregnancyNormal ->
+                    translationSet Normal
+
+                PrePregnancyOverweight ->
+                    translationSet Overweight
+
+                PrePregnancyObesity ->
+                    { english = "Obesity"
+                    , kinyarwanda = Nothing
+                    , kirundi = Nothing
```
AronNovak, line 17586: As I see more translations as usual in this PR, I am wondering if we need all translations for this piece of feature or not.

AronNovak (approved):

client/src/elm/Translate.elm
anvmn: Usually, we don't wait for translations  to be completed.
@Diama1 completes them as follow ups.

AronNovak: Right, that's why I never see the whole process :smile:

Issue #1526, opened 2025-07-02
Changes to Next Steps - mild to moderate anemia [4h]


adamhstewart:
## Issue

<img width="961" height="723" alt="Image" src="https://github.com/user-attachments/assets/97aba541-8f35-4673-9f0a-c68824988843" />

Changes:

ACTION TO TAKE:

Reinforce adherence to: **Fefol 200mg/0.4 mg**
1 tablet by mouth daily

Have you counseled the patient on adherence to FeFol
- yes/no 

Reinforce adherence to **MMS**
1 tablet by mouth daily

Have you counseled  the patient on adherence to FeFol
- yes/no 

Have you counseled the patient to return to the health center to repeat the hemoglobin test in four weeks.
- yes/no

_There are no conditional follow-ups for no on either of these questions._

## Tasks
- Add the new questions and revise the adherence quidleines

PR #1533, opened 2025-07-06
ANC: Mild to moderate anemia medicaitons update

anvmn:
#1526

**Note: Requires thorough testing.**
- Initial and recurrent phases.
- Standalone and in combination with other medications (treatments for another diagnoses).


anvmn:
Selection_1583

AronNovak (approved):

Issue #1623, opened 2026-01-26
Remove conditional question "Why has the patient NOT had an ANC visit in her first trimester?" [2h]


adamhstewart:
# Issue

This is a revert of #1520

anvmn:
@adamhstewart 

Is this only for Healthy Start, or global?

adamhstewart:
@anvmn global

PR #1628, opened 2026-01-28
Remove Late First ANC Visit reason input

anvmn:
#1623


client/src/elm/Backend/Measurement/Model.elm
```diff
@@ -657,7 +657,8 @@ type alias LastMenstrualPeriodValue =
     , prePregnancyWeight : Maybe WeightInKg
     , confident : Bool
     , notConfidentReason : Maybe LmpDateNotConfidentReason
-    , lateFirstVisitReason : Maybe LateFirstANCVisitReason
+    , -- Deprecated, but kept for backward compatibility.
+      lateFirstVisitReason : Maybe LateFirstANCVisitReason
```
AronNovak, line 666: Do we need any kind of update hook to transform existing data to the new way?

AronNovak (approved):

Issue #1471, opened 2025-05-06
Add "Measurement Cannot be Taken" Box to Weight and Height Measurements [3h]


adamhstewart:
## Issue

This is a follow-up to: https://github.com/TIP-Global-Health/eheza-app/issues/1448

Instead of making these fields optional, we should add a checkbox that says "Unable to take measurement" for weight and height fields in nutrition and SPV

In this way, users will need to go through the activity and submit a response to be brought back to the activity tray.

## Tasks
- Replace optionality for height and weight with height with a checkbox
= replace optionality in these fields in ALL environments where optionality currently exists.

anvmn:
@adamhstewart 

Deployed to:
- Nutrition env, Rwanda site
- QA env, Burundi site

PR #1653, opened 2026-02-08
"Measurement Cannot be Taken" box for Weight and Height Measurements (for CHWs at  VHW and UVL)

anvmn:
#1471

Reverts TIP-Global-Health/eheza-app#1641


anvmn:
@adamhstewart 

Deployed to:
- Nutrition env, Rwanda site
- QA env, Burundi site

anvmn (commented):
## Code Review

Overall this is a well-structured PR. The refactoring of the PHP base class (`HedleyRestfulIndividualEncounter`) to handle `$fields`, `$multiFields`, and `$dateFields` generically is a nice DRY improvement. The function renames (`encodeActivityAsString` → `activityToString`, etc.) are clean and consistent.

Below are the issues I found:

### 1. Alphabetical ordering violation in Msg types (project convention)

In `Backend/NutritionEncounter/Model.elm`, `AddSkippedForm` and `RemoveSkippedForm` are placed after `CloseNutritionEncounter`, but alphabetically they should be:
```
= AddSkippedForm SkippedForm
| CloseNutritionEncounter
| HandleUpdatedNutritionEncounter (WebData ())
| RemoveSkippedForm SkippedForm
| SaveHeight ...
```

Same issue in `Backend/WellChildEncounter/Model.elm` — `AddSkippedForm` and `RemoveSkippedForm` should come before `CloseWellChildEncounter` and `HandleUpdatedWellChildEncounter`.

The corresponding `case` branches in `Backend/NutritionEncounter/Update.elm` and `Backend/WellChildEncounter/Update.elm` should also be reordered to match.

### 2. Missing translations for `UnableToTakeMeasurement`

```elm
UnableToTakeMeasurement ->
    { english = "Unable to take measurement"
    , kinyarwanda = Nothing
    , kirundi = Nothing
    }
```

This is a user-facing string shown to CHWs. Since the app is used in Rwanda and Burundi, Kinyarwanda and Kirundi translations should be provided.

### 3. Unnecessary `RemoveSkippedForm` when saving an actual measurement

In `Pages/Nutrition/Activity/Update.elm` (`SaveHeight` and `SaveWeight` handlers) and `Pages/WellChild/Activity/Update.elm` (same handlers), when saving an actual measurement value, a `RemoveSkippedForm` message is always dispatched — even if the form was never skipped. This creates unnecessary PATCH requests to the encounter. Consider only sending `RemoveSkippedForm` when `EverySet.member SkippedHeight/SkippedWeight skippedForms` is true (same guard already used in the skip branch).

### 4. `allowEndingEncounter` simplification in WellChild may be too aggressive

In `Pages/WellChild/Encounter/Utils.elm`, the old code handled multiple cases for pending activities:
- `[ WellChildNutritionAssessment ]` → could end if mandatory nutrition tasks were done
- `[ WellChildDangerSigns, WellChildNutritionAssessment ]` → could end if both were done
- etc.

The new code removes all `WellChildNutritionAssessment` cases, so if `WellChildNutritionAssessment` is the only pending activity, the encounter can no longer end (falls to `_ -> False`). While the skip checkbox should prevent this scenario in most cases, if a CHW navigates away without checking the box, they may be unable to close the encounter — which was previously possible (for non-PediatricCare) when mandatory tasks were fulfilled.

### 5. Unused `child` parameter in `mandatoryActivitiesCompleted`

In `Pages/Nutrition/Activity/Utils.elm`:
```elm
mandatoryActivitiesCompleted currentDate zscores features child isChw assembled db =
    allMandatoryActivities isChw
        |> List.all (activityCompleted currentDate zscores features isChw assembled db)
```

The `child : Person` parameter is declared but never used. Pre-existing issue, but worth cleaning up in this PR since the function signature is already being changed.

Issue #1601, opened 2026-01-04
Review and extend Simpletest coverage


anvmn:
Sub issue of #1606

Task:

1. Review classes that `extends HedleyWebTestBase`:
- Check if tests still valid.
- Extend test coverage to include new content types (we got over 220 types).
2. Suggest new tests.

PR #1654, opened 2026-02-09
Align nodes permissions

anvmn:
#1601

While wording on simpletests, found that permissions of several node types were off.
This PR aligns the permissions and adds a simpletest to make sure node types permissions are set properly.


anvmn (commented):
@copilot 

Add a simpletest that verifies that for each type of node:

administrator user got 5 permissions:
- Create new content
- Edit own content
- Edit any content
- Delete own content
- Delete any content

sync user got 3 permissions:
- Create new content
- Edit own content
- Edit any content

Issue #1665, opened 2026-02-17
Frontend: Family Encounter screen layout and activity pages (MUAC, Aheza) [6h]


adamhstewart:
## Context
Part of the Family Encounter epic. Implements the main encounter screen and the four sub-activity pages.

## Encounter Screen
Display the mother and her children (if any) with activity completion status:

- **Mother section**: Aheza Mother + MUAC checkboxes/status
- **Per-child section** (0–5): Aheza Child + MUAC checkboxes/status
- "+ Add Child" button (if < 5 children)

## Activity Pages

**Aheza Mother / Aheza Child:**
- Single **integer input field** for supplement amount
- Label: "Aheza Mother" / "Aheza Child" (not "FBF")
- No dropdown/distribution notice — just the integer amount
- On save: creates `family_fbf_mother` or `family_fbf_child` measurement

**MUAC Mother / MUAC Child:**
- Float input field in cm (reuse existing MUAC input pattern)
- On save: creates `family_muac_mother` or `family_muac_child` measurement

## Activity Completion Logic
- Each activity is complete when a measurement exists for the current encounter
- Encounter is complete when all activities for all participants are done (or explicitly closed)
- Follow existing pending/completed tab pattern

## Files
- `Pages/FamilyEncounter/Activity/` — new submodule (Model, View, Update)
- `Pages/FamilyEncounter/Overview/` — encounter overview/dashboard

## Acceptance Criteria
- [ ] Encounter screen shows mother and all children
- [ ] Each participant shows correct activity completion status
- [ ] Aheza input accepts integers only
- [ ] MUAC input accepts decimal cm values
- [ ] Measurements save correctly and associate with the right encounter + person
- [ ] Activities can be edited after initial save
- [ ] Can add children mid-encounter (up to 5)

adamhstewart:
@anvmn missing requirement I noticed from the screenshots: Aheza distribution is kg/month rather than packets per month

adamhstewart:
## QA Analysis
1. Note the issue above: https://github.com/TIP-Global-Health/eheza-app/issues/1665#issuecomment-3941952499

<img width="992" height="753" alt="Image" src="https://github.com/user-attachments/assets/c48939f5-eb0f-42e0-9625-6bce8ed91d64" />

2. The Aheza subactivity should be called: Aheza Mother (for the mother) and Aheza Child (for the child)
3. MUAC subactivity for the mother should not have the help text with the parameters of measurement
[...]

PR #1675, opened 2026-02-21
Family nutrition measurements

anvmn:
#1665


anvmn:
[Screencast from 2026-02-22 20-16-01.webm](https://github.com/user-attachments/assets/fd57c738-8bc2-4bf5-81d8-b7c96b219f13)

PR #1700, opened 2026-03-12
Add cron job to prune `cache_field`

AronNovak:
Add hook_cron to hedley_general that deletes cache_field entries older than 30 days, preventing unbounded table growth.


AronNovak:
We have some evidence from production backup that otherwise, this table becomes by far the largest in size.

anvmn (approved):

PR #1701, opened 2026-03-12
Prune advancedqueue entries older than 6 months in cron

AronNovak:
## Summary
- Add cleanup of `advancedqueue` table entries older than 6 months to the existing `hedley_general_cron()`, preventing unbounded queue growth.


AronNovak:
@anvmn I have more uncertainty about this one. Is it inline with what those queue items are supposed to achieve?

anvmn (approved):

PR #1709, opened 2026-03-19
[HotFix] "Overnight Operations" Jenkins job

anvmn:
https://github.com/Gizra/ihangane/issues/3227


anvmn:
Both overnight tasks are green:

image

PR #1777, opened 2026-06-04
Batch FBF session-attendance lookup into one query per clinic

### LINKED ISSUE #1776: Per-participant COUNT N+1 when computing FBF session attendance stats
## Problem

`hedley_stats_get_session_attendance_stats_by_health_center` (`hedley_stats/hedley_stats.module:2663`) computes "missed session" attendance by running a nested loop — **per participant × per period** — and issuing one database `COUNT` on every pass:

```php
foreach ($participants_ids as $participant_id) {     // :2740   P participants
  foreach ($periods as $period) {                    // :2769   3 periods
    $sessions_ids = $sessions_by_period[$period];
    if (!empty($sessions_ids)) {
      // :2788  one COUNT query per (participant, period)
      $count = hedley_activity_count_person_measurements_taken_during_sessions($child_id, $sessions_ids);
      if ($count == 0) {
        $missed_session[] = $expected_person;
      }
    }
    ...
```

and the callee (`hedley_activity/hedley_activity.module:461`) runs a two-JOIN `COUNT` each time:

```php
hedley_general_join_field_to_query($query, 'node', 'field_person');   // join 1
hedley_general_join_field_to_query($query, 'node', 'field_session');  // join 2
$query->condition('field_session...target_id', $sessions_ids, 'IN');
return $query->countQuery()->execute()->fetchField();
```

## Scaling

**P × 3** two-join COUNT queries per FBF clinic. A health center with 200–500 PMTCT participants → **600–1,500 COUNT queries** per stats recalculation, per clinic. Stats recalc runs across the population on sync (Advanced Queue), so this inflates queue processing time and DB load.

## Proposed fix

Replace the per-participant counting with **one grouped query per clinic**: select the distinct `(field_person_target_id, field_session_target_id)` pairs for `field_session_target_id IN (all sessions across all periods)`, once, then in PHP determine each `(participant, period)` attendance by in-memory set membership (does the participant have any measurement in that period's session set?).

**P × 3 queries → 1.**

The period-window guards (`join_date`/`graduate_date` vs the period range) and `completed_program` accounting stay unchanged; only the attendance COUNT moves to the single grouped query.

## Verification

Output (attendance / missed-session counts) must be preserved exactly. Pair with `ddev phpcs`, the SimpleTest suite, and a before/after comparison of the generated stats on a representative dataset.

## Context

Finding #2 from the backend report/stats performance review (proposal #4).

anvmn:
Fixes #1776

## Problem

`hedley_stats_get_session_attendance_stats_by_health_center` (`hedley_stats.module`) decided whether a child missed a session by running a **two-JOIN `COUNT`** query per **(PMTCT participant × period)** — `hedley_activity_count_person_measurements_taken_during_sessions()` once for every participant, for each of 3 periods.

For a clinic with **P** participants that's **P × 3** COUNT queries; a health center with hundreds of participants issues hundreds-to-thousands of queries per stats recalculation, which runs across the population on sync (Advanced Queue).

## Fix

Added `hedley_activity_get_persons_sessions_with_measurements($sessions_ids)` — a **single** query that returns, for the clinic, the `(person → set of session IDs they have measurements in)` map. The attendance loop now determines each participant's per-period attendance by **in-memory set membership**.

**P × 3 queries → 1 per clinic.**

## Behavior preservation

The batched query mirrors the COUNT helper's exact semantics — same `hedley_general_create_db_select_query_excluding_deleted('node')`, `status = published`, inner joins on `field_person` and `field_session`, and `field_session …_target_id IN (sessions)` — just without the per-person condition, selecting the (person, session) pairs instead of counting.

`isset($persons_sessions[$child_id][$session_id])` is set iff a matching published/non-deleted measurement exists, so the membership check is exactly equivalent to the old `count > 0`:
- A child with no measurement in any of a period's sessions → not attended → `missed_session` (same as `count == 0`).
- Extra persons in the map (mothers, non-participants) are never looked up.
- The period-window guards (`join_date`/`graduate_date`) and `completed_program` accounting are untouched — only the attendance COUNT moved.

The now-unused `hedley_activity_count_person_measurements_taken_during_sessions()` is left in place (no other callers) and can be removed in a follow-up.

## Testing

- `php -l` clean (both files).
- `ddev phpcs` → exit 0 (Drupal + DrupalPractice).
- Runtime equivalence relies on the **SimpleTest** suite (CI); reviewers may also want a before/after check of the attendance / missed-session counts on a representative dataset.

Finding #2 from the backend report/stats performance review (proposal #4).

### DIFF
--- a/server/hedley/modules/custom/hedley_activity/hedley_activity.module
+++ b/server/hedley/modules/custom/hedley_activity/hedley_activity.module
@@ -475,3 +475,45 @@ function hedley_activity_count_person_measurements_taken_during_sessions($person
 
   return $query->countQuery()->execute()->fetchField();
 }
+
+/**
+ * Loads, per person, the sessions during which they have measurements.
+ *
+ * Batched companion to
+ * hedley_activity_count_person_measurements_taken_during_sessions(): instead of
+ * one COUNT query per person, this resolves attendance for many persons at once
+ * with a single query, so callers can determine it in memory.
+ *
+ * @param array $sessions_ids
+ *   The session node IDs to consider.
+ *
+ * @return array
+ *   A map keyed by person node ID; each value is a set of session node IDs (as
+ *   array keys) during which that person has at least one measurement.
+ */
+function hedley_activity_get_persons_sessions_with_measurements(array $sessions_ids) {
+  $persons_sessions = [];
+  if (empty($sessions_ids)) {
+    return $persons_sessions;
+  }
+
+  $query = hedley_general_create_db_select_query_excluding_deleted('node');
+  $query->condition('status', NODE_PUBLISHED);
+
+  $person_field = 'field_person';
+  hedley_general_join_field_to_query($query, 'node', $person_field);
+
+  $session_field = 'field_session';
+  hedley_general_join_field_to_query($query, 'node', $session_field);
+  $session_field_name = $session_field . '.' . $session_field . '_target_id';
+  $query->condition($session_field_name, $sessions_ids, 'IN');
+
+  $query->distinct();
+
+  $result = $query->execute();
+  foreach ($result as $row) {
+    $persons_sessions[$row->field_person][$row->field_session] = TRUE;
+  }
+
+  return $persons_sessions;
+}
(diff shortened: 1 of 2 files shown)


anvmn:
## Manual test on production data

Validated against a ~5M-node production DB import. This branch keeps the old per-`(child, period)` COUNT helper alongside the new batched-map helper, so both were computed over the real data and compared directly:

- 37 FBF clinics, **7,554 `(participant, period)` checks, 0 mismatches** — the old `count == 0` and the new `isset($map[child][session])` membership agree in every case.

PR #1781, opened 2026-06-07
Drop redundant node_load when resolving encounter measurement IDs

### LINKED ISSUE #1780: Redundant node_load when resolving individual-encounter measurement IDs
## Problem

`hedley_person_load_individual_encounter_measurements_ids($nid)` (`hedley_person/hedley_person.module`) does `node_load($nid)` solely to read `$encounter->type` (to build the `field_<bundle>` join):

```php
$encounter = node_load($nid);
$bundle = $encounter->type;
$encounter_field = "field_$bundle";
...
```

But every caller already holds the fully-loaded encounter node (they pass `$encounter->nid` / `$node->nid`), so the extra `node_load` is wasted work. It's called once per encounter throughout report/completion generation (10 call sites).

## Fix

Add an optional `$bundle` parameter; when supplied, skip the `node_load` (fall back to it only when absent, for safety). Update the call sites to pass `$encounter->type`.

## Verification

`php -l`, `ddev phpcs`, and a real-data equivalence check: for a sample of encounters, confirm the bundle-supplied path returns the identical measurement-ID set as the `node_load` path.

Finding #4 from the backend report/stats performance review (proposal #4).

anvmn:
Fixes #1780

## Problem

`hedley_person_load_individual_encounter_measurements_ids($nid)` did `node_load($nid)` purely to read `$encounter->type` (to build the `field_<bundle>` join), even though all 10 call sites already hold the loaded encounter node. It runs once per encounter throughout report/completion generation.

## Fix

Add an optional `$bundle` parameter; when supplied, skip the `node_load` (kept as a fallback when absent). Pass `$encounter->type` (or `$node->type`) at the 10 call sites.

## Behavior preservation

Same query, same results — only the redundant load is removed. Verified `php -l` + `ddev phpcs` (Drupal + DrupalPractice). Real-data equivalence test below.

Finding #4 from the backend report/stats performance review (proposal #4).

### DIFF
--- a/server/hedley/modules/custom/hedley_person/hedley_person.module
+++ b/server/hedley/modules/custom/hedley_person/hedley_person.module
@@ -1331,7 +1331,7 @@ function hedley_person_process_deleted_update($node) {
     $encounter_types[] = 'session';
 
     if (in_array($node->type, $encounter_types) && $deleted) {
-      $measurements_ids = hedley_person_load_individual_encounter_measurements_ids($node->nid);
+      $measurements_ids = hedley_person_load_individual_encounter_measurements_ids($node->nid, $node->type);
       foreach ($measurements_ids as $measurement_id) {
         $item_wrapper = entity_metadata_wrapper('node', $measurement_id);
         $item_wrapper->field_deleted->set($deleted);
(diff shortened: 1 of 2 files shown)


anvmn:
## Real-data equivalence test

Against a ~5M-node production DB import. Because the function keeps a `node_load` fallback when `$bundle` is absent, both paths exist on this branch and were compared directly over real data:

For a sample of up to 80 encounters of **each** individual-encounter bundle (12 bundles: prenatal/acute-illness/well-child/nutrition/ncd/child-scoreboard/hiv/tuberculosis/home-visit/… + session):
- old path `f($nid)` (node_load → type) vs new path `f($nid, $bundle)`
- **860 checks, 0 mismatches — IDENTICAL** measurement-ID sets.

PR #1785, opened 2026-06-07
Stop silently truncating clinic sessions / PMTCT participants in stats

### LINKED ISSUE #1784: Stats: silent range(0,100) truncation of clinic sessions / PMTCT participants
## Problem (correctness)

`hedley_stats_get_clinic_sessions_by_period()` and `hedley_stats_get_pmtct_participants_by_clinic()` apply `->range(0, $range)` with a default `$range = 100`, and **silently** drop anything beyond the cap. The session-attendance caller uses the **default** for sessions (`hedley_stats_get_clinic_sessions_by_period($clinic_id, $period)`), so if a clinic ever has >100 sessions in a period, attendance / missed-session stats are computed on a truncated set — with no indication.

In practice the caps aren't reached today (periods are monthly; participants pass an explicit 5000), so this is a latent bug, but a silent one.

## Fix

- `get_clinic_sessions_by_period()`: default `$range` to `NULL` (no cap). Sessions per period are inherently bounded, so there's no memory risk and the result is never silently truncated.
- Both functions: when a cap **is** in effect and the result reaches it, log a `watchdog()` warning (clinic + count), so any future truncation is visible/actionable instead of silent. The participants memory guard (caller passes 5000) is kept.

## Verification

`php -l`, `ddev phpcs`, and a real-data check: confirm no clinic/period currently exceeds the old 100 cap (so the un-cap is behavior-identical on real data) and no clinic exceeds the 5000 participant guard.

Finding #6 (correctness) from the backend report/stats performance review (proposal #4).

anvmn:
Fixes #1784

## Problem (correctness)

`hedley_stats_get_clinic_sessions_by_period()` and `hedley_stats_get_pmtct_participants_by_clinic()` applied `->range(0, $range)` with a default `$range = 100` and **silently** dropped anything beyond it. The session-attendance caller uses the default for sessions, so a clinic with >100 sessions in a period gets stats computed on a truncated set, with no indication.

## Fix

- `get_clinic_sessions_by_period()`: default `$range` to `NULL` (uncapped). Sessions per period are inherently bounded → no memory risk, never silently truncated.
- Both functions: when a cap is in effect and reached, log a `watchdog()` warning (clinic + count) so any future truncation is visible. The PMTCT-participants memory guard (caller passes 5000) is kept.

## Real-data result

See the test comment. Short version: identical on the monthly periods that callers actually use; the un-cap **corrects a real silent truncation** in the (default) one-year path (a clinic with 135 sessions was being capped to 100); the participant guard is never reached (max 2293 < 5000).

`php -l` + `ddev phpcs` (Drupal + DrupalPractice) clean.

Finding #6 (correctness) from the backend report/stats performance review (proposal #4).

### DIFF
--- a/server/hedley/modules/custom/hedley_stats/hedley_stats.module
+++ b/server/hedley/modules/custom/hedley_stats/hedley_stats.module
@@ -3698,8 +3698,20 @@ function hedley_stats_get_pmtct_participants_by_clinic($clinic_id, $range = 100)
     ->range(0, $range);
 
   $result = $query->execute();
+  $ids = !empty($result['node']) ? array_keys($result['node']) : [];
 
-  return !empty($result['node']) ? array_keys($result['node']) : [];
+  if (count($ids) >= $range) {
+    // Results reached the cap and may be truncated. Log it rather than
+    // silently dropping participants from the statistics.
+    watchdog(
+      'hedley_stats',
+      'Clinic @clinic PMTCT participants hit cap @range; may be truncated.',
+      ['@clinic' => $clinic_id, '@range' => $range],
+      WATCHDOG_WARNING
+    );
+  }
+
+  return $ids;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


anvmn:
## Real-data test

Against a ~5M-node production DB import. Because the function now accepts both `NULL` (uncapped) and an explicit range, both behaviours were compared directly over real data — across all 31 FBF clinics × 4 periods:

**Sessions — uncapped (new default) vs old 100-cap:**
- Monthly periods (`two-months-ago`, `last-month`, `this-month`) — the only periods any caller actually uses: **93 checks, 0 mismatches → IDENTICAL.**
- `one-year` period (the function default, currently unused by callers): **1 clinic (`7105`) had 135 sessions silently capped to 100** by the old code. The fix returns all 135 — i.e. it corrects a real, previously-silent truncation. This is the intended correctness fix and affects no live computation today (no caller uses the one-year period here).

**PMTCT participants:** max in any clinic = **2293 < 5000** guard → the cap is never reached; output unchanged. (The participants memory guard is retained, now with a watchdog if it's ever hit.)

`php -l` + `ddev phpcs` (Drupal + DrupalPractice) clean.

PR #1787, opened 2026-06-07
Use exclusive upper bound for stats period ranges (fix boundary double-count)

### LINKED ISSUE #1786: Stats period boundary double-count: inclusive upper bound where end == next start
## Problem (correctness)

`hedley_stats_get_range_for_period()` sets each period's `end` to the *next* period's `start` (both at midnight) — e.g. last-month `end` = `last day of previous month + 1 day` = first-of-this-month 00:00:00 = this-month `start`. `hedley_stats_get_period()` then formats both to `Y-m-d 00:00:00`.

Consumers use an **inclusive** upper bound:
- `hedley_stats_get_base_query()`: `condition(date, [start, end], 'BETWEEN')` (inclusive both ends).
- `hedley_stats_get_clinic_sessions_by_period()`: `field_scheduled_date <= end`.

So a record timestamped exactly at a boundary midnight (common for date-only fields such as session `field_scheduled_date`, where the time is `00:00:00`) matches **both** adjacent periods → it is double-counted, and mis-attributed to the earlier period.

## Fix

Make the upper bound **exclusive** (`>= start AND < end`). A boundary record (e.g. 1st of the month 00:00:00) then belongs only to the period that *starts* there, which is the correct attribution, and adjacent periods become disjoint.

## Verification

`php -l`, `ddev phpcs`, and a real-data check: for adjacent periods, the inclusive bound double-counts N boundary records while the exclusive bound has zero overlap and preserves the union (no record lost, just attributed once).

Finding #7 (correctness) from the backend report/stats performance review (proposal #4).

anvmn:
Fixes #1786

## Problem (correctness)

`hedley_stats_get_range_for_period()` sets each period's `end` to the next period's `start` (both at midnight). Two consumers used an **inclusive** upper bound:
- `hedley_stats_get_base_query()` — `BETWEEN [start, end]`
- `hedley_stats_get_clinic_sessions_by_period()` — `field_scheduled_date <= end`

So a record timestamped exactly at a boundary midnight is counted in **both** adjacent periods (and mis-attributed to the earlier one). Date-only fields (e.g. session `field_scheduled_date`, time `00:00:00`) make this common.

## Fix

Make the upper bound **exclusive** (`>= start AND < end`). A boundary record then belongs only to the period that *starts* there (the correct attribution), and adjacent periods are disjoint.

## Real-data result

See the test comment — at a single month boundary, the old inclusive bound double-counted **27 sessions** and **337 measurements**; the fix brings the overlap to **0** while preserving the union (no records lost, just counted once).

`php -l` + `ddev phpcs` (Drupal + DrupalPractice) clean.

Finding #7 (correctness) from the backend report/stats performance review (proposal #4).

### DIFF
--- a/server/hedley/modules/custom/hedley_stats/hedley_stats.module
+++ b/server/hedley/modules/custom/hedley_stats/hedley_stats.module
@@ -3525,10 +3525,11 @@ function hedley_stats_get_base_query($health_center_id, $node_type, $period = FA
     $date_field_name = $date_field . '.' . $date_field . '_value';
 
     $date = hedley_stats_get_period($period, $extend);
-    $query->condition($date_field_name, [
-      $date['start'],
-      $date['end'],
-    ], 'BETWEEN');
+    // Lower bound inclusive, upper bound exclusive: a period's end equals the
+    // next period's start (both at midnight), so an inclusive BETWEEN would
+    // count a boundary record in both adjacent periods.
+    $query->condition($date_field_name, $date['start'], '>=');
+    $query->condition($date_field_name, $date['end'], '<');
   }
 
   return $query;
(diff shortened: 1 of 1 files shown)


anvmn:
## Real-data test

Against a ~5M-node production DB import. For the two adjacent periods that share a midnight boundary (last-month `end` == this-month `start` == `2026-06-01 00:00:00`), compared the inclusive (`<=`, old) vs exclusive (`<`, new) upper bound via raw queries on the relevant date fields:

| date field | OLD `<=` boundary overlap (double-counted) | NEW `<` overlap | union preserved |
|---|---|---|---|
| sessions `field_scheduled_date` | **27** | **0** | yes (2770) |
| measurements `field_date_measured` | **337** | **0** | yes (17578) |

So at a single month boundary the old code double-counted 27 sessions and 337 measurements; the fix removes the overlap entirely and the **union is unchanged** — boundary records are simply counted once, attributed to the period that starts at the boundary (the correct period). This is a behavior change confined to boundary-midnight records (the bug); all other records are unaffected.

`php -l` + `ddev phpcs` (Drupal + DrupalPractice) clean.

PR #1997, opened 2026-07-26
Measurements: make sure birth weight is entered in grams

### LINKED ISSUE #1981: Measurements: low birth weight indicator is ~90% false positives: kilograms are being entered into a grams field
Found while investigating #1980. **Verified against live production data on ihangane.live** (read-only queries).

## The indicator is wrong

`hedley_ncda.module` computes:

```php
$data['low_birth_weight'] = $birth_weight < 2000;
```

`field_birth_weight` is stored in **grams** (`WeightInGrm`, displayed with `Translate.Grams`). But the input has no range check, so kilograms get entered into it — and any kilogram value is trivially below 2000.

Live breakdown of the 14,048 birth weight records:

| Bucket | Records |
|---|---|
| under 10 — kilograms entered, **falsely** flagged low | **1,817** |
| 10-499 — implausible, also falsely flagged | 165 |
| 500-1,999 — genuinely low birth weight | **229** |
| 2,000+ — not flagged | 11,837 |

**Of 2,211 records currently flagged low birth weight, only 229 are genuine — about 90% are false positives.** The indicator over-reports roughly ninefold.

## Evidence the sub-10 values are kilograms

The most common values under 10 are 3.0 (298), 2.8 (146), 3.2 (133), 3.5 (127), 2.9 (113), 2.5 (103), 3.3 (90), 3.1 (90), 3.4 (79), 3.8 (73). That is a textbook newborn weight distribution in kg. Mean of the whole column is 3,029 g, confirming grams is the intended unit.

Also present: **145 records above 10,000 g**, up to a maximum of **350,022** (350 kg).

No birth weight records exist on the Burundi site yet, so this is Rwanda-only today — but the same unguarded input ships to both.

## Two parts to fix

**1. Stop new bad data.** Birth weight needs its own constraint. ⚠️ **The existing weight constant cannot be reused**: `getInputConstraintsWeight` is `{ minVal = 0.5, maxVal = 200 }` in **kilograms**. Wiring it here would reject every genuine newborn (3,000 > 200) while happily accepting all 1,817 wrong kg entries (3.0 sits inside 0.5-200) — locking in the corruption and blocking correct data. A gram-scaled constant is needed, roughly 500-6,000.

Worth considering whether the field should accept kg and convert, given that nurses are demonstrably entering kg — 13% of the time. That is a product decision, not just a validation one.

**2. Decide what to do with the existing records.** They fall into distinct groups:
- ~1,817 sub-10 values are **mechanically recoverable** (multiply by 1,000), though that is an assumption about intent and should be confirmed before running it.
- 165 in the 10-499 range and 145 above 10,000 are **not recoverable** — there is no way to know what was meant.
- Until they are corrected, any report using `low_birth_weight` is unreliable.

## Related

- #1980 — the wider missing-range-gate sweep
- #1917 — added the constraint helpers this would extend

anvmn:
Fixes #1981

## The problem

Birth weight is stored in grams, but the field accepts anything: a weight typed in kilograms is read as grams. On the Rwanda site **1,817 of 14,048** birth weights are kilograms, and since every one of them is below 2000 they all count as a low birth weight. Of the 2,211 records the indicator reports, only **229** genuinely are — roughly a ninefold over-report.

## Changes

- **`getInputConstraintsBirthWeight` — 300 to 7000 grams.** Wide enough for the extremes that do occur: babies have survived under 500 g, and a very large baby can be over 6000 g, and refusing those would leave the nurse no way to record the real weight. The existing `getInputConstraintsWeight` is kilograms (`{0.5, 200}`) and cannot be used here: it would refuse every real newborn (3000 > 200) while accepting every wrong kilogram entry (3.0 sits inside 0.5–200), locking the corruption in. The docblock says so at the definition.
- **`birthWeightOutsideConstraints`** — a pure predicate in `Measurement/Utils.elm`, so the rule is unit-testable rather than buried in a view. Unlike `outsideConstraints`, a weight that has not been entered is not reported: whether the measurement still has to be taken is the task count's job.
- **A warning popup instead of saving**, following the `PreSave…` pattern already used for head circumference. Closing it leaves the form as it was so the weight can be entered again; nothing is saved until it is in range.

## Where the weight is asked

| Path | Gate |
|---|---|
| WellChild newborn exam (Pregnancy Summary) | `PreSavePregnancySummary` branches to popup or save |
| NCDA — WellChild, ChildScoreboard, Nutrition **and group sessions** | gated once inside the shared `viewNCDAContent` |

In the NCDA form the weight is asked on the first step but saved on the last, so **every step forward** is stopped rather than only the save — otherwise the step tabs walk a bad value straight past the gate. The gate applies only when the weight is actually shown on that form, so a form that does not ask for it can never become impossible to leave.

## Verification

Elm 0.19.2, cold: `elm make` 548 modules, `elm-test` **2902 passing** (+9 new), `elm-review` clean, `elm-format` clean.

Discrimination test — swapping in the kilogram range reproduces the exact failure the issue warns about:

```
✗ an ordinary birth weight in grams is inside the range    (3000 g rejected)
✗ the same weight typed in kilograms is outside the range  (3 accepted)
```

## Known limitation

This stops **new** bad data. It does not correct the ~1,982 existing bad records, so `hedley_ncda.module:530` (`low_birth_weight = $birth_weight < 2000`) keeps over-reporting until those are dealt with — part 2 of #1981. Related: where a newborn exam already recorded the weight, NCDA does not ask for it, so an NCDA re-save can still carry a pre-existing bad value through.


anvmn:
[video.webm](https://github.com/user-attachments/assets/05a9b061-3b21-4b6a-9d94-67c1597c4d31)

Issue #2006, opened 2026-07-27
META: Every anthropometric measurement refuses an implausible value and says why


anvmn:
## Goal

**Every anthropometric measurement refuses an implausible value and says why.**

#1997 established the mechanism for birth weight: type a value outside its range, press the button to go on, and a warning explains the range — nothing is saved until it is corrected. This meta issue tracks applying that same mechanism to the rest.

## Where we are today

Two worse behaviours are in production:

**Silently disabled button — 5 sites.** The value is refused, but nothing says why. #1980 describes what the nurse sees: *"3/3 tasks complete" beside a Save button that will not respond, with nothing explaining it.*

| Measurement | Site |
|---|---|
| height, weight, MUAC | Nutrition encounter |
| height, weight, MUAC | Well Child nutrition assessment |
| MUAC | Acute Illness physical exam |
| MUAC | Family Nutrition — and it never states the range at all |
| height, weight, MUAC | Group session |

**No check at all — 4 places.** The value simply saves.

| Measurement | Site |
|---|---|
| height, weight, MUAC | Prenatal — `grep outsideConstraints` returns zero hits across `Pages/Prenatal/` |
| weight, MUAC | NCDA form — reached from four call sites, the widest single gap |
| birth length | Well Child newborn exam |
| fundal height | Prenatal obstetrical exam |

## Decisions taken (2026-07-27)

1. **Anthropometrics only** — height, weight, MUAC, birth length, fundal height. Not vitals, not lab values.
2. **Convert the five silent-disable sites too**, so there is one behaviour rather than two.
3. **Popup only** — an out-of-range value is not additionally counted as an incomplete task (#1986's approach). See *Known limitation*.
4. **Head circumference is out of scope** — it keeps its existing macro/microcephaly z-score popup.
5. **New ranges: birth length 15–60 cm, fundal height 1–60 cm.** Deliberately wide — they catch typing mistakes, not clinical extremes, which are already reported separately.

## The work

Each issue delivers one nurse-visible functionality and carries the tests that prove it, e2e wherever the behaviour is reachable through the UI. Shared helpers are added by the issue that first needs them; later issues are based on that branch rather than duplicating them.

- [ ] #1998 — The newborn exam checks birth length ← **adds the shared helpers; start here**
- [ ] #1999 — Well Child measurements say why they are refused *(based on #1998)*
- [ ] #2000 — Nutrition encounter measurements say why they are refused *(based on #1998)*
[...]

Issue #1998, opened 2026-07-27
Measurements: make sure birth length is entered in centimetres


anvmn:
Part of #2006. **Carries the shared helpers the rest of the series is based on.**

## What the nurse gets

Entering an implausible birth length on the Well Child newborn exam shows a warning explaining the range and refuses to save, exactly as birth weight already does since #1997.

## Why this one first

It is the first form where **two** ranged measurements sit behind one button, so it forces the shared pieces into the right shape straight away.

## What it adds

- `AnthropometricMeasurement` tag in `Measurement/Model.elm`
- `anthropometricOutOfRange : Site -> AnthropometricMeasurement -> Maybe Float -> Bool` in `Measurement/Utils.elm`, with **popup semantics — an unset value is NOT out of range** (the task count already answers "not taken yet")
- `measurementOutOfRangePopup` taking a **`List`**, not a `Maybe` — later forms put three ranged measurements behind one button, and the nurse must be told about each in turn
- `getInputConstraintsBirthLength` — **15–60 cm**
- Composes the message from `Translate.AllowedValuesRangeHelper` (already translated in all four languages) plus the measurement name, so almost no new untranslated text

`birthWeightOutsideConstraints` and `birthWeightOutOfRangePopup` become thin wrappers, so #1997's tests pass **unedited** — the cheapest proof the refactor changed nothing. Keep the `birth-weight-out-of-range` CSS class among the popup's classes so the existing e2e selector still matches.

## Where

Extends the existing `PreSavePregnancySummary` branch (`Pages/WellChild/Activity/Update.elm:181`). Gate only when `form.birthLengthAvailable == Just True` (`Pages/WellChild/Activity/View.elm:344`).

## Tests

- unit tests for the predicate and the new constant, beside `birthWeightOutsideConstraintsTest` (`Measurement/Test.elm:285`)
- **e2e** in `well-child-encounter-chw.spec.ts` next to the existing birth-weight case, reusing `expectBirthWeightInKilogramsRefused` (`client/e2e/helpers/common.ts`) as the template: bad length -> popup -> form still there -> good length -> saves
- discrimination-test the gate: break the predicate, the new tests must fail, restore

## Notes

Review this one hardest — every later issue copies its shape.

Related: #1980, #1997

anvmn:
Delivered by #2007, which reached `develop` today in the #2007 merge (`ade7d38f9`).

[...]

Issue #1980, opened 2026-07-22
Half the MUAC, height and weight inputs have no range check at all


anvmn:
Found while working #1978.

## What exists

`Measurement/Utils.elm:82-107` defines the constraint records — height 25-250, weight 0.5-200, MUAC 5-99 cm (50-999 mm for Burundi). `outsideConstraints` and `muacOutsideConstraints` (`Measurement/Utils.elm:280-297`) compare against them, and pages use the result to keep the Save button disabled. Both helpers were added by #1917.

`muacOutsideConstraints` exists because MUAC is stored in cm while the Burundi constraints are in mm — it applies `muacValueForSite` before comparing, so the conversion cannot be forgotten.

## Where they are wired — and where they aren't

**Gated:**

| Kind | Module | Gate |
|---|---|---|
| height, MUAC, weight | Nutrition encounter | `Pages/Nutrition/Activity/View.elm:258, 287, 448` |
| height, MUAC, weight | WellChild nutrition assessment | `Pages/WellChild/Activity/Utils.elm:475, 483, 491` |
| MUAC | AcuteIllness physical exam | `Pages/AcuteIllness/Activity/Utils.elm:384` |
| MUAC | FamilyNutrition encounter | `Pages/FamilyNutrition/Encounter/View.elm:556` |
| height, weight, MUAC | Group session | `Measurement/View.elm:383-390` — differs from the rest: it returns `Nothing` for `saveMsg` rather than adding a disjunct to `disabled` |

**Not gated:**

| Kind | Module | Input |
|---|---|---|
| height, weight, MUAC | **Prenatal** → Examination → NutritionAssessment | `Pages/Prenatal/Activity/View.elm:3357, 3392, 3442` |
| weight, MUAC, birth weight | **shared NCDA form** | `Measurement/View.elm:3272, 3341, 3665` |
| all of it | **ChildScoreboard** | its only capture is the NCDA form (`Pages/ChildScoreboard/Activity/View.elm:140-148`) |
| birth weight, birth length | **WellChild** pregnancy summary | `Pages/WellChild/Activity/View.elm:327-337, 348-360` |

Grepping `outsideConstraints|withinConstraints|getInputConstraints` returns **zero hits** across both `Pages/Prenatal/` and `Pages/ChildScoreboard/`.

The NCDA form is the widest single gap: those three inputs are reached from four call sites — Nutrition, WellChild, ChildScoreboard, and the group session.

## What happens today when a value is out of range

[...]

PR #2012, opened 2026-07-27
Measurements: make sure Well Child height, weight and MUAC are entered in the right units

### LINKED ISSUE #1999: Measurements: make sure Well Child height, weight and MUAC are entered in the right units

## What the nurse gets

Height, weight and MUAC in the Well Child nutrition assessment explain the allowed range instead of quietly deadening the Save button. #1980 describes today's experience: *"3/3 tasks complete" beside a Save button that will not respond, with nothing explaining it.*

**No change to what can be saved** — these values are already refused today. Only the feedback changes.

## Where

- delete `nutritionAssessmentSaveDisabled` (`Pages/WellChild/Activity/Utils.elm:469-492`)
- `disabled` at `Pages/WellChild/Activity/View.elm:927` goes back to `tasksCompleted /= totalTasks`
- add a `PreSave*` per task, following the Issue 1 template

## Tests

- unit tests on `update` per task
- **e2e** in `well-child-encounter-nurse.spec.ts`

Related: #1980

anvmn:
Fixes #1999

Based on #2007 — **please merge that first**, and this PR's base moves to `develop` once it lands.
The diff shown here is against `i1998-birth-length-range`, so it is only this issue's work.

Part of #2006.

## What a nurse gets

Height, weight and MUAC on the Well Child nutrition assessment deadened the Save button when the
value was outside the range it can take, with nothing on screen to say why — the task count read
3/3 and the button did not answer. That is what #1980 describes.

The measurement is named on a warning instead, the way the newborn exam already does for birth
weight and birth length, and the button goes back to answering the task count alone.

**What can be saved does not change** — these values were refused before.

## The one design decision worth reading

MUAC's range depends on the site, so `anthropometricOutOfRange` needed a site.

I first tried the smaller change: let the `MeasurementMuac` variant carry its own site, which is
what #2007's docblock originally described. **The module graph forbids it** —
`Measurement.Model` → `SyncManager.Model` → `Backend.Session.Model` → `Measurement.Model` is a
cycle, and `elm make` says so. So the site is passed alongside, and that reaches
`anthropometricConstraints`, `birthWeightOutsideConstraints`, `birthWeightBlocksNCDAForm`,
`measurementOutOfRangePopup` and their tests. #1998's "the existing tests pass unedited" does not
survive this issue.

MUAC is also held on the forms in centimetres and shown in millimetres at Burundi. That
conversion now happens inside `anthropometricOutOfRange`, so the next caller cannot forget it —
getting it backwards makes every Burundi MUAC pop up.

## Two behaviours to check are what you want

- A measurement the nurse said **could not be taken** is not asked about — as the old gate did.
- An **unset** measurement is no longer reported. It used to keep Save disabled; the task count
  already says the measurement is still to be taken, and reporting it would bring back the very
  button that will not answer. The button is still disabled by the task count until it is filled.

## Tests

- The ten `nutritionAssessmentSaveDisabled` tests are rewritten as nine on
  `nutritionAssessmentOutOfRange`, keeping the Burundi-holds-cm / Rwanda-rejects-125 unit cases.
- Five new tests on `update` itself, covering what the predicate cannot: that the gate is wired
  to the button, names the measurement on the popup, and saves nothing.
- **e2e** in `well-child-encounter-nurse.spec.ts`: each of the three measurements is entered
  wrong first, and the run only continues if the warning names it and the form is still there.

`elm make src/elm/Main.elm` — 548 modules. `elm-format --validate` clean. **2913 tests pass**
(2902 on `develop`). The e2e passes in 1.3m.

**Discrimination-tested:** making MUAC never report fails 2 tests; ignoring "could not be taken"
for height fails 1; restored, all 2913 pass. The e2e is self-discriminating — its helper blocks
on `div.ui.active.modal.measurement-out-of-range`, which nothing but the gate produces.


anvmn (commented):
Local `/code-review medium` findings, all low severity. Recorded here before anything is changed.

The review chased and **ruled out** the things that looked most dangerous, which is worth having on the record:

- Removing the range gate from `disabled` does **not** regress the CHW "unable to take measurement" path — `heightFormAndTasks`/`weightFormAndTasks` return `[maybeToBoolTask form.measurementNotTaken]` when ticked, so the task count completes and Save stays enabled, matching the old short-circuit.
- Dropping `muacOutsideConstraints`' "unset ⇒ disabled" is covered — `muacFormInputsAndTasks` returns `[maybeToBoolTask form.muac]`, so an unset MUAC still leaves the task incomplete and the button disabled.
- `anthropometricOutOfRange`/`asEntered` reproduces `muacOutsideConstraints` exactly for the Burundi cm↔mm conversion and is identity for birth weight, so `birthWeightOutsideConstraints` is behaviour-preserving.

One finding belongs to #2007 rather than to this PR: `client/e2e/helpers/well-child.ts:567` reaches the saved activity with `page.locator('.icon-task-history')`, and `Backend/WellChildActivity/Utils.elm:98,110` gives that icon to **both** `WellChildPregnancySummary` and `WellChildNCDA` (confirmed). It is unambiguous today only because PregnancySummary is the first activity completed in both callers, so it is alone on the Completed tab. Reorder so NCDA completes first and Playwright strict mode fails with "resolved to 2 elements".

Also noted: `elm-review` cannot run in this checkout at all — a pre-existing collision between two modules named `Config`, unrelated to this work.

One more that GitHub will not let me anchor, because the line itself is unchanged — **`client/src/elm/Pages/WellChild/Activity/Utils.elm:29`**: four imports are dead now that `nutritionAssessmentSaveDisabled` is gone — `getInputConstraintsHeight`, `getInputConstraintsWeight`, `outsideConstraints`, `muacOutsideConstraints`. Confirmed: each appears exactly once in the file, on that import line. The file is not in `NoUnused.Variables`' ignore list in `review/src/ReviewConfig.elm`, so `elm-review` would flag them once the `Config`-module collision is fixed.

client/src/elm/Translate.elm
```diff
@@ -12074,6 +12074,27 @@ translationSet trans =
...
 
+                Measurement.Model.MeasurementHeight ->
+                    { english = "Height is recorded in centimetres."
+                    , kinyarwanda = Nothing
+                    , kirundi = Nothing
+                    , somali = Nothing
+                    }
+
+                Measurement.Model.MeasurementMuac ->
+                    { english = "MUAC is outside the range it can take."
```
anvmn: **low.** "MUAC is outside the range it can take." is circular — it restates the popup's premise, and it is the only one of the five that does not name a unit.

MUAC is precisely the measurement whose entry unit varies by site: millimetres at Burundi, centimetres elsewhere. The mistake being caught is a mm value typed into a cm field. A nurse at Rwanda who types `125` reads "MUAC is outside the range it can take. Allowed values are between 5 and 99." with nothing to say centimetres are wanted, while the height, weight and birth-weight branches all name their unit.

I wrote it this way deliberately, because a single sentence cannot name a unit that differs by site — but that reasoning produced a sentence that says nothing.

client/src/elm/Pages/WellChild/Activity/Update.elm
```diff
@@ -1940,3 +1953,42 @@ update currentDate site id db msg model =
...
+    -> WellChildEncounterId
+    -> ModelIndexedDb
+    -> Model
+    -> NutritionAssessmentTask
+    -> Msg
+    -> ( Model, Cmd Msg, List App.Model.Msg )
+preSaveNutritionAssessment currentDate site id db model task saveMsg =
+    let
+        outOfRange =
+            nutritionAssessmentOutOfRange site
```
anvmn: **low, and the one I would fix.** The gate reads the raw `model.nutritionAssessmentData.{height,muac,weight}Form`, while the view renders and counts tasks from the `*FormWithDefault` versions (`Pages/WellChild/Activity/View.elm:816-826`).

That matters for `measurementNotTaken`: the view builds it with `heightFormWithDefault assembled.encounter.skippedForms data.heightForm`, so when "not taken" comes from the encounter's `skippedForms` rather than from a click this session, the raw form's field is `Nothing` — and the short-circuit this function documents is silently not exercised in exactly the case it names.

Benign as traced (the raw value is `Nothing` in all reachable states, and `SaveHeight` already works off the raw form the same way), but the gate disagrees with what is on screen, and stops being benign the moment a defaulted value can differ.

Fixable without new plumbing: `PreSaveHeight` and `PreSaveWeight` already carry `skippedForms` and the saved measurement, and `muacFormWithDefault` needs only the value — so the defaulted forms can be rebuilt here, as `PreSavePregnancySummary` already does.

client/src/elm/Measurement/View.elm
```diff
@@ -3689,6 +3689,15 @@ measurementOutOfRangeClass measurement =
...
             "birth-weight-out-of-range"
 
+        MeasurementHeight ->
+            "height-out-of-range"
+
+        MeasurementMuac ->
+            "muac-out-of-range"
+
+        MeasurementWeight ->
+            "weight-out-of-range"
```
anvmn, line 3699: **low, introduced by this PR.** `"birth-weight-out-of-range"` contains `"weight-out-of-range"` as a substring, and the e2e helper matches these with `new RegExp(cls)` (`client/e2e/helpers/common.ts:424,429`).

So a future `notNamed: ['weight-out-of-range']` assertion against a birth-weight popup would silently fail, and a `popupClass: 'weight-out-of-range'` assertion would silently pass on a popup that only names birth weight. No current call site hits it — the collision arrived with this PR's `weight-out-of-range` class.

Anchoring the match per class token would close it.

client/src/elm/Translate.elm
anvmn: Fixed in d224fd144 — the warning now names the unit: "MUAC is recorded in millimetres." at Burundi, "in centimetres." elsewhere, so it reads like the height and weight branches beside it.

I had assumed one sentence could not name a site-dependent unit without new plumbing. It needed none: `Translate.elm` already imports `Site(..)`, and `VaccineType Site VaccineType` / `WellChildImmunisationDescription Site WellChildVaccineType` are the existing precedent for a translation that takes it. So `MeasurementOutOfRangeWarning` takes the site too, and the popup — which already had it — passes it through.

client/src/elm/Pages/WellChild/Activity/Update.elm
anvmn: Fixed in d224fd144. Each `PreSave*` now builds its own defaulted form and asks about that:

```elm
getMeasurementValueFunc saved
    |> heightFormWithDefault skippedForms model.nutritionAssessmentData.heightForm
    |> heightOutOfRange site
```

That forced a simplification worth having: `nutritionAssessmentOutOfRange` dispatched on the task over all three raw forms, but a `PreSave*` only carries its own saved measurement, so it can only default its own form. It is now `heightOutOfRange` / `muacOutOfRange` / `weightOutOfRange`, and the empty `TaskHeadCircumference` and `TaskNutrition` cases are gone with the dispatch.

Not covered by a test: the case that made this worth fixing needs a saved `WellChildHeight` fixture — a whole measurement record — and the suite has no fixture helper for one. The existing `update` tests pass `saved = Nothing`, where defaulting is identity, so they show no regression rather than the new behaviour.

client/src/elm/Measurement/View.elm
anvmn: Fixed in d224fd144, on the matching rather than the class names. `client/e2e/helpers/common.ts` now builds `new RegExp("(^| )" + cls + "( |$)")`, so a class is matched whole and `birth-weight-out-of-range` can no longer satisfy an assertion about `weight-out-of-range`.

Both e2e tests that use the helper pass — the CHW newborn exam, which asserts the birth-weight and birth-length classes, and the nurse encounter, which asserts height, MUAC and weight while checking the other two are absent.

PR #2014, opened 2026-07-27
Measurements: make sure group session height, weight and MUAC are entered in the right units

### LINKED ISSUE #2002: Measurements: make sure group session height, weight and MUAC are entered in the right units

## What the nurse gets

Height, weight and MUAC in a group session explain the allowed range instead of quietly deadening Save. **No change to what can be saved** — only the feedback.

## Where

`Measurement/View.elm:378-388` (`viewFloatForm`'s `saveMsg`). `ModelChild` gains a popup field, `MsgChild` a `PreSave…`/`SetOutOfRangePopup` pair, `FloatFormConfig` the measurement tag.

## Two traps

1. **Units.** The value reaching the gate here is already in **display units** (mm at Burundi, matching `config.constraints = getInputConstraintsMuac site` at `Measurement/View.elm:197`). Use the raw predicate, **not** the cm-converting MUAC one — backwards, every Burundi MUAC pops up.
2. **No task count.** Unlike every other form, this one has no task list, so `saveMsg = Nothing` when the string does not parse to a float is load-bearing and must survive. Only the `withinConstraints` branch becomes a popup.

## Tests

- unit tests on `updateChild` for **both sites**, using `updateChildSetMuacTest` (`Measurement/Test.elm:190`) as the template — `updateChild` is already directly unit-tested, so this ends up among the better-covered sites
- **e2e** in `group-session-nurse.spec.ts`

Related: #1980

anvmn:
Fixes #2002

Based on #2013 — **merge that first**; the base moves down the stack as each one lands. The diff
here is only this issue's work.

Part of #2006.

## What a nurse gets

A group session form refused an implausible height, weight or MUAC by leaving its Save button
with nothing to do, and said nothing about why. The measurement is named on a warning instead.

**What can be saved does not change.**

## The two traps in the issue, and what was done about them

**Units — this form is the opposite of the others.** It holds what the nurse typed, in the unit
it shows: millimetres for MUAC at Burundi. The range above the input is `getInputConstraintsMuac
site`, in that same unit, and the value is compared before being put into the unit it is stored
in. So the cm-converting `muacOutOfRange` used by Well Child and Nutrition would be exactly wrong
here — it would report every MUAC at Burundi.

This adds `outOfRangeAsEntered constraints measurement value`, which compares as-is, with a
docblock saying why it exists beside its converting sibling. Six tests pin it in both directions
at both sites.

**No task count.** `saveMsg = Nothing` when the string does not parse is untouched. On every
other form the task count says a measurement is still to be entered; this form has no task list,
so that `Nothing` is the only thing saying it. Only the `withinConstraints` branch became a
warning.

## Shape

`FloatFormConfig` gains `measurement`, so the warning can name what the form asked for.
`ModelChild` gains `measurementOutOfRangePopupState`, `MsgChild` its setter, and `viewFloatForm`
now takes the site, because the warning names the unit and the unit depends on it.

`withinConstraints` has no callers outside `Measurement.Utils` any more, so it is no longer
exposed — asking about a range now goes through one of the named questions.

## Tests

Nine new: six on `outOfRangeAsEntered` covering millimetres and centimetres at both sites and
both ends of a range, three on `updateChild` for naming and closing the warning.

**e2e** in `group-session-nurse.spec.ts`: height `1050`, weight `850` and MUAC `125` each
refused with a warning naming that measurement, then the real value recorded.

548 modules compile, `elm-format --validate` clean, **2927 tests pass**, elm-review clean in a
fresh clone under CI's checkout, group session e2e passes in 1.3m.

**Discrimination:** adding a conversion to `outOfRangeAsEntered` fails all six of its tests;
**taking the gate out of the form makes the e2e fail**, waiting for a warning that never comes —
which is the only coverage of that wiring, since the decision lives inside `viewFloatForm` and no
unit test reaches it.


anvmn (commented):
Superseded — the same three findings are now pinned to the lines they concern, in the review below.

What the review checked and cleared, which is worth keeping: comparing `inputAsFloat` as typed against `config.constraints` is correct here, because the group session `ModelChild` holds MUAC in millimetres at Burundi and the range shown above the input uses the same `getInputConstraintsMuac site` — check, range helper and the popup range text all agree, with no double conversion. The unparseable-input path still yields a disabled button, unchanged.

client/e2e/helpers/group-session.ts
```diff
@@ -547,34 +547,70 @@ async function saveActivity(page: Page) {
...
-export async function completeMuac(page: Page, value: string) {
+export async function completeMuac(page: Page, value: string, checkRange = false) {
   await openActivity(page, 'MUAC');
   // MUAC uses a number input (spinbutton) — find it near the Save button.
-  const muacInput = page.locator('input[type="number"]').first();
-  await muacInput.fill(value);
+  const input = 'input[type="number"]';
+  if (checkRange) {
+    // Millimetres typed where the site asks for centimetres.
+    await refusesOutOfRange(page, input, '125', 'muac-out-of-range');
```
anvmn, line 612: **medium.** The bare `input[type="number"]` goes into `refusesOutOfRange`, which does `page.locator(input).fill(value)` with no `.first()` — while line 613 below narrows the same selector with `.first()`. If this page ever renders a second number input, the `checkRange` path throws a strict-mode violation while the ordinary path keeps working, so it breaks only in the spec that opted in.

The Elm form sets `name config.blockName` (`Measurement/View.elm:274`), so `input[type="number"][name="muac"]` is available and matches how `completeHeight` and `completeWeight` already select.

anvmn, line 612: **low, same line.** `125` is out of range only away from Burundi, where MUAC runs 50–999. Under `EHEZA_SITE=burundi` it is a valid value: no warning appears and the test dies on the 10s `popup.waitFor` with a timeout that says nothing about why.

The height (`1050`) and weight (`850`) probes are site-independent — this is the one assertion in the new helper that quietly assumes a site. `1500` is outside both ranges.

client/src/elm/Measurement/View.elm
```diff
@@ -429,6 +440,16 @@ viewFloatForm config language currentDate isChw child measurements previousValue
             ]
         , div [ class "actions" ] <|
             saveButton language saveMsg measurements
+        , viewModal <|
```
anvmn, line 443: **low, but the one worth thinking about.** `measurementOutOfRangePopupState` lives on `ModelChild`, which `Pages/Session/Model.elm` keeps in `childForms` for the life of the session page. Confirmed: `getChildForm` only rebuilds from measurement data when the key is absent, and neither `SetSelectedActivity` nor `SetSelectedTab` clears it — the popup's own Close button is the only thing that empties the list.

The `.overlay` blocks in-app clicks, so reaching it needs a route change that goes around it — the back gesture, ordinary on these tablets. Then: Height → type `1050` → Save → warning → back → open that child's **MUAC** form, and the warning reappears unprompted naming the height, because all three float forms share the one field.

Clearing it when the selected activity or participant changes would close that off.

anvmn (commented):
Reposting the same three findings pinned to the lines they concern — the earlier one put them all in the body, which was my mistake and makes them harder to act on.

client/e2e/helpers/group-session.ts
anvmn: Fixed in e70956936 — the helper and the entry that follows it now use the same `input[type="number"][name="muac"]`, so the `.first()` hedge is gone rather than duplicated.

You were right that the name was already there: the Elm form sets `name config.blockName` at `Measurement/View.elm:274`. The comment on the old line ("find it near the Save button") suggests nobody had looked.

anvmn: Not taking this one, deliberately.

`125` is the mistake this range exists to catch — millimetres typed where centimetres are asked for — and the warning it triggers reads "MUAC is recorded in centimetres." `1500` is a value nobody would type, so the test would keep passing while no longer showing what it is there to show.

The failure you describe is also not reachable: the suite hard-codes Rwanda fixtures in **82 places** (`Akanduga`, `Nyange Health Center`), so a Burundi run fails at participant selection long before MUAC. What was worth taking from this is that the assumption was silent, so it is now written down at the call site:

```ts
// Millimetres typed where centimetres are asked for, which is the mistake
// the range is here to catch. Out of range at the sites asking for
// centimetres, which is every site this suite runs against.
```

client/src/elm/Measurement/View.elm
anvmn: Fixed in e70956936. `SetSelectedActivity` (`Pages/Participant/Update.elm:53`) already had `childForm` in hand and passed it straight through, so clearing it there needed no new message or plumbing:

```elm
{ childForm | measurementOutOfRangePopupState = [] }
```

This was the only one of the three a nurse could see, and your trace was right about why: the state is per-child, not per-activity, so the warning outlived the form that raised it and could name a measurement the next form does not ask for.

Group session e2e passes (1.3m), 2927 tests, 548 modules.

anvmn:
[video.webm](https://github.com/user-attachments/assets/c0fe6182-e2ff-49bc-9fcf-8cdd876c57a5)

client/src/elm/Measurement/View.elm
```diff
@@ -378,11 +385,15 @@ viewFloatForm config language currentDate isChw child measurements previousValue
...
-                        config.saveMsg (Maybe.map Tuple.first measurements.current) forBackend |> Just
+                    case outOfRangeAsEntered config.constraints config.measurement asFloat of
+                        [] ->
+                            config.saveMsg (Maybe.map Tuple.first measurements.current) forBackend |> Just
+
+                        outOfRange ->
+                            -- The button answers, and says what is wrong. What
+                            -- does not parse leaves it with nothing to do, as
+                            -- before: this form has no task count to say so.
+                            Just <| SetMeasurementOutOfRangePopupState outOfRange
```
anvmn, line 396:
```suggestion
                            Just SetMeasurementOutOfRangePopupState outOfRange
```

client/src/elm/Measurement/View.elm
```diff
@@ -378,11 +385,15 @@ viewFloatForm config language currentDate isChw child measurements previousValue
...
-                        config.saveMsg (Maybe.map Tuple.first measurements.current) forBackend |> Just
+                    case outOfRangeAsEntered config.constraints config.measurement asFloat of
+                        [] ->
+                            config.saveMsg (Maybe.map Tuple.first measurements.current) forBackend |> Just
+
+                        outOfRange ->
+                            -- The button answers, and says what is wrong. What
+                            -- does not parse leaves it with nothing to do, as
+                            -- before: this form has no task count to say so.
+                            Just SetMeasurementOutOfRangePopupState outOfRange
```
anvmn:
```suggestion
                            Just <| SetMeasurementOutOfRangePopupState outOfRange
```

PR #2015, opened 2026-07-28
Measurements: make sure Acute Illness and Family Nutrition MUAC is entered in the right unit

### LINKED ISSUE #2001: Measurements: make sure Acute Illness and Family Nutrition MUAC is entered in the right unit

## What the nurse gets

MUAC in the Acute Illness physical exam and in the Family Nutrition encounter explains the allowed range instead of quietly deadening Save.

Family Nutrition is the worst of the two today: it passes `showRangeHelper = False` (`Pages/FamilyNutrition/Encounter/View.elm:549`), so it enforces a range it **never states anywhere**. The popup fixes that.

## Where

- `Pages/AcuteIllness/Activity/Utils.elm:384` — delete `physicalExamSaveDisabled` (`:373-388`)
- `Pages/FamilyNutrition/Encounter/View.elm:556`

## Tests

- unit tests on both pages' `update`
- **e2e** in `acute-illness-encounter-nurse.spec.ts` and `family-nutrition-encounter-chw.spec.ts`

Related: #1980

anvmn:
Fixes #2001

Based on #2014 — **merge that first**; the base moves down the stack as each one lands. The diff
here is only this issue's work.

Part of #2006.

## What a nurse gets

Both encounters refused a MUAC outside the range it can take by deadening the Save button, saying
nothing. The measurement is named on a warning instead, and the button goes back to answering the
task count alone. **What can be saved does not change.**

**Family Nutrition is the worse of the two, and the reason this issue is worth doing.** It draws
its MUAC input with `showRangeHelper = False`, so the range was enforced and *never stated
anywhere* — the button simply did not respond. The warning is now the only place a nurse is ever
told what the range is, which is noted at the model field and in the update's docblock so the next
person does not remove it as duplication.

Acute Illness follows the shape of the rest of the series: `physicalExamSaveDisabled` deleted,
`disabled` back to `tasksCompleted /= totalTasks`, and `PreSaveMuac` reading the merged form.

## Tests

`Pages/FamilyNutrition/Encounter/Test.elm` is new — that page had no test module at all. Six
cases: the mother's MUAC and a child's are asked the same way, both sites, and closing the warning
saves nothing on the way out.

The six `physicalExamSaveDisabled` tests are rewritten against `update`, keeping both sites: what
the form holds is centimetres either way, and only the unit it is shown in differs.

**e2e** in `acute-illness-encounter-nurse.spec.ts` and `family-nutrition-encounter-chw.spec.ts`:
each types `125` — millimetres where centimetres are asked for — and goes on only if the warning
names the MUAC.

One trap worth recording: the Acute Illness check is on the test with a **child**. I first put it
on the one with an adult, where the MUAC task is not drawn at all — the check would have passed by
doing nothing, and the test would still have been green.

548 modules compile, `elm-format --validate` clean, **2931 tests pass**, elm-review clean in a
fresh clone under CI's checkout, both e2e specs pass (4 tests, 2.5m).

**Discrimination:** forcing either gate to save fails exactly its own out-of-range cases — two on
Acute Illness, three on Family Nutrition.

### DIFF
--- a/client/src/elm/Pages/AcuteIllness/Activity/Update.elm
+++ b/client/src/elm/Pages/AcuteIllness/Activity/Update.elm
@@ -388,7 +388,10 @@ update site selectedHealthCenter id db msg model =
                     model.physicalExamData
                         |> (\data -> { data | activeTask = Just task })
             in
-            ( { model | physicalExamData = updatedData }
+            -- Moving to another task forgets what the one before it was
+            -- complaining about: the warning names a measurement, and the
+            -- tasks of this activity share one place to put it.
+            ( { model | physicalExamData = updatedData, measurementOutOfRangePopupState = [] }
             , Cmd.none
             , []
             )
--- a/client/src/elm/Pages/AcuteIllness/Activity/Update.elm
+++ b/client/src/elm/Pages/AcuteIllness/Activity/Update.elm
@@ -30,7 +30,9 @@ import Gizra.Update exposing (sequenceExtra)
 import Maybe.Extra exposing (isJust, isNothing, unwrap)
 import Measurement.Utils
     exposing
-        ( ongoingTreatmentReviewFormWithDefault
+        ( muacFormWithDefault
+        , muacOutOfRange
+        , ongoingTreatmentReviewFormWithDefault
         , toHealthEducationValueWithDefault
         , toMuacValueWithDefault
         , toOngoingTreatmentReviewValueWithDefault
@@ -117,6 +119,31 @@ update site selectedHealthCenter id db msg model =
 
         SetWarningPopupState diagnosis ->
             ( { model | warningPopupState = diagnosis }, Cmd.none, [] )
+
+        SetMeasurementOutOfRangePopupState state ->
+            ( { model | measurementOutOfRangePopupState = state }, Cmd.none, [] )
+
+        PreSaveMuac personId saved nextTask ->
+            let
+                outOfRange =
+                    getMeasurementValueFunc saved
+                        |> muacFormWithDefault model.physicalExamData.muacForm
+                        |> muacOutOfRange site
+
+                extraMsgs =
+                    if List.isEmpty outOfRange then
+                        [ SaveMuac personId saved nextTask ]
+
+                    else
+                        -- Tell the nurse what is wrong and leave the form as it
+                        -- is, so the measurement can be entered again.
+                        [ SetMeasurementOutOfRangePopupState outOfRange ]
+            in
+            ( model
+            , Cmd.none
+            , []
+            )
+                |> sequenceExtra (update site selectedHealthCenter id db) extraMsgs
 
         SetPertinentSymptomsPopupState isOpen ->
             ( { model | showPertinentSymptomsPopup = isOpen }, Cmd.none, [] )
--- a/client/src/elm/Pages/AcuteIllness/Activity/View.elm
+++ b/client/src/elm/Pages/AcuteIllness/Activity/View.elm
@@ -105,13 +105,11 @@ viewHeaderAndContent language currentDate site geoInfo id isChw activity db mode
                 SetWarningPopupState
                 assembled
         , viewModal <|
-            pertinentSymptomsPopup language
-                model.showPertinentSymptomsPopup
-                (SetPertinentSymptomsPopupState False)
-                assembled.measurements
-        , viewModal <|
             if List.isEmpty model.measurementOutOfRangePopupState then
-                Nothing
+                pertinentSymptomsPopup language
+                    model.showPertinentSymptomsPopup
+                    (SetPertinentSymptomsPopupState False)
+                    assembled.measurements
 
             else
                 Just <|
--- a/client/src/elm/Pages/AcuteIllness/Activity/View.elm
+++ b/client/src/elm/Pages/AcuteIllness/Activity/View.elm
@@ -41,10 +41,10 @@ import Measurement.Utils
         , treatmentReviewInputsAndTasks
         , vitalsFormWithDefault
         )
-import Measurement.View exposing (viewSendToHealthCenterForm, viewSendToHospitalForm)
+import Measurement.View exposing (measurementOutOfRangePopup, viewSendToHealthCenterForm, viewSendToHospitalForm)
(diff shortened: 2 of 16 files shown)


client/src/elm/Pages/AcuteIllness/Activity/Update.elm
```diff
@@ -118,6 +120,31 @@ update site selectedHealthCenter id db msg model =
         SetWarningPopupState diagnosis ->
             ( { model | warningPopupState = diagnosis }, Cmd.none, [] )
 
+        SetMeasurementOutOfRangePopupState state ->
```
anvmn, line 123: **low, and the same staleness #2014 was fixed for.** This is the only thing that ever empties the warning — the popup's own Close button. `Pages/Participant/Update.elm:68` got an explicit reset on `SetSelectedActivity` precisely for this, but the four encounter pages did not: Acute Illness here, plus `Pages/Nutrition/Activity/Update.elm:58`, `Pages/WellChild/Activity/Update.elm:109` and `Pages/FamilyNutrition/Encounter/Update.elm:118`. Confirmed by grep — none of the four clears it anywhere else.

Their models persist across navigation (`App/Model.elm` keys them by encounter and activity, `App/Update.elm` restores with `Maybe.withDefault emptyModel`), and the `.overlay` is `position: fixed; z-index: 35000`, so it covers the app's own back arrow but not the hardware one.

Sequence: Physical Exam → type `125` into MUAC → Save → warning appears → back gesture instead of Close → re-open Physical Exam, and the warning is on screen again with nothing submitted. Recoverable by pressing Close, hence low — but it is the same defect the group session already had fixed.

anvmn (commented):
Local `/code-review medium`: three findings, all low, all confirmed. Nothing changed yet.

What it traced and cleared first, which is the part that mattered: every removed `disabled` gate has an equivalent replacement — the CHW "unable to take measurement" case still completes its task count, so `tasksCompleted /= totalTasks` reproduces the old short-circuit exactly; every `SaveMuac`/`SaveMuacChild`/`SaveMuacMother` call site now goes through a `PreSave*` with no raw dispatch left; and the gate reads the same merged form the save persists, so it cannot be bypassed by a dirty-vs-saved mismatch.

Two findings cannot be anchored to a line, so they are here.

**`client/src/elm/Pages/AcuteIllness/Activity/Utils.elm:19,24` — low.** `muacOutsideConstraints` and `MuacForm` are now dead: `physicalExamSaveDisabled`, their only consumer, was deleted in this PR. Each appears exactly once, on its import line. The equivalent cleanup was done in `Pages/WellChild/Activity/Utils.elm`, so this file is the odd one out — on a branch series that has a commit titled "Leave nothing behind that is no longer used".

Worth recording *why* this got through: I ran elm-review on a fresh clone under CI's checkout and it reported no errors. That is because `src/elm/Pages/AcuteIllness/Activity/Utils.elm` is in `NoUnused.Variables`' ignore list at `review/src/ReviewConfig.elm:119`. So nothing automated will ever catch dead code in this particular file — the check I have been relying on all series is blind here.

**`client/src/elm/Translate.elm:12061` — low, but the most nurse-facing of the three.** All five `MeasurementOutOfRangeWarning` sentences have `kinyarwanda`, `kirundi` and `somali` as `Nothing`, so they fall back to English. A nurse at Rwanda who types `125` into MUAC now reads:

> MUAC is recorded in centimetres. Imibare yemewe iri hagati ya 5 na 99.

English sentence, Kinyarwanda range — because `AllowedValuesRangeHelper` *is* translated. Before this series the only signal was the `range-helper` line, which was fully localised. CLAUDE.md's rule to leave non-English as `Nothing` is scoped to strings identical to English, like unit abbreviations; these are full sentences.

The strings were added in #2012 and #2013, so this is not new to this PR — it is the series as a whole, and it is not something I can fix by writing the translations myself.

client/src/elm/Pages/AcuteIllness/Activity/View.elm
```diff
@@ -109,6 +109,16 @@ viewHeaderAndContent language currentDate site geoInfo id isChw activity db mode
                 model.showPertinentSymptomsPopup
                 (SetPertinentSymptomsPopupState False)
                 assembled.measurements
+        , viewModal <|
```
anvmn, line 107: Can't we display the popup using `viewModal` at line 107?

client/src/elm/Pages/FamilyNutrition/Encounter/View.elm
```diff
@@ -77,6 +77,16 @@ viewHeaderAndContent language currentDate site id model data =
         [ header
         , content
         , viewModal dialog
+        , viewModal <|
```
anvmn, line 66: Can't we display the popup using viewModal at line 79?

client/src/elm/Pages/AcuteIllness/Activity/Update.elm
anvmn: Fixed in 96872dfa8, and your trace was right about the mechanism.

Cleared where each page can notice the nurse moving on: `SetActiveNutritionAssessmentTask` on Well Child, `SetActivePhysicalExamTask` on Acute Illness, `SetSelectedActivity` **and** `SetSelectedFamilyMember` on Family Nutrition, and leaving the activity on Nutrition.

One thing that only showed up on writing it: Well Child has no field of its own — the warning is a variant of `WarningPopupType`, shared with macrocephaly and the nutrition assessment. Clearing the field outright would have silently dropped a head circumference popup, so it clears only `PopupMeasurementOutOfRange` and leaves the others standing.

Bounded honestly: this covers moving **between** tasks. Re-entering the **same** task by the back gesture still shows it, because nothing tells an encounter page it was re-entered — there is no route-change hook there like the one `Pages/Participant` has. The case that could name the wrong measurement is gone; the harmless re-show is not.

client/src/elm/Pages/AcuteIllness/Activity/View.elm
anvmn: Yes — fixed in c084621ab. It now chooses inside the `viewModal` at that line, so the out-of-range warning and the pertinent-symptoms popup share the one modal slot.

You are right that it matters beyond tidiness: each `viewModal` draws its own `.overlay` at `z-index: 35000`, so two active at once would stack with the lower unreachable.

The inconsistency was mine within this same series — the Nutrition encounter already chose between its warnings in a single `viewModal`, and I appended a second one here.

client/src/elm/Pages/FamilyNutrition/Encounter/View.elm
anvmn: Yes — fixed in c084621ab. It chooses between `dialog` and the warning inside the one `viewModal`, so the end-encounter dialog and the warning cannot both put up an overlay.

anvmn:
[video.webm](https://github.com/user-attachments/assets/d371f5e7-7769-4990-a822-3f033a6aed27)
[video.webm](https://github.com/user-attachments/assets/5e26af26-ad5f-4f60-9986-b93a376a0994)

Issue #2005, opened 2026-07-27
Child scoreboard: make sure Child Scorecard weight and MUAC are entered in the right units


anvmn:
Part of #2006. Based on #2004 — that fix has to land first, or a hidden stale value blocks the form with nothing on screen to correct.

## What the nurse gets

NCDA weight and MUAC explain the allowed range and refuse to save, as birth weight already does. **This is the widest single gap** — `Measurement/View.elm:3290` (weight) and `:3354` (MUAC) are reached from four call sites: Nutrition, Well Child, Child Scoreboard and group sessions. On Child Scoreboard the NCDA form is the only capture there is.

## Approach

Generalise `birthWeightBlocksNCDAForm` (`Measurement/Utils.elm:343`) to return a `List AnthropometricMeasurement`, taking **resolved visibility booleans** rather than `Person` / `NominalDate` — there are no `Person` fixtures anywhere in the test suite, and a predicate that needs one will not get tested.

Compute the violation **inside `ncdaFormInputsAndTasks`**, where the age condition and the not-taken handling already live, so the check cannot desync from what is actually rendered. Extract the visibility conditions into named Utils functions the view also uses — the precedent is #1997 moving `showNCDAQuestionsByNewbornExam` out of the view for exactly this reason.

**Worth adding:** the popup's close action should also `setStepMsg` to the step holding the first offender. The action button is dual-purpose, so the measurement named in the popup may be on a different step from the button that was pressed.

## Why this is last

#1983 documents **three** failed attempts, each defeated by a different thing: MUAC only renders from 6 months of age so an absent value permanently disabled Save; the step tabs bypassed the gate; and hydration resurrected values that "not taken" had cleared (Issue 7).

## Tests

- unit tests over every combination: **under 6 months**, "not taken" ticked, value hydrated from a saved record, ANC step absent
- **e2e** in `child-scoreboard-encounter-chw.spec.ts` **including a child under 6 months** — the exact regression #1983 documents
- discrimination-test each condition separately

Related: #1980, #1983

anvmn:
Delivered by #2018, which reached `develop` today in the #2007 merge (`ade7d38f9`).

Closing by hand: #2018 merged into `i1998-birth-length-range` rather than the default branch, so its `Fixes #2005` never fired.

Issue #2004, opened 2026-07-27
Child scoreboard: make sure a measurement marked as not taken is not kept on the Child Scorecard


anvmn:
Part of #2006.

## What the nurse gets

Ticking "measurement not taken" for NCDA weight or MUAC actually removes the value, instead of silently keeping the one saved earlier.

## The bug

Ticking the box sets `{ weightNotTaken = Just True, weight = Nothing }` (`Measurement/View.elm:3274`), but hydration ignores it:

```elm
weight = or form.weight value.weight   -- Measurement/Utils.elm:8447  (MUAC :8449)
```

so a previously saved value comes straight back. The input is hidden while the stale value stays live — and is saved again at the end.

## The fix

**Use the existing `weightNotTaken` / `muacNotTaken` fields — do not add dirty flags.** They already record the nurse's intent, they persist on the same `NCDAForm` across every step, and they are present at save time, which is what this multi-step flow needs, where the save happens only after all the steps are filled. A dirty flag would be redundant state saying the same thing.

Because `toNCDAValueWithDefault = ncdaFormWithDefault >> toNCDAValue` (`Measurement/Utils.elm:8457`), **one change at the hydration site fixes both what is rendered on any step and what is saved at the end.**

## Tests

- unit tests on hydration directly: saved value + `notTaken = Just True` must yield `Nothing`; an untouched form must still inherit the saved value
- **e2e** in `child-scoreboard-encounter-chw.spec.ts`: save a weight, reopen, tick "not taken", save, assert the backend value is gone
- discrimination-test the fix

## Risk

Four call sites share this hydration (Nutrition, Well Child, Child Scoreboard, group session).

Related: #1983

anvmn:
Delivered by #2017, which reached `develop` today in the #2007 merge (`ade7d38f9`).

Closing by hand: #2017 merged into `i1998-birth-length-range` rather than the default branch, so its `Fixes #2004` never fired.

PR #2016, opened 2026-07-28
Measurements: make sure Prenatal encounter height, weight, MUAC and fundal height are entered in the right units

### LINKED ISSUE #2003: Measurements: make sure Prenatal encounter height, weight, MUAC and fundal height are entered in the right units

## What the nurse gets

Height, weight, MUAC and fundal height are checked on ANC. **Nothing is checked there today** — `grep outsideConstraints` returns zero hits across `Pages/Prenatal/`, so a mistyped measurement saves silently.

Adds `getInputConstraintsFundalHeight` — **1–60 cm**.

## Where

- height `Pages/Prenatal/Activity/View.elm:3364`, weight `:3399`, MUAC `:3449`, fundal height `:3512`
- add a variant to `WarningPopupType msg` (`Pages/Prenatal/Activity/Types.elm:50`) and a branch in `warningPopup` (`View.elm:203`)

## Three traps

1. **Make the popup action `setStateMsg Nothing`, not the save Msg.** Prenatal's three existing popup gates (`View.elm:1916`, `:2457`, `:4193`) are acknowledge-**and-save**, which is the opposite of what is wanted here.
2. **Carried-over height must be excluded.** The input is hidden when a height was measured earlier (`hideHeightInput = isJust previouslyMeasuredHeight`, `View.elm:3194`) and `SaveNutritionAssessment` then overwrites `form.height` with that value — gating it there is an inescapable dead end. #1986 made the same exclusion.
3. **MUAC is stored in cm** here (`setMuacValueForSite`), so use the cm-converting predicate — the opposite of the group form in Issue 5.

## Cleanup

When this lands, `outsideConstraints` and `muacOutsideConstraints` have no callers left. **Delete them here** — that is what stops the next feature adding another silently-disabled button.

## Tests

- unit tests on `update` for each of the four, plus the height-already-measured path
- **e2e** in `prenatal-encounter-nurse.spec.ts`

Related: #1980, #1982, #1986

anvmn:
Fixes #2003

Based on #2015 — **merge that first**; the base moves down the stack as each one lands.

Part of #2006. This is the last of the encounter pages; #2004 and #2005 (the Child Scorecard) remain.

## What a nurse gets

**Nothing was checked here at all.** A height in millimetres, a weight with the decimal in the
wrong place, a MUAC in the unit the site does not ask for, a fundal height that could not be one —
each saved without a word, and went on into the assessment. Each is named on a warning now, and
nothing is saved until it is entered again.

Fundal height gets a range of its own, **1 to 60 cm**.

## The three traps, and how each is held down

**1. Closing the warning must not save.** This page's three existing warnings are
acknowledge-and-save; this one is the opposite — the measurement is still there to be corrected.
The popup's action is `SetWarningPopupState Nothing`.

**2. A carried-over height cannot be gated.** When a height was measured at an earlier encounter
the input is not drawn and `SaveNutritionAssessment` overwrites the form with it, so refusing it
would be an inescapable dead end. Excluded, and *pinned*: forcing that branch on fails the test
that covers it.

**3. MUAC is stored in centimetres here** — the opposite of the group session forms in #2014 — so
it uses the converting question. A Burundi case pins that direction.

Fundal height is asked about only when the uterus can be felt, for the same reason as the height:
otherwise there is nothing on screen to correct. Also pinned.

## The cleanup this issue exists to make possible

`outsideConstraints` and `muacOutsideConstraints` — the pair that answered "outside the range, or
not entered at all" — have no callers left once this lands, so they are **deleted**. That is what
stops the next form refusing a measurement with a button that will not answer and says nothing.

## Tests

Ten new on `update`: each of the four measurements, all three named together, both exclusions, the
Burundi MUAC direction, and the within-range paths. Twelve go with the deleted predicates.

**e2e** in `prenatal-encounter-nurse.spec.ts`: a MUAC of `125` refused with the warning naming it,
the form still up, then `25` recorded.

548 modules compile, `elm-format --validate` clean, **2929 tests pass**, elm-review clean in a
fresh clone under CI's checkout, prenatal e2e passes (3 tests, 5.6m).

**Discrimination:** removing either exclusion fails exactly the test that covers it.

One note on the e2e, since it failed first time and the failure was mine: this page renders its
Save inline with `classList [ ( "ui fluid primary button", True ), ( "disabled", … ) ]`, so an
enabled button carries no `active` class at all. The assertion now checks for the absence of
`disabled`, as the Acute Illness and group session checks already did.

### DIFF
--- a/client/src/elm/Measurement/View.elm
+++ b/client/src/elm/Measurement/View.elm
@@ -2614,6 +2614,20 @@ viewNCDAContent language currentDate site personId person config helperState sho
                 )
                 currentStep
                 |> Maybe.withDefault ( emptyNode, emptyNode )
+
+        outOfRangeWarning =
+            if showBirthWeightOutOfRangePopup then
+                Just <|
+                    measurementOutOfRangePopup language
+                        site
+                        [ MeasurementBirthWeight ]
+                        (config.setBirthWeightOutOfRangePopupMsg False)
+
+            else
+                Nothing
+
+        helperDialog =
+            viewNCDAHelperDialog language (config.setHelperStateMsg Nothing) helperState
     in
     [ header
     , viewTasksCount language tasksCompleted totalTasks
@@ -2622,12 +2636,10 @@ viewNCDAContent language currentDate site personId person config helperState sho
             [ div [ class "ui form ncda" ] viewForm ]
         , actions
         ]
-    , viewModal <|
-        if showBirthWeightOutOfRangePopup then
-            Just <| measurementOutOfRangePopup language site [ MeasurementBirthWeight ] (config.setBirthWeightOutOfRangePopupMsg False)
 
-        else
-            viewNCDAHelperDialog language (config.setHelperStateMsg Nothing) helperState
+    -- One at a time, and the warning wins: it answers what the nurse just
+    -- pressed, while the dialog only explains a question.
+    , viewModal <| Maybe.Extra.or outOfRangeWarning helperDialog
     ]
 
 
--- a/client/src/elm/Measurement/View.elm
+++ b/client/src/elm/Measurement/View.elm
@@ -3708,6 +3708,9 @@ measurementOutOfRangeClass measurement =
         MeasurementBirthWeight ->
             "birth-weight-out-of-range"
 
+        MeasurementFundalHeight ->
+            "fundal-height-out-of-range"
+
         MeasurementHeight ->
             "height-out-of-range"
 
--- a/client/e2e/helpers/prenatal.ts
+++ b/client/e2e/helpers/prenatal.ts
@@ -1,4 +1,4 @@
-import { Page, Locator } from '@playwright/test';
+import { Page, Locator, expect } from '@playwright/test';
 import { click } from './auth';
 import {
   WAIT,
@@ -428,7 +428,12 @@ export async function completeHistory(
  */
 export async function completeExamination(
   page: Page,
-  options?: { isPostpartum?: boolean; vitals?: { sys?: string; dia?: string } },
+  options?: {
+    isPostpartum?: boolean;
+    vitals?: { sys?: string; dia?: string };
+    // Type a MUAC in the wrong unit first, and check the nurse is told.
+    checkRanges?: boolean;
+  },
 ) {
   await openActivity(page, 'prenatal', 'examination');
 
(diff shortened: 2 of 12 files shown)


anvmn (commented):
Local `/code-review medium`: three findings, all low. Nothing changed yet.

What it chased and cleared first is the part that matters on a PR that removes guards from Save buttons across six flows:

- Dropping the `(measurementNotTaken /= Just True) && …` wrapper is safe — the task list already counts `measurementNotTaken` as the completed task, so a CHW can still save a skipped measurement.
- Losing `outsideConstraints … Nothing == True`, which used to keep Save disabled for an unset value, is covered — every affected form's task is `maybeToBoolTask form.<field>`, so an unset or unparseable value still leaves `tasksCompleted /= totalTasks`.
- Units are right in both directions: `outOfRangeAsEntered` compares the typed value against the shown range on the group session forms, while `outOfRange` converts stored-centimetre MUAC before comparing. Prenatal's `form.muac` goes through `setMuacValueForSite`, so it is stored-cm and correctly converted.
- Every `PreSave*` is a strict if/else — no path both warns and saves — and no save path bypasses a check: `outsideConstraints`, `muacOutsideConstraints`, `physicalExamSaveDisabled`, `nutritionAssessmentSaveDisabled` and `PopupBirthWeightOutOfRange` have no stragglers left anywhere.
- Both Prenatal exclusions match the view exactly, so neither can strand a form with no way out.

All three are about lines this diff does not touch — the modal fold landed in #2015 — so none can be anchored and they are all here.

**`client/src/elm/Measurement/View.elm:3361` — low, and deliberately out of scope.** The NCDA form's own height, weight and MUAC inputs are the one anthropometric entry point still unguarded: no range helper, and only birth weight is checked on save. A nurse at Rwanda typing `125` into the NCDA MUAC field still saves 125 cm, which is exactly the failure this PR prevents everywhere else.

That is **#2005**, the next issue in this series — "Make sure Child Scorecard weight and MUAC are entered in the right units", which its own text calls the widest single gap. It is deliberately not in #2003, and it is deliberately after #2004, because a stale hidden value has to stop blocking the form first.

**`client/src/elm/Pages/AcuteIllness/Activity/View.elm:101` — low, and left alone on instruction.** That page still renders two sibling `viewModal`s, so a diagnosis warning and the measurement warning would stack as two full-screen overlays if both states were ever set. Folding mine into the pertinent-symptoms modal was what was asked for; making it a single three-way choice would change behaviour for popups this series did not introduce, and the standing instruction is that side-by-side modals not introduced by recent work are skipped.

**`client/src/elm/Measurement/View.elm:2630` — low, and worth saying out loud.** Merging the two `viewModal` calls does make the helper dialog conditional on the warning not being up — they were independent before.

That is the intended reading: one modal slot, warning first. Two `viewModal`s each draw their own `.overlay` at `z-index: 35000`, so the alternative is stacking them with the lower one unreachable, which is why this was changed.

It is unreachable in practice for the reason you give — the overlay blocks clicks, so `helperState` cannot be set while the warning is up. But you are right that a bare `else` reads as a fallthrough rather than as precedence, and the next person could miss that. Happy to name it, e.g. by binding the two branches to `warning` and `helper` with a comment on which wins.

client/src/elm/Measurement/View.elm
anvmn, line 1: Two findings from the local review land in this file, on lines outside this diff's single hunk — so they are pinned to the file rather than to a line.

**`:3361` — the NCDA form's own height, weight and MUAC inputs are still unguarded.** No range helper, and only birth weight is checked on save. A nurse at Rwanda typing `125` into the NCDA MUAC field still saves 125 cm — the exact failure this PR prevents on every other form.

That is **#2005**, the next issue in this series, which its own text calls the widest single gap. It is sequenced after #2004 deliberately: a stale hidden value has to stop blocking the form first. Out of scope here, not an oversight.

**`:2630` — folding the two `viewModal` calls made the helper dialog conditional on the warning not being up.** They were independent before. That is the intended reading — one modal slot, warning first — because two `viewModal`s each draw a full-screen `.overlay` at `z-index: 35000` and would stack with the lower one unreachable.

Unreachable in practice, since the overlay blocks clicks and `helperState` cannot be set while the warning is up. But the point stands that a bare `else` reads as a fallthrough rather than as precedence. Naming the two branches would make the ordering explicit.

anvmn: Named in the commit above.

`viewModal <| Maybe.Extra.or outOfRangeWarning helperDialog`, with the two bound in the `let` and a comment saying the warning wins because it answers what the nurse just pressed. `Maybe.Extra.or` is already the idiom in this file — `Measurement/View.elm:2430` uses it fourteen lines away.

One thing worth adding to your reasoning: the two states **can** both be set, so this is not only about reading. `helperState` lives on `NCDAData`, which persists across navigation — open the dialog, leave by the back gesture, come back and press Save with a birth weight out of range, and both are set. The warning winning is the right answer there, but it is no longer only a hypothetical.

elm-review clean on a fresh clone (`NoPrematureLetComputation` did not object to the hoist), 2929 tests, and the Child Scorecard e2e passes — the one I showed reaches this warning, since suppressing it makes that test time out.

PR #2017, opened 2026-07-29
Child scoreboard: make sure a measurement marked as not taken is not kept

### LINKED ISSUE #2004: Child scoreboard: make sure a measurement marked as not taken is not kept on the Child Scorecard

anvmn:
Fixes #2004

Based on `i1998-birth-length-range`, which now carries the whole range series (#2007 and the five
merged into it). Part of #2006. #2005 follows this one, and needs it: a stale hidden value would
otherwise block that form with nothing on screen to correct.

## The bug

Ticking "measurement not taken" sets `{ weightNotTaken = Just True, weight = Nothing }`, but
hydration ignored it:

```elm
weight = or form.weight value.weight
```

so the measurement saved at an earlier encounter came straight back. It was then **live behind a
box saying it had not been taken** — and because the Child Scorecard is filled over several steps
and saved only at the end, it was written again.

This is the one issue in the series that fixes a **data** bug rather than feedback, and four pages
share this hydration: Nutrition, Well Child, Child Scorecard and the group sessions.

## The fix

One named question, asked by each of the measurements:

```elm
measurementUnlessNotTaken : Maybe Bool -> Maybe a -> Maybe a -> Maybe a
```

No dirty flags, as #2004 asks: `weightNotTaken` and `muacNotTaken` already record the intent, they
persist on the same form across every step, and they are present at save time. Because
`toNCDAValueWithDefault = ncdaFormWithDefault >> toNCDAValue`, the one change fixes both what is
rendered on any step and what is saved at the end.

## Beyond what the issue asks

The issue names weight and MUAC. **Stunting level has the identical bug** — same checkbox clearing
the value (`stuntingLevel = Nothing`), same `or form.stuntingLevel value.stuntingLevel` merge. It
is fixed with them; leaving one of three identical cases would have been arbitrary.

## Tests

Six on hydration directly: each of the three does not come back, saying one was not taken leaves
the others alone, an untouched form still inherits what was saved, and what the nurse typed still
wins over what was saved.

**e2e** in `child-scoreboard-encounter-chw.spec.ts`: complete the scorecard with a weight, reopen
it from the Completed tab, check the saved weight is on the form, tick "not taken", check the input
goes, save through the remaining steps, and read the field back from the backend — it must be
**null**. That needed a new query: the existing one only reports whether a node exists, and here
the node always exists; the question is whether the measurement was written into it.

**Discrimination, both directions:**

- putting the old merge back fails four tests
- fixing **only** weight and MUAC, exactly as the issue names, fails the stunting-level test — which
  is what makes the third case evidence rather than my speculation

548 modules compile, `elm-format --validate` clean, **2935 tests pass**.

### DIFF
--- a/client/e2e/child-scoreboard-encounter-chw.spec.ts
+++ b/client/e2e/child-scoreboard-encounter-chw.spec.ts
@@ -6,6 +6,8 @@ import { syncAndWait } from './helpers/common';
 import {
   createChildAndStartEncounter,
   completeNCDA,
+  queryNCDAWeight,
+  reopenNCDAAndSayWeightNotTaken,
   completeVaccinationHistory,
   endChildScoreboardEncounter,
   queryChildScoreboardNodes,
@@ -51,7 +53,12 @@ test.describe('CHW: Child Scoreboard Encounter — First NCDA + Vaccination Hist
     //    ChildBehindOnVaccination answered "No" to trigger VaccinationHistory.
     await completeNCDA(page);
 
-    // 3. Complete VaccinationHistory activity (answer "No" to prior doses for each vaccine).
+    // 3. Reopen the scorecard and say the weight could not be taken. The form
+    //    is saved only at the end of its steps, so this checks the weight saved
+    //    a moment ago does not come back and get written again (#2004).
+    await reopenNCDAAndSayWeightNotTaken(page);
+
+    // 4. Complete VaccinationHistory activity (answer "No" to prior doses for each vaccine).
     await completeVaccinationHistory(page);
 
     // 4. End encounter (no diarrhea popup since we answered No).
@@ -77,6 +84,10 @@ test.describe('CHW: Child Scoreboard Encounter — First NCDA + Vaccination Hist
     const nodes = queryChildScoreboardNodes(fullName, expectedTypes);
 
     expect(nodes['child_scoreboard_ncda'], 'child_scoreboard_ncda should exist').toBe(true);
+
+    // The weight was saved, then said not to have been taken: it should be gone
+    // rather than written again from what the form was holding.
+    expect(queryNCDAWeight(fullName), 'the weight should not be kept').toBeNull();
     expect(nodes['child_scoreboard_bcg_iz'], 'child_scoreboard_bcg_iz should exist').toBe(true);
     expect(nodes['child_scoreboard_opv_iz'], 'child_scoreboard_opv_iz should exist').toBe(true);
     expect(nodes['child_scoreboard_dtp_iz'], 'child_scoreboard_dtp_iz should exist').toBe(true);
--- a/client/e2e/child-scoreboard-encounter-chw.spec.ts
+++ b/client/e2e/child-scoreboard-encounter-chw.spec.ts
@@ -61,13 +61,13 @@ test.describe('CHW: Child Scoreboard Encounter — First NCDA + Vaccination Hist
     // 4. Complete VaccinationHistory activity (answer "No" to prior doses for each vaccine).
     await completeVaccinationHistory(page);
 
-    // 4. End encounter (no diarrhea popup since we answered No).
+    // 5. End encounter (no diarrhea popup since we answered No).
     await endChildScoreboardEncounter(page);
 
-    // 5. Sync to backend.
+    // 6. Sync to backend.
     await syncAndWait(page);
 
-    // 6. Verify backend nodes.
+    // 7. Verify backend nodes.
     // For a 10-month-old male on Rwanda site with no vaccination history,
     // all 7 common vaccines are overdue (BCG, OPV, DTP, PCV13, Rotarix, IPV, MR).
     // DTPStandalone is Burundi-only, so absent on Rwanda.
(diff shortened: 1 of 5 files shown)


client/e2e/helpers/child-scoreboard.ts
anvmn, line 1: Stale docblock. "Complete the Vaccination History activity…" now sits above `reopenNCDAAndSayWeightNotTaken`, because the new function was inserted between it and the function it describes — which is left undocumented below. Same at the `queryNCDAWeight` docblock.

client/e2e/helpers/child-scoreboard.ts
```diff
@@ -453,6 +520,51 @@ export async function endChildScoreboardEncounter(page: Page) {
...
+    }
+    \\$node = node_load(key(\\$r['node']));
+    \\$wrapper = entity_metadata_wrapper('node', \\$node);
+    \\$weight = \\$wrapper->field_weight->value();
+    echo json_encode(['weight' => \\$weight]);
+  `;
+
+  const { drushCmd, cwd } = drushEnv();
+  const output = execSync(`${drushCmd} eval "${php}"`, { cwd, timeout: 30000, encoding: 'utf-8' });
+  const parsed = JSON.parse(output.trim());
```
anvmn, line 580: Unlike `queryMeasurementNodes` (`e2e/helpers/common.ts:155`), this parses drush output with a bare `JSON.parse(output.trim())` — no `stdio: 'pipe'`, no try/catch, no retry. Any stray drush notice on stdout turns a data check into an opaque `SyntaxError` instead of a readable failure. `key($r['node'])` also takes an arbitrary NCDA node if the person ever has more than one.

client/e2e/child-scoreboard-encounter-chw.spec.ts
```diff
@@ -51,7 +53,12 @@ test.describe('CHW: Child Scoreboard Encounter — First NCDA + Vaccination Hist
...
-    // 3. Complete VaccinationHistory activity (answer "No" to prior doses for each vaccine).
+    // 3. Reopen the scorecard and say the weight could not be taken. The form
+    //    is saved only at the end of its steps, so this checks the weight saved
+    //    a moment ago does not come back and get written again (#2004).
+    await reopenNCDAAndSayWeightNotTaken(page);
+
+    // 4. Complete VaccinationHistory activity (answer "No" to prior doses for each vaccine).
     await completeVaccinationHistory(page);
 
     // 4. End encounter (no diarrhea popup since we answered No).
```
anvmn: Step numbering broken by the renumber: a "// 4. Complete VaccinationHistory" was added but "// 4. End encounter" left below it, so steps 4–6 are off by one.

client/src/elm/Measurement/Utils.elm
```diff
@@ -8551,17 +8551,34 @@ ncdaFormWithDefault form saved =
...
 
                 -- Nutrition measurements.
-                , stuntingLevel = or form.stuntingLevel value.stuntingLevel
+                , stuntingLevel = measurementUnlessNotTaken form.stuntingLevelNotTaken form.stuntingLevel value.stuntingLevel
                 , stuntingLevelNotTaken = or form.stuntingLevelNotTaken (isNothing value.stuntingLevel |> Just)
-                , weight = or form.weight value.weight
+                , weight = measurementUnlessNotTaken form.weightNotTaken form.weight value.weight
                 , weightNotTaken = or form.weightNotTaken (isNothing value.weight |> Just)
-                , muac = or form.muac value.muac
+                , muac = measurementUnlessNotTaken form.muacNotTaken form.muac value.muac
```
anvmn, line 8558: Consistency gap worth noting, not a regression. Clearing `muac` also removes the *Targeted Interventions* question it gates (`muacMeasurementIsOff form.muac`, `Measurement/View.elm:3456`). A nurse who records MUAC 12.0, answers "treated for acute malnutrition" = Yes, then goes back and ticks MUAC "not taken" loses the question from the later step while `form.treatedForAcuteMalnutrition` stays `Just True` — and `toNCDAValue` writes the sign unconditionally, so the encounter saves an answer to a question the form no longer asks. The same is already true when filling the form for the first time, so this predates the change.

Issue #1983, opened 2026-07-22
NCDA weight and MUAC need range gates, but the form defeats the usual approach


anvmn:
Split out of #1980. The shared NCDA form was going to be gated alongside the prenatal nutrition assessment and the well child birth length, but three attempts failed for three different reasons and it was dropped from that change. Recording what was learned so the next attempt starts from here.

## Why it matters

`Measurement/View.elm:3272` (weight) and `:3341` (MUAC) have no range check, and the form is reached from **four** call sites — Nutrition, WellChild, ChildScoreboard and the group session. It is the single widest gap in #1980. On ChildScoreboard the nutrition assessment step is the only NCDA step shown.

## What was tried, and how each failed

**1. Gate scoped to the current step, with per-field not-taken guards.**
- The MUAC input is only rendered from six months of age: `ageInMonths currentDate person |> Maybe.map (\ageMonths -> if ageMonths >= 6 then ... else ([], [])) |> Maybe.withDefault ([], [])`. For a younger child, or one with no birth date, neither the input nor its not-taken checkbox exists. `outsideConstraints` counts an absent value as outside, so **Save was permanently disabled with nothing on screen to correct**.
- The step tabs are clickable whenever inactive (`navigationAction`), and the save stores the whole form, so moving to another step **bypassed the gate entirely**.

**2. Check only measurements that are actually set (`isJust value && check value`).**
- Defeated by hydration. `ncdaFormWithDefault` does `weight = or form.weight value.weight` and `muac = or form.muac value.muac`, and `NCDAForm` has **no dirty flag** for either. Ticking "measurement not taken" sets `weight = Nothing` on the raw form, but the view renders the hydrated form, and `or Nothing (Just 850)` brings the saved value straight back. Input hidden, stale out-of-range value live, **Save dead again**.

**3. Either of the above, at the button.**
- The button is dual-purpose: `actionButton actionMsg`, where `actionMsg` is `config.setStepMsg <next incomplete step>` when tasks remain and `config.saveMsg` when they don't. Disabling it can block **navigation between steps**, not just saving.

## Why the other two sites were fine

Both use dirty flags, so an edit wins over the saved value: prenatal via `valueConsideringIsDirtyField form.heightDirty ...`, well child via `maybeValueConsideringIsDirtyField form.birthLengthDirty ...` (and toggling birth length availability off sets `birthLengthDirty = True`). NCDA has no equivalent.

## Suggested direction

[...]

PR #2018, opened 2026-07-29
Child scoreboard: make sure weight and MUAC are entered in the right units

### LINKED ISSUE #2005: Child scoreboard: make sure Child Scorecard weight and MUAC are entered in the right units

anvmn:
Fixes #2005. Part of #2006, based on #2004.

## What the nurse gets

NCDA weight and MUAC now explain the allowed range and refuse to save, as birth weight already does. This was the widest single gap: neither was checked at all, so a MUAC in millimetres typed into a field holding centimetres, or a weight in grams typed into one holding kilograms, was saved as it stood. The same form is reached from Child Scoreboard, Nutrition, Well Child and group sessions.

The warning also opens the form on the step that asks for what it names. The action button is stopped on every step, so the measurement named is often asked on another one — previously the nurse was left to go looking for it.

## Only what the form asks is checked

`birthWeightBlocksNCDAForm` is generalised to `ncdaMeasurementsOutOfRange`, returning each offending measurement together with the step that asks for it. Three separate conditions decide whether a measurement is checked at all:

| Measurement | Asked when |
|---|---|
| Birth weight | the Antenatal Care step is part of the form **and** the newborn exam did not already record it |
| Weight | the Nutrition Assessment step is part of the form (**not** at a health centre) |
| MUAC | that step **and** the child is at least six months old |

A value saved earlier stays on the form after its question is gone, so checking one that is nowhere to be seen would stop the form with nothing on screen to correct — how this was defeated three times before (#1983). The age condition is now a named function, `ncdaMuacAsked`, used both by the check and by the input that draws the MUAC, so the two cannot come to disagree.

`SetBirthWeightOutOfRangePopup` is renamed to `SetMeasurementOutOfRangePopup` across the four pages, since it no longer reports birth weight alone, and its payload carries the step to open.

## Tests

15 unit tests over every combination, and each condition was discrimination-tested on its own — breaking one at a time and confirming the matching test, and only that test, fails:

| Broken | Caught by |
|---|---|
| birth weight ignores the Antenatal Care step | `without that step it is not named` |
| birth weight ignores the newborn exam | `when the newborn exam already recorded one` |
| weight ignores the Nutrition Assessment step | `at a health centre that step is not asked` |
| MUAC ignores the age it is asked from | `under six months it is not asked` |
| MUAC is not converted to the site units | the two Burundi cases |
| an absent measurement counts as out of range | `a form holding no weight is not named` |
| "not taken" no longer empties the measurement | `nothing left to be out of range` |

2946 unit tests pass; `elm make src/elm/Main.elm` compiles 548 modules; `elm-format --validate` clean.

E2E, including a child under six months, is being added next.


client/e2e/child-scoreboard-encounter-chw.spec.ts
```diff
@@ -99,3 +102,46 @@ test.describe('CHW: Child Scoreboard Encounter — First NCDA + Vaccination Hist
...
+  //             MUAC input is part of the form.
+  // The regression #1983 documents: the range check must ask only for what the
+  // form shows. A MUAC that is checked but never drawn leaves the nurse with a
+  // button that will not answer and no input to correct — three earlier attempts
+  // were defeated this way. The weight, which is asked at every age, is still
+  // refused when it is out of range, so this is not passing by checking nothing.
+  test('is not asked for a MUAC, and still saves', async ({ page }) => {
+    const { fullName } = await createChildAndStartEncounter(page, { ageMonths: 3 });
+
+    await expect(
```
anvmn: **Confirmed defect — this assertion proves nothing.**

`createChildAndStartEncounter` returns with the page on `div.page-encounter.child-scoreboard`, the activity *list* (`helpers/child-scoreboard.ts:126-128`). The NCDA form is not open until `completeNCDA` runs on line 135, and even inside it the MUAC input only exists on the Nutrition Assessment step. So this locator has count 0 for a child of **any** age.

Failure scenario: revert `ncdaFormInputsAndTasks` to draw the MUAC at every age — the exact #1983 regression this test is named for — and the assertion still passes.

The test is not wholly empty (the weight range check runs, and `queryNCDAWeight === 8.5` is verified against the backend row), but the assertion carrying its name is vacuous. It needs to move after the step is opened.

client/src/elm/Measurement/Model.elm
```diff
@@ -283,7 +283,7 @@ type MsgChild
     | ToggleANCVisitDate NominalDate
     | SetNCDABoolInput (Bool -> NCDAForm -> NCDAForm) Bool
     | SetBirthWeight String
-    | SetBirthWeightOutOfRangePopup Bool
+    | SetMeasurementOutOfRangePopup (Maybe NCDAStep)
```
anvmn, line 286: Two near-identically named constructors now sit in the same union: `SetMeasurementOutOfRangePopupState (List AnthropometricMeasurement)` at :270 (group height/MUAC/weight forms) and this `SetMeasurementOutOfRangePopup (Maybe NCDAStep)` (NCDA). The old name `SetBirthWeightOutOfRangePopup` was unambiguous.

The payload types differ, so a mix-up would not silently type-check — but the existing test is *named* `updateChild SetMeasurementOutOfRangePopupState`, so a reader looking for coverage of the new branch will land on the wrong one.

client/src/elm/Pages/ChildScoreboard/Activity/Update.elm
```diff
@@ -240,11 +240,25 @@ update currentDate site id db msg model =
...
-        SetBirthWeightOutOfRangePopup isOpen ->
+        SetMeasurementOutOfRangePopup stepAsking ->
             let
+                -- Open on the step that asks for the measurement, so that the
+                -- input the warning names is the one on screen behind it.
+                updatedForm =
+                    Maybe.map (\step -> { form_ | step = Just step }) stepAsking
+                        |> Maybe.withDefault form_
+
+                form_ =
```
anvmn: Two points on this branch, which is duplicated identically in four modules (`Measurement/Update.elm:424`, `Nutrition/Activity/Update.elm:538`, `WellChild/Activity/Update.elm:1696`):

1. **Duplication against the repo convention.** Each copy re-implements the body of `SetNCDAFormStep`, which sits ~15 lines above in the same `case` and does exactly `{ form | step = Just step }` with no extra `Cmd`. Eight copies of the same record update in total.
2. **`form_` is used before it is defined.** Legal — Elm let bindings are unordered, and the app compiles — but it inverts this file's own idiom, where `form_` is always bound first (e.g. `Nutrition/Activity/Update.elm:121`).

No test covers this branch either: nothing asserts that `Just step` sets both `showMeasurementOutOfRangePopup` **and** `form.step`, or that `Nothing` closes without moving. This is the part duplicated four times and therefore the most likely to drift — dropping `form = updatedForm` from one copy would leave that page naming a measurement on a step the nurse is not on, with no test firing.

client/e2e/helpers/child-scoreboard.ts
```diff
@@ -67,13 +67,20 @@ async function answerNCDAYesNo(page: Page, questionSubstring: string, answer: 'Y
 /**
  * Click an NCDA step tab icon and wait for it to become active.
  */
-async function clickNCDAStepTab(page: Page, iconClass: string) {
+async function clickNCDAStepTab(page: Page, iconClass: string): Promise<boolean> {
   const tab = page.locator(`.link-section:has(.icon-activity-task.icon-${iconClass})`);
+  if (!(await tab.isVisible({ timeout: 1000 }).catch(() => false))) {
```
anvmn, line 76: `clickNCDAStepTab` now fails soft for **every** caller, but only the Nutrition Behavior one checks the result. The other four (`ncda-antenatal`, `nutrition-assessment`, `ncda-targeted-intervention`, `ncda-infrastructure-environment`) ignore the boolean.

Failure scenario: a regression drops one of those steps. Previously the tab click failed loudly; now it returns `false`, and the following `answerNCDAYesNo` calls answer questions on whatever step happens to be showing — or fail later with an unrelated message. Only Nutrition Behavior is genuinely optional; the rest could stay strict.

client/src/elm/Measurement/View.elm
```diff
@@ -2541,11 +2544,17 @@ viewNCDAContent language currentDate site personId person config helperState sho
...
                             Pages.Utils.saveButton language
                                 (tasksCompleted == totalTasks)
-                                (if birthWeightOutOfRange then
-                                    config.setBirthWeightOutOfRangePopupMsg True
-
-                                 else
-                                    msg
+                                (case List.head measurementsOutOfRange of
+                                    Just ( _, stepAsking ) ->
+                                        -- Open on the step that asks for it: the
```
anvmn, line 2549: The popup names **every** out-of-range measurement but moves the form to `List.head`'s step, with nothing telling the nurse the form moved.

Scenario: birth weight `3` and weight `8500` are both wrong. The popup lists both, the form silently jumps to Antenatal Care; after fixing the birth weight she presses Save and is moved again to Nutrition Assessment. Correct, but two unexplained page moves.

anvmn (commented):
Findings from two review passes: a local pass scoped to this PR's diff (`i2004..i2005`), and Copilot. A third pass (`/code-review medium`) reviewed the whole `develop..HEAD` stack and raised two items that live in already-merged code, not in this diff — they are described in the conversation rather than here.

One finding is a confirmed defect I introduced (the vacuous assertion below); the rest are hygiene. Nothing has been changed yet.

PR #2020, opened 2026-07-29
Measurements: tidy the out-of-range warning: one range, and no stale warning on return

### LINKED ISSUE #2019: Measurements: tidy the out-of-range warning: one range, and no stale warning on return
Follow-up to #2006, from the reviews of #2017 and #2018. Both items are small and neither is urgent.

## 1. The allowed range was written three times

`FloatFormConfig` carried its own `constraints`, so the group-session forms stated one range (`AllowedValuesRangeHelper`), enforced a second (`outOfRangeAsEntered config.constraints`) and quoted a third in the warning (`anthropometricConstraints site measurement`). All three were the same range written three times, and editing `getInputConstraintsHeight` alone would have left the nurse reading a range the button was not enforcing.

## 2. A warning was waiting on the way back

`Pages/AcuteIllness/Activity/Update.elm` cleared `measurementOutOfRangePopupState` when moving between physical-exam tasks, but not on `SetActivePage`. The page is kept per encounter (`App/Model.elm` keys `acuteIllnessActivityPages` by encounter and activity), so a nurse who used Back instead of Close met the warning again on returning — over a measurement that may since have been corrected. `Pages/Nutrition/Activity/Update.elm` already cleared on leaving.

## Still open, deliberately not included

A hidden NCDA question keeps its answer: ticking MUAC "not taken" removes the *Targeted Interventions* question that `muacMeasurementIsOff` gates, while `toNCDAValue` still writes `TreatedForAcuteMalnutrition` from the form. The encounter then saves an answer to a question the form no longer asks. This predates all of #2006 and needs a decision about what a hidden NCDA question should do with its answer, so it wants its own issue rather than a quiet fix.

Related: #1980, #2006.

anvmn:
Fixes #2019 in part. Follow-up to #2006, from the reviews of #2017 and #2018.

Based on `i1998-birth-length-range` (#2007) rather than `develop`, because the code it touches only exists there.

## The allowed range is said once

`FloatFormConfig.constraints` is gone. The range now follows from `config.measurement` via `anthropometricConstraints site`, taken once in `viewFloatForm` where the site is known — so the range the form **states** (`AllowedValuesRangeHelper`), the range it **enforces** (`outOfRangeAsEntered`), and the range the warning **quotes** are one value.

Before, they were three independent definitions that happened to agree. No live bug; editing `getInputConstraintsHeight` alone would have made one — the nurse would have read a range the button was not enforcing.

One file, 14 insertions.

## What was dropped from this PR

The second item in #2019 — a warning left open being waiting on the way back — is **not** fixed here, and #2019 stays open for it.

The per-page clear I first wrote turned out to be unreachable: the warning is drawn over an overlay that blocks the page behind it, so leaving is only possible once the warning has been closed, which already forgets it. The only thing pressable with a warning up is the browser's own Back, which reaches no page message at all.

Fixing it properly means forgetting on arrival, in the one place every navigation passes through — which reaches `App/Update.elm` and four page models. That is more of the app than a range consolidation should carry, so it was reverted and left for a change of its own.


client/src/elm/Measurement/View.elm
```diff
@@ -272,8 +269,8 @@ viewFloatForm site config language currentDate isChw child measurements previous
             [ type_ "number"
             , placeholder <| translate language config.placeholderText
             , name config.blockName
-            , Attr.min <| String.fromFloat config.constraints.minVal
-            , Attr.max <| String.fromFloat config.constraints.maxVal
+            , Attr.min <| String.fromFloat (anthropometricConstraints site config.measurement).minVal
```
anvmn: The funciton `anthropometricConstraints` activated twice for no reason.
`(anthropometricConstraints site config.measurement).minVal` is bad code style.

client/src/elm/Pages/AcuteIllness/Activity/Update.elm
```diff
@@ -109,7 +109,11 @@ update site selectedHealthCenter id db msg model =
             ( model, Cmd.none, [] )
 
         SetActivePage page ->
```
anvmn, line 111: **This clear cannot fire in the scenario the commit message describes — verified.**

The warning is rendered through `viewModal`, whose `.overlay` is `position: fixed`, full-viewport, `z-index: 35000`, and the SCSS comment says outright that it blocks clicks to what is behind it. So the page's own back link cannot be reached while the warning is up: `SetActivePage` is only reachable *after* the popup has been closed by `SetMeasurementOutOfRangePopupState []`, which already clears the state. In that path this line is a no-op.

The path that **can** fire with the warning open is the browser or Android hardware Back button. That arrives as `UrlChanged` (`App/Update.elm:1494`), which recomputes `activePage` and issues a few `bindDropZone`/GPS commands — it dispatches no page message at all. `acuteIllnessActivityPages` is keyed by `(encounterId, activity)` and never removed from, so the stale warning survives and is waiting on return.

So the fix is correct in itself but does not reach the case it was written for. Clearing on page entry (`App/Update.elm:723`, where `Maybe.withDefault emptyModel` already runs) or on `UrlChanged` would close it — and would cover all four pages at once rather than one at a time.

anvmn (commented):
Findings from the `/code-review medium` pass on this branch. The refactor in `Measurement/View.elm` was checked and found behaviour-preserving; the findings concern the *reach* of the second fix.

Three further findings are the same class in files **not in this diff**, so they cannot be anchored here: `Pages/WellChild/Activity/Update.elm:103` (`SetActivePage` does not drop `Just (PopupMeasurementOutOfRange _)`, while the sibling `SetActiveNutritionAssessmentTask:399` does exactly that filtering), `Pages/Prenatal/Activity/Update.elm:194` (same for `WarningPopupMeasurementOutOfRange`), and `Pages/FamilyNutrition/Encounter/Update.elm:231` (clears `dialogState` but not `measurementOutOfRangePopupState`, though `SetSelectedActivity:324` and `SetSelectedFamilyMember:343` both do). All four page models are kept per `(encounterId, activity)` and never removed.

Issue #2009, opened 2026-07-27
Measurements: A birth length can be stored without the sign that shows its input, and is then re-saved unchecked


anvmn:
Found while reviewing #1998. **Pre-existing** — the range check added there did not create it, but it is a hole in that check.

## What happens

`toPregnancySummaryValue` (`Pages/WellChild/Activity/Utils.elm:305`) writes the birth length **unconditionally**:

```elm
|> andMap (Just form.birthLength)
```

regardless of `form.birthLengthAvailable`, which is what decides whether the input is shown at all (`Pages/WellChild/Activity/View.elm:347`) and what sets the `BirthLength` sign.

The decoder makes the mismatch reachable. `decodePregnancySummaryValue` (`Backend/Measurement/Decoder.elm:4675`) decodes `height` independently of `pregnancy_summary_signs`, and the signs fall back to `NoPregnancySummarySigns` when missing or unparsable. So a record holding a height with no `BirthLength` sign loads as:

- `birthLengthAvailable = Just False` → **the input is not rendered**
- `birthLength = Just (HeightInCm 0.5)` → **the value is still live**

The nurse sees no length on the form, and every subsequent save writes `0.5` back out.

Records like this do occur: the same decoder fallback is the reason `showNCDAQuestionsByNewbornExam` (`Measurement/Utils.elm`) exists.

## Why it matters now

#1998 gates the birth length on `birthLengthAvailable == Just True`, deliberately — the input is hidden otherwise, and stopping on a length that is nowhere to be seen would leave the form impossible to save. That is right for the nurse, but it means a length in this state **skips the range check and is re-saved unchecked**, which is the very thing #1998 set out to prevent.

## Fix — needs a decision, not just code

The obvious change is to stop writing a length the form says is not available. That is consistent, and it removes the hidden-value state altogether. But it also means a re-save would **clear** a stored length whenever the sign is missing, which is data loss rather than a display problem.

So the options are, roughly:

1. Don't write the length unless `birthLengthAvailable == Just True` — consistent, but can clear stored values.
2. Derive `birthLengthAvailable` from the value when the sign is missing, so the input is shown and the nurse can correct or confirm it — no data loss, and the range check then applies.
3. Leave the write as it is and check the length whenever it is set **and** would be written, accepting that the nurse may be stopped over a field she cannot see — the trap #1998 avoided.

[...]

PR #2021, opened 2026-07-30
Measurements: make sure the Apgar scores are entered as scores out of ten

### LINKED ISSUE #2022: Measurements: make sure the Apgar scores are entered as scores out of ten
Part of #2006. Noticed while investigating #2009 (*Also worth checking*), but a separate defect: #2009 is about a value stored without the sign that shows its input, whereas this is about no range check existing at all.

## What the data says

Read-only query against `ihangane.live`, all 596 `well_child_pregnancy_summary` records:

| field | min | max | recorded | out of range |
|---|---|---|---|---|
| birth length | 4.5 | 65 | 32 | 3 (9%) — range 15–60 |
| Apgar 1-minute | 0 | **3000** | 129 | **25 (19%)** |
| Apgar 5-minute | 0 | **30000** | 129 | **32 (25%)** |

The Apgar score is a score out of 10. A quarter of every five-minute score on record cannot be one.

The out-of-range values cluster at 30–40 — `36` appears six times, then `30`, `35`, `15` — with a tail through `246`, `3000`, `30000`. The clustering suggests another quantity being typed into the box; whatever it is, it is not an Apgar score.

## Why it was missed

#2006 scoped itself to anthropometric measurements — height, weight, MUAC, birth length, fundal height. The Apgar scores are not measurements of the body, so they were never in scope. They turn out to be the worst-affected field on the newborn exam: the one field the mechanism never covered.

## What is wanted

Both scores refused when what is entered cannot be a score out of 10, named on the warning the newborn exam already shows for a birth weight or birth length.

The scores sit behind their own `apgarScoresAvailable` question and their inputs are not drawn when the answer is no, so the check must be gated on that question — stopping on a field that is nowhere to be seen would leave the form impossible to save, the regression #1983 documents.

## Not included

The 57 records already holding impossible scores. A nurse only meets the check on re-save, so they stay as they are until touched. Worth deciding separately whether they need a cleanup.

Related: #1998, #2006, #2009

anvmn:
Fixes #2022. Part of #2006. Based on `i1998-birth-length-range` (#2007), where the newborn exam gate lives.

Noticed while investigating #2009, but a separate defect — #2009 stays open for the hidden-value shape it was written about (see the end of this description).

## Why this first

Production data (read-only query on `ihangane.live`, 596 pregnancy summaries):

| field | min | max | recorded | out of range |
|---|---|---|---|---|
| birth length | 4.5 | 65 | 32 | 3 (9%) |
| Apgar 1-minute | 0 | **3000** | 129 | **25 (19%)** |
| Apgar 5-minute | 0 | **30000** | 129 | **32 (25%)** |

The Apgar score is a score out of 10. A quarter of every five-minute score on record cannot be one. The out-of-range values cluster at 30–40 — 36 appears six times, then 30, 35, 15 — with a tail through 246, 3000, 30000. Whatever is being typed, it is not an Apgar score.

Nothing checked them: #2006 scoped itself to anthropometrics, so the worst-affected field on the form was the one field the mechanism never covered.

## The change

`AnthropometricMeasurement` becomes **`RangedMeasurement`**. Its own docblock said "a measurement of the body", which an Apgar score is not — the honest common property is "a number entered on a form, with a range it has to be within". `measurementConstraints` and `measurementOutOfRange` are renamed to match. Mechanical, 21 files.

Two variants are added, `MeasurementApgarOneMinute` and `MeasurementApgarFiveMinutes`, with a 0–10 range, and both are named on the warning the newborn exam already shows for a birth weight or birth length.

`birthLengthOutsideConstraints` is replaced by **`pregnancySummaryMeasurementsOutOfRange`**, which asks the whole form at once and returns everything wrong with it. The nurse is told about all of them together rather than being sent back once per measurement.

Each score is only checked when `apgarScoresAvailable == Just True`, exactly as the length is gated on `birthLengthAvailable` — the inputs are not drawn otherwise, and stopping on a field that is nowhere to be seen would leave the form impossible to save (#1983, #1998).

## Tests

20 cases replacing the 7 that were there, each existing birth-length case kept. Two conditions discrimination-tested separately:

| Broken | Caught by |
|---|---|
| the question no longer hides its scores | `scores are not reported when the form does not ask for them` (and the birth length equivalent) |
| the 0–10 range widened | the three "refused" cases, and only those |

2940 unit tests pass; `elm make src/elm/Main.elm` compiles 548 modules; `elm-format --validate` clean.

## Not in this PR

The rest of #2009 — a value stored without the sign that shows its input, then re-saved unchecked. **The production query found zero such records** (0 of 32 heights, 0 of 129 Apgar scores), so it is a latent hole rather than live damage, and it wants its own change.

The 57 records already holding impossible scores are not repaired here: a nurse only meets the check on re-save. Worth deciding separately whether they need a cleanup.

### DIFF
--- a/client/src/elm/Measurement/Utils.elm
+++ b/client/src/elm/Measurement/Utils.elm
@@ -393,8 +393,11 @@ outOfRangeAsEntered constraints measurement value =
         [ measurement ]
 
 
-{-| The range a measurement of the body has to be within, in the unit it is
-entered in.
+{-| The range a measurement has to be within, in the unit it is entered in.
+
+The Apgar scores have one too, though they are neither of the body nor carried
+in a unit: they are scores out of ten.
+
 -}
 measurementConstraints : Site -> RangedMeasurement -> FloatInputConstraints
 measurementConstraints site measurement =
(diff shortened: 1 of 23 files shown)


anvmn (commented):
Findings from `/code-review medium`. All four verified against the code. Two of them fail `elm-review`, so CI on this branch will be red — my fault for not running it before pushing, which is the rule I follow on every other branch.

client/src/elm/Pages/WellChild/Activity/Update.elm
```diff
@@ -15,16 +15,16 @@ import Gizra.Update exposing (sequenceExtra)
...
 import Measurement.Model
     exposing
-        ( AnthropometricMeasurement(..)
-        , ImmunisationTask(..)
+        ( ImmunisationTask(..)
+        , RangedMeasurement(..)
         , VaccinationFormViewMode(..)
         , emptyPhotoForm
         )
 import Measurement.Utils exposing (birthWeightOutsideConstraints, contributingFactorsFormWithDefault, heightFormWithDefault, heightOutOfRange, muacFormWithDefault, muacOutOfRange, ncdaFormWithDefault, nutritionFormWithDefault, toAdministrationNoteWithDefault, toContributingFactorsValueWithDefault, toHealthEducationValueWithDefault, toHeightValueWithDefault, toMuacValueWithDefault, toNCDAValueWithDefault, toNutritionCaringValueWithDefault, toNutritionFeedingValueWithDefault, toNutritionFollowUpValueWithDefault, toNutritionFoodSecurityValueWithDefault, toNutritionHygieneValueWithDefault, toNutritionValueWithDefault, toSendToHCValueWithDefault, toVaccinationValueWithDefault, toVitalsValueWithDefault, toWeightValueWithDefault, vaccinationFormWithDefault, vaccineDoseToComparable, weightFormWithDefault, weightOutOfRange)
```
anvmn: **Confirmed — `elm-review` fails here.** `birthWeightOutsideConstraints` appears exactly once in this file, in this import: its only call site was the inline list I replaced with `pregnancySummaryMeasurementsOutOfRange`. `NoUnused.Variables` rejects it (auto-fixable).

Same line 19: `RangedMeasurement(..)` imports the constructors, but they were only used by that deleted list. The bare type is still needed for the `preSaveNutritionAssessment` signature at :1985, so the fix is to drop the `(..)`, not the import.

client/src/elm/Pages/WellChild/Activity/Utils.elm
```diff
@@ -210,20 +210,49 @@ expectActivity currentDate zscores site features isChw assembled db activity =
...
+    in
+    List.filterMap
+        (\( measurement, value ) ->
+            if measurementOutOfRange site measurement value then
+                Just measurement
+
+            else
+                Nothing
+        )
+        (( MeasurementBirthWeight, Maybe.map (\(WeightInGrm weight) -> weight) form.birthWeight )
```
anvmn: **The order is wrong, and the test that asserts it is named after the thing it gets wrong.**

This returns birth weight, Apgar 1 min, Apgar 5 min, birth length. But `viewPregnancySummaryForm` renders `estimatedDueDateSection ++ apgarScoresSection ++ birthWeightSection ++ birthLengthSection` (`View.elm:571-575`) — verified — so the form asks for the **Apgar scores before the birth weight**.

A nurse who enters both a kilogram birth weight and an impossible Apgar sees the warning list them in the opposite order to the fields in front of her. And the new test is called *"every measurement that is wrong is named, in the order the form asks them"* while asserting the order the form does **not** ask them in — so it locks the defect in under a name that claims the opposite.

Moving the birth-weight pair after `apgar` fixes the behaviour and makes the test's name true.

client/src/elm/Measurement/Utils.elm
```diff
@@ -385,9 +396,15 @@ outOfRangeAsEntered constraints measurement value =
 {-| The range a measurement of the body has to be within, in the unit it is
```
anvmn: Stale docblock: "The range **a measurement of the body** has to be within, in the unit it is entered in." This function now also answers for the two Apgar variants, which are neither of the body nor carried in a unit.

This is the same staleness the commit deliberately fixed on the `RangedMeasurement` type doc two files over — I updated the type's comment and missed this one.

PR #2023, opened 2026-07-30
Measurements: make sure a measurement the record still holds is asked for

### LINKED ISSUE #2009: Measurements: A birth length can be stored without the sign that shows its input, and is then re-saved unchecked

anvmn:
Fixes #2009. Part of #2006. Based on #2021, which renames the type this touches and replaces `birthLengthOutsideConstraints`.

## The problem

A measurement and the sign that says it was asked for are stored apart, and `decodePregnancySummaryValue` falls back to `NoPregnancySummarySigns` when the signs cannot be read. A record in that state loads as:

- `birthLengthAvailable = Just False` → **the input is not drawn**
- `birthLength = Just (HeightInCm 0.5)` → **the value is still live**

The nurse sees nothing, can correct nothing, and every save writes `0.5` straight back out. The range check cannot reach it either, because it deliberately asks only about what the form shows — so the one thing #1998 was built to stop goes through untouched. The Apgar scores have the same shape behind their own question.

## The fix — option 2 from the issue

`pregnancySummaryFormWithDefault` now derives availability from **the value as well as the sign**:

```elm
asked sign values =
    List.member sign signsFromValue
        || List.any isJust values
```

What is stored becomes visible; the nurse confirms or corrects it; saving writes the sign back, so the record heals itself. The range check then covers it, with no change to the check.

**Nothing is cleared.** Option 1 in the issue (don not write unless available) would silently delete a stored length whenever the sign was missing — data loss rather than a display problem.

**The sign is still asked on its own**, so answering "yes" and saving before entering anything does not hide the question again.

## Production data — this is latent, not live

Read-only query against `ihangane.live`, all 596 `well_child_pregnancy_summary` records:

| | records | value without its sign | sign without a value |
|---|---|---|---|
| birth length | 32 | **0** | 0 |
| Apgar scores | 129 | **0** | 0 |

So the decoder fallback has never actually fired in production. This closes the hole before it opens rather than repairing damage — which is also why it costs nothing: deriving availability from a value does nothing when there is no value.

## Tests

7 new cases, and both halves of the condition discrimination-tested separately:

| Broken | Caught by |
|---|---|
| only the sign is asked (the bug) | `still asks for the birth length it holds`, `…the Apgar scores it holds`, `…when it holds only one of them`, `and the measurement it holds is then checked` |
| only the value is asked | `a sign with nothing stored is still asked for` |

That second one closed a gap in my own tests: the first run of that mutation failed nothing, so nothing pinned the sign-present-value-absent case.

Also covered: a form holding nothing is asked for nothing (so no empty inputs appear as tasks still to do), and the nurse answering "no" wins over what is stored.

2953 unit tests pass; `elm make src/elm/Main.elm` compiles 548 modules; `elm-format --validate` and `elm-review` clean.

### DIFF
--- a/client/src/elm/Pages/WellChild/Activity/Utils.elm
+++ b/client/src/elm/Pages/WellChild/Activity/Utils.elm
@@ -281,6 +281,18 @@ pregnancySummaryFormWithDefault form saved =
 
                     signsFromValue =
                         EverySet.toList value.signs
+
+                    -- A record can hold a measurement without the sign that
+                    -- says it was asked for: the two are stored apart, and the
+                    -- signs fall back to none when they cannot be read. Asking
+                    -- the value as well as the sign makes what is stored
+                    -- visible, so the nurse can correct or confirm it and the
+                    -- range check can reach it. Were only the sign asked, the
+                    -- input would stay hidden while the measurement it holds
+                    -- was written back out on every save.
+                    asked sign values =
+                        List.member sign signsFromValue
+                            || List.any isJust values
                 in
                 { expectedDateConcluded = or form.expectedDateConcluded (Just value.expectedDateConcluded)
                 , dateSelectorPopupState = form.dateSelectorPopupState
@@ -288,13 +300,17 @@ pregnancySummaryFormWithDefault form saved =
                     or form.deliveryComplicationsPresent
                         (listNotEmptyWithException NoDeliveryComplications deliveryComplications |> Just)
                 , deliveryComplications = or form.deliveryComplications (Just deliveryComplications)
-                , apgarScoresAvailable = or form.apgarScoresAvailable (List.member ApgarScores signsFromValue |> Just)
+                , apgarScoresAvailable =
+                    or form.apgarScoresAvailable
+                        (asked ApgarScores [ value.apgarOneMin, value.apgarFiveMin ] |> Just)
                 , apgarOneMin = maybeValueConsideringIsDirtyField form.apgarDirty form.apgarOneMin (Maybe.andThen .apgarOneMin saved)
                 , apgarFiveMin = maybeValueConsideringIsDirtyField form.apgarDirty form.apgarFiveMin (Maybe.andThen .apgarFiveMin saved)
                 , apgarDirty = form.apgarDirty
                 , birthWeight = maybeValueConsideringIsDirtyField form.birthWeightDirty form.birthWeight (Maybe.andThen .birthWeight saved)
                 , birthWeightDirty = form.birthWeightDirty
-                , birthLengthAvailable = or form.birthLengthAvailable (List.member BirthLength signsFromValue |> Just)
+                , birthLengthAvailable =
+                    or form.birthLengthAvailable
+                        (asked BirthLength [ Maybe.map (\(HeightInCm length) -> length) value.birthLength ] |> Just)
                 , birthLength = maybeValueConsideringIsDirtyField form.birthLengthDirty form.birthLength (Maybe.andThen .birthLength saved)
                 , birthLengthDirty = form.birthLengthDirty
                 , birthDefectsPresent =
--- a/client/src/elm/Pages/WellChild/Activity/Test.elm
+++ b/client/src/elm/Pages/WellChild/Activity/Test.elm
@@ -321,6 +321,14 @@ savedPregnancySummary =
     }
 
 
+{-| The same record with its signs lost, which is what a decoder that cannot
+read them leaves behind. The measurements are still there.
+-}
+savedWithoutSigns : PregnancySummaryValue
+savedWithoutSigns =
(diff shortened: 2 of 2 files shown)


client/src/elm/Pages/WellChild/Activity/Utils.elm
```diff
@@ -281,20 +281,36 @@ pregnancySummaryFormWithDefault form saved =
...
+
+                    -- A record can hold a measurement without the sign that
+                    -- says it was asked for: the two are stored apart, and the
+                    -- signs fall back to none when they cannot be read. Asking
+                    -- the value as well as the sign makes what is stored
+                    -- visible, so the nurse can correct or confirm it and the
+                    -- range check can reach it. Were only the sign asked, the
+                    -- input would stay hidden while the measurement it holds
+                    -- was written back out on every save.
+                    asked sign values =
```
anvmn, line 293: **Real in principle, unreachable in the data — and the blocking half is not new.**

The finding: `List.any` means a record holding only one of the two Apgar scores (signs lost) draws both inputs, and the empty second one blocks the save, whose only escape is answering "No" — which clears both and discards the score that was there.

Two things settle it.

**Partial Apgar records do not exist.** Read-only query against `ihangane.live`, all 596 records:

| both scores | only 1-minute | only 5-minute | partial **and** sign lost |
|---|---|---|---|
| 129 | **0** | **0** | **0** |

**And the blocking behaviour predates this change.** This clause only alters what happens when the sign is *missing*. Whenever the sign is present — which is all 129 records — `List.member` already returned `True`, so both inputs were already drawn and an empty second score already blocked the save. Nothing here makes that worse; it extends the same treatment to a state that has never occurred.

The destructive escape hatch is worth knowing about, but it belongs to the bool handler (`View.elm:285-298`), which has cleared both scores on "No" since the feature landed and is untouched here.

So: no change. Recorded rather than acted on, and the query is in the description if it is worth re-checking before a release.

anvmn (commented):
One finding from `/code-review medium`, below. It asked for a fact rather than a judgement — whether partial Apgar records exist — so I queried production rather than reasoning about it. They do not.

The pass also independently confirmed the premise and the reach of the fix: the decoder fallback collapses the whole sign set on any unrecognised string; `PreSavePregnancySummary` checks the *resolved* form, so the healed value does reach the range check; the `ApgarScores`/`BirthLength` signs have no other consumers, so healing them on save cannot desync a report; and activity completion keys off measurement presence, so no encounter flips to incomplete.

Issue #1975, opened 2026-07-21
Batch scripts: batch scripts page by node id without ordering, so records can be passed over


anvmn:
Backlog item **B-099**.

## Mechanism

`hedley_ncda/scripts/generate-data-for-all.php` and its `hedley_reports` twin, plus the ten `hedley_reports/completion-generate-*.php` scripts, all page through nodes with a keyset walk:

```php
$query->propertyCondition('nid', $nid, '>');
$result = $query->range(0, $batch)->execute();
$ids = array_keys($result['node']);
// ...
$nid = end($ids);
```

None of them asks the database for an order. Nothing guarantees the last row of a batch carries the highest nid, and if it doesn't, every node between the true maximum and the value taken by `end($ids)` is stepped over and never processed — silently. The base query comes from `hedley_general_create_entity_field_query_excluding_deleted()`, whose tag adds a LEFT JOIN, and eight of the completion scripts also carry a `fieldCondition`, which routes the query through field storage with the entity table joined in — both give the planner more latitude in how rows come back.

## Status of the claim

The mechanism is clear from the code. **Whether it actually occurs on MariaDB 10.5 for these queries has not been observed** — it depends on the planner. So the fix is a guard and a consistency correction, not a repair of a reproduced failure.

It is also the minority pattern: of 35 scripts using `propertyCondition('nid')`, **14 already order by nid** and 21 do not. `hedley_stats/scripts/recalculate-stats.php` is structurally identical to `generate-data-for-all.php` and does order, which is the shape the fix follows.

## Fix

Add `->propertyOrderBy('nid')` to the base query in the 12 affected scripts.

## If it has been happening

Worth being explicit, since the fix prevents future skips but repairs nothing already missing:

- **NCDA** — `generate-data-for-all.php` supports `--exclude_set=1`, which processes only persons whose data was never set. That is a cheap way to fill any gaps.
- **Completion reports** — no equivalent flag; closing a gap means a full re-run from nid 0.

Neither is triggered by this change; they are operator decisions if a gap is ever suspected.

## Not included

Eight other keyset-paging scripts also lack the ordering and are deliberately left alone — see the follow-up issue. `hedley_migrate/scripts/validate-villages-migration.php` uses `propertyCondition('nid')` but does not keyset-page, so it is unaffected.

anvmn:
Delivered in #1977 (merged to `develop`).

PR #2024, opened 2026-08-02
Batch scripts: drop the ones nobody runs

### LINKED ISSUE #1976: Batch scripts: eight other batch scripts still page by node id without ordering
Follow-up to #1975, which fixed 12 of the 20 keyset-paging scripts. Filing so the split is a recorded decision rather than something rediscovered later.

## Still unordered, and they do keyset-page

| Script | Note |
|---|---|
| `hedley_admin/scripts/generate-nutrition-report.php` | admin-scripts family — verify liveness first |
| `hedley_admin/scripts/locate-duplicate-measurements.php` | same family |
| `hedley_health_center/scripts/migrate-prenatal-participants.php` | looks like a migration one-shot |
| `hedley_migrate/scripts/export-patients.php` | already has its own open issue for unrelated defects |
| `hedley_person/scripts/migrate-children.php` | looks like a migration one-shot |
| `hedley_person/scripts/migrate-field-child.php` | looks like a migration one-shot |
| `hedley_reports/scripts/backfill-prenatal-gwg-indicators.php` | previously assessed as a **retired one-shot** — likely moot |
| `hedley_schedule/scripts/migrate-clinic.php` | looks like a migration one-shot |

`hedley_migrate/scripts/validate-villages-migration.php` uses `propertyCondition('nid')` but does **not** keyset-page, so it is not affected and needs nothing.

## Why they were left out of #1975

The 12 fixed there are all live: the two `generate-data-for-all.php` scripts and the ten completion generators are driven by E2E infrastructure and by new-site enablement, so the ordering matters on an ongoing basis.

The eight above are mostly one-shot migration scripts whose liveness is unverified. Existing practice here is to confirm a script is actually still used before changing it, because a real code gap in a dead script is moot — and at least one of these (`backfill-prenatal-gwg-indicators.php`) has already been assessed as retired. Patching all eight blindly would add churn across scripts nobody runs and would make them look maintained when they are not.

## Suggested resolution

Decide per script: confirm liveness, then either add `->propertyOrderBy('nid')` or delete the script. Doing this as one pass would leave the whole family in a known state.

## The deeper point

The batch walk itself — clone the base query, `range(0, $batch)`, `node_load_multiple`, `$nid = end($ids)` — is hand-copied across roughly 20 scripts, which is why 21 of 35 ended up missing the ordering. There is nothing shared to omit it from. A helper owning the batch walk, alongside the existing `hedley_general_create_entity_field_query_excluding_deleted()`, would make the ordering impossible to forget and is probably the real fix for this class.

anvmn:
Fixes #1976.

Nine scripts keyset-page through nodes — `propertyCondition('nid', $nid, '>')`, `range(0, $batch)`, `$nid = end($ids)` — without `propertyOrderBy('nid')`. Without an ORDER BY, MySQL may return rows in any order, so the walk can revisit some records and skip others. #1975 fixed the twelve of that family that are live. #1976 recorded the rest as a decision to make rather than a patch to apply, because most looked like one-shot migrations whose liveness was unverified, and ordering a script nobody runs makes it look maintained.

## The decision

The user confirmed these are not in use. Deleted:

| Script | What it was |
|---|---|
| `hedley_admin/scripts/locate-duplicate-measurements.php` | diagnostic |
| `hedley_admin/scripts/generate-nutrition-report.php` | weekly nutrition report |
| `hedley_health_center/scripts/migrate-prenatal-participants.php` | one-shot migration |
| `hedley_migrate/scripts/export-patients.php` | export tool |
| `hedley_person/scripts/migrate-children.php` | one-shot migration |
| `hedley_person/scripts/migrate-field-child.php` | one-shot migration |
| `hedley_reports/scripts/backfill-prenatal-gwg-indicators.php` | retired backfill |
| `hedley_schedule/scripts/migrate-clinic.php` | one-shot migration |

`hedley_migrate/scripts/validate-villages-migration.php` uses `propertyCondition('nid')` but does not keyset-page, so it was never affected and is untouched.

## What pointed at the nutrition report

Two things, both removed with it, or the deletion would have left a call to a file that is not there:

- `infrastructure_setup/reporting.sh` — the reporting cron, which invoked it against Pantheon live via `terminus remote:drush`
- `reportNutrition($region)` in `server/RoboFile.php`
- the `reporting.sh` line in the README crontab example

The other three reports that cron ran — demographics, ANC, closed pregnancies — keep their Robo tasks (`reportDemographics`, `reportAnc`, `reportPregnancy`), so they remain callable and are not orphaned by this.

## Left alone, deliberately

`hedley_person/scripts/migrate-field-mother.php:7` still says *"Run migrate-mothers and migrate-children first"*, and `migrate-children.php` is now gone. `migrate-mothers.php` and `migrate-field-mother.php` are the same one-shot family but do not keyset-page, so they are outside this issue. Worth deciding on separately rather than editing a migration's instructions here.

## The deeper point, unaddressed

The batch walk is hand-copied across roughly twenty scripts, which is how 21 of 35 ended up missing the ordering. A shared helper owning the walk, next to `hedley_general_create_entity_field_query_excluding_deleted()`, is what stops this recurring. Out of scope here.

## Verification

`php -l` clean on `RoboFile.php`; phpcs Drupal + DrupalPractice reports no errors on it. The README long-line warnings are pre-existing (26 on `develop`, 26 here).


server/hedley/modules/custom/hedley_reports/scripts/backfill-prenatal-gwg-indicators.php
```diff
@@ -1,629 +0,0 @@
-<?php
```
anvmn, line 1: **high — this one is not a retired one-shot, and #1976 was wrong about it.**

Added 2026-05-06 in `f7c51114b`, three months ago. `hedley_reports.module:999` reads `adequate-gwg` off `field_prenatal_indicators`, and the only writer of that field is the client, through `HedleyRestfulPrenatalEncounter` — so the app stamps it on encounters saved *after* the feature shipped, and this script is the sole mechanism for populating it on encounters that already existed.

If the backfill has not been run on every live site (rwanda / burundi / somalia), historical prenatal encounters report no GWG data permanently, and deleting this removes the only tool that could fix that. Whether it ran is an operational fact I cannot check from here.

⚠️ Worth confirming before merge.

infrastructure_setup/reporting.sh
```diff
@@ -1,42 +0,0 @@
...
-if [[ $REPORT_RECIPIENT =~ $EMAIL_REGEX ]] ; then
-    echo "Report to $REPORT_RECIPIENT"
-else
-    echo "Reporting is not enabled for $PANTHEON_ID"
-    exit 0
-fi
-
-cat /dev/null > $REPORT_FILE
-terminus remote:drush $PANTHEON_ID scr profiles/hedley/modules/custom/hedley_admin/scripts/generate-demographics-report.php >> $REPORT_FILE
-terminus remote:drush $PANTHEON_ID scr profiles/hedley/modules/custom/hedley_admin/scripts/generate-nutrition-report.php >> $REPORT_FILE
```
anvmn, line 32: **medium — this cron drove four reports, not one.**

It invoked `generate-demographics-report.php`, `generate-nutrition-report.php`, `generate-anc-report.php` and `generate-closed-pregnancies-report.php` against Pantheon live. Only the nutrition one is deleted here; the other three survive and keep their Robo tasks (`reportDemographics`, `reportAnc`, `reportPregnancy`), so they stay callable by hand — but their *automated* weekly email stops with this file.

If any Jenkins job or crontab still points at `infrastructure_setup/reporting.sh`, it starts failing on the next checkout, silently. This is covered by the premise that the reporting cron is no longer in use; recording it so the premise is visible rather than implied.

infrastructure_setup/reporting.sh
```diff
@@ -1,42 +0,0 @@
...
-ID="$(date +"%Y-%U")"
-REPORT_FILE="/tmp/eheza-app-weekly-report-$ID-$PANTHEON_ID.txt"
-
-if [[ -f "$REPORT_FILE.success" ]];
-then
-  echo "Report already exists for this week"
-  exit 0
-fi
-
-export REPORT_RECIPIENT=$(terminus remote:drush $PANTHEON_ID vget hedley_admin_report_email -- --format=string)
```
anvmn, line 21: **low — `hedley_admin_report_email` loses its only reader.**

This `vget` was the only consumer. `hedley_admin.module:98` still exposes the setting in the admin form, so after this PR an admin can fill in a report recipient and get nothing back, with no indication why.

Either drop the setting or keep a consumer. Not a blocker, but it leaves a dead control in the UI.

README.md
```diff
@@ -152,10 +152,9 @@ job from those or you can invoke them via `cron` or
 top of the scripts and study the file head comment that contains more
 information of the dependencies of the scripts.
 
-Example crontabs:
+Example crontab:
```
anvmn, line 152: **low — the surrounding docs still describe the job this removes.**

I took out the crontab line, but the section above still lists "Weekly report processing" as one of two recurring jobs to configure and says "There are two main alternatives", and line ~166 still advises "For the reporting, a simple cron job might be sufficient". `infrastructure_setup/` now contains only `advancedqueue.sh`, so a deployer following this looks for a script that is not there.

server/hedley/modules/custom/hedley_person/scripts/migrate-children.php
```diff
@@ -1,126 +0,0 @@
-<?php
```
anvmn, line 1: **low — the legacy person-migration chain is left half-deleted.**

`migrate-children.php` and `migrate-field-child.php` go; `migrate-mothers.php` and `migrate-field-mother.php` stay, because they do not keyset-page and so fall outside #1976. `migrate-field-mother.php:7` still instructs "Run migrate-mothers and migrate-children first", naming a script that is no longer in the tree.

Flagged in the PR description as a deliberate scope boundary: delete all four or keep all four, but that is a decision beyond this issue.

server/hedley/modules/custom/hedley_admin/scripts/locate-duplicate-measurements.php
```diff
@@ -1,133 +0,0 @@
-<?php
```
anvmn, line 1: **low — the read-only diagnostic goes while both destructive counterparts stay.**

`delete-duplicate-measurements.php` and `delete-duplicates.php` are kept. Operators lose the ability to inspect duplicates before running a deletion under `hedley_super_user_mode`, which is the safer half of that pair.

anvmn (commented):
Findings from `/code-review medium` on this branch. I verified each against the code before posting; my analysis of which ones should change the PR is in the conversation.

The reviewer confirmed the safety property this PR rests on: no surviving code, `.info` file, RoboFile task, CI config or doc references any deleted script, no deleted file was `require_once`'d by a survivor, and the functions they defined have no external callers. All 10 CI checks pass. The findings below are about consequences outside the diff, not about the deletions being unsafe.

server/hedley/modules/custom/hedley_reports/scripts/backfill-prenatal-gwg-indicators.php
anvmn, line 1: **Resolved — verified against all three live sites, read-only.**

The backfill ran where it mattered.

**Rwanda (`ihangane`)** — 23,985 prenatal encounters created *before* 2026-05-06 carry a GWG verdict. The field is only ever written by the client through `HedleyRestfulPrenatalEncounter`, so pre-feature encounters could not have acquired it any other way. Values: `adequate-gwg` 17,705, `inadequate-gwg` 6,307 (plus 735 `past-labs-completed`), against 68,803 prenatal encounters. This is also the only site with `hedley_admin_feature_healthy_start_enabled` set. The ~35% coverage matches the script's documented behaviour: encounters whose verdict cannot be computed — no current weight, no LMP, no pre-pregnancy weight — are skipped.

**Burundi (`vhw`)** — `field_data_field_prenatal_indicators` holds **zero rows** across 1,444 prenatal encounters, including the 312 created since the feature shipped. The indicator is not in use there at all, so there was nothing to back-fill.

**Somalia (`tip-somalia`)** — zero prenatal encounters.

So the deletion destroys no irreplaceable capability. The script is idempotent by design, and recoverable from git history if the classification maths is ever revised.

anvmn:
Review responses, in `dc4472fe7`.

**Fixed**

- **`locate-duplicate-measurements.php` is restored.** Checking it before removing it: it already had `->propertyOrderBy('nid', 'ASC')` on the base query it clones each iteration, so it never had the defect #1976 attributed to it — the issue listed it in error. It is also the only way to inspect duplicates before running either of the two scripts that delete them, both of which are kept. Deleting the safe half of that pair would have made the remaining workflow more dangerous.
- **`hedley_admin_report_email` is gone**, along with the `admin/config/hedley-reporting` page, its menu entry and `hedley_admin_settings_form()`. The reporting cron's `vget` was its only reader, so the field would have accepted an address that nothing reads. `hedley_admin_update_7003()` deletes the stored variable.
- **The rest of the person-migration chain goes.** `migrate-mothers.php` and `migrate-field-mother.php` join the two already removed. Half a chain runs nothing, and the surviving file's header instructed running `migrate-children` first, which is no longer here. Nothing in the repo now references any of the four.
- **The README** no longer lists weekly report processing as a recurring job, and no longer advises a cron for it.

**Not changed, deliberately**

- The GWG backfill deletion stands — see the verification against all three live sites on that file.
- Removing `reporting.sh` does stop the automated demographics / ANC / closed-pregnancies emails. That is the premise of this PR rather than an oversight: the cron is no longer in use. All three scripts survive with their Robo tasks (`reportDemographics`, `reportAnc`, `reportPregnancy`) and remain callable.

`php -l` and phpcs (Drupal + DrupalPractice) are clean on every changed PHP file.

PR #2026, opened 2026-08-03
Patient merge: move the person references a merge leaves behind

### LINKED ISSUE #2025: Patient merge: patient merge leaves two person references behind: WhatsApp records and trace-contact referrals
The last two gaps from the patient-merge coverage audit (2026-07-14). The audit found six fields that point at a person; four are handled, two are not.

## What is missed

**`whatsapp_record.field_person`** — names the person a shared report is about. The bundle carries no encounter indicator, so it is not a measurement type and `_consolidate_patients_execute()`'s measurement sweep never sees it. It also has no `field_deleted`, so the person-delete cascade cannot soft-delete it either. The record simply keeps naming a person who is no longer there.

**`acute_illness_trace_contact.field_referred_person`** — the person who was contacted. This is the one case where the merge *does* see the node: a trace contact is a measurement, so its `field_person` (the person who reported it) is re-pointed. The second reference is not.

Both fields are already named in `hedley_general_get_person_association_fields()`, the canonical list added as the root-cause fix for this class — so the knowledge is recorded, the merge just does not act on it.

## Scale on live, read-only, 2026-08-03

| | Rwanda | Burundi |
|---|---|---|
| `whatsapp_record` rows | 275 | 22 |
| `acute_illness_trace_contact` rows | 0 | 0 |
| either field pointing at a soft-deleted person | **0** | **0** |
| soft-deleted persons | 159 | 0 |

**Nothing is broken on live today.** 159 merges on Rwanda have not once involved a person who had a WhatsApp record, and the trace-contact bundle has never held a row anywhere. This is worth doing to close the class, not to repair damage: with these two, every field in the canonical list is handled and the merge-gap group is provably complete.

Note `whatsapp_record` is growing — 270 rows when audited three weeks ago, 275 now — so the probability of a merge meeting one is small but rising.

## Fix

One function each, both built on `_hedley_patient_content_referencing_person()`, the helper introduced for the education-session transfer, called from `_consolidate_patients_execute()` alongside the existing three. No new way of discovering content that references a person.

No shard handling in either: `whatsapp_record` has no `field_shards` at all, and a trace contact's shards derive from `field_person`, which the measurement sweep already handles.

Backlog: B-115, B-117.

anvmn:
Fixes #2025.

The merge walks the duplicate patient's measurements, then re-points the references that are not measurements by hand. Three were still missed — the last three of the six in `hedley_general_get_person_association_fields()`.

## The change

**`_consolidate_patient_group_participations()`** — a `pmtct_participant` names the enrolled child in `field_person` and the adult who brings them in `field_adult`. **This is the one that loses data rather than dangling.** `field_adult` IS followed by `hedley_person_content_affected_by_deletion()`, so marking the duplicate deleted soft-deleted the participation — and a child with no involvement in the merge silently lost a group enrolment. Dedups against a participation the original already holds for the same child and group, and drops one where the original is the enrolled child rather than making it its own adult. Shards are cleared so the presave hook recomputes the union of both persons, as the relationship transfer does.

Participations where the duplicate is the *child* are deliberately untouched: `_consolidate_session_content_validate()` already refuses the merge unless the original participates in every group the duplicate has measurements in.

**`_consolidate_patient_whatsapp_records()`** — a `whatsapp_record` names the person a report was shared about, carries no encounter indicator (so the measurement sweep never finds it) and has no `field_deleted` (so the delete cascade leaves it in place). It kept naming a person who is no longer there.

**`_consolidate_patient_trace_contact_referrals()`** — a trace contact IS a measurement, so the sweep re-points `field_person`, the person who reported it. Its second reference, `field_referred_person`, stayed on the duplicate.

All three use `_hedley_patient_content_referencing_person()`, so none adds a new way of discovering content that references a person.

## Also

Corrects a comment in `hedley_patient.module` that said a trace contact takes its shards from the encounter it was recorded at. It takes them from `field_person`, which is why that field reaches trace contacts in the recompute and `field_referred_person` is the one excluded.

## Shards, checked rather than assumed

- `pmtct_participant` — two-person node, union of both, and `hedley_device_assign_shards()` skips an already-populated field, so the field is cleared before save.
- `whatsapp_record` — **has no `field_shards` at all**. My first draft cleared it and would have thrown.
- `acute_illness_trace_contact` — has shards, but they derive from `field_person`, which the measurement sweep already handles.

## Tests

Seven cases, 12-18, folded into `HedleyPatientConsolidation::testContentIsTransferredOnMerge()` as extra assertions rather than new methods, so the group keeps paying for one profile install:

- 12-13 WhatsApp records, directly and through the full merge
- 14-15 trace-contact referrals, including that the reporter in `field_person` is left alone
- 16 the participation survives the full merge, with the original as the adult and the child untouched — the case that fails on `develop`, where the cascade soft-deletes it
- 17 a participation the original already holds for the same child and group is dropped, not duplicated
- 18 a participation naming the original on both sides is dropped

## Scope, honestly

The WhatsApp and trace-contact halves repair nothing on live: zero records of either kind point at a soft-deleted person today, against 159 soft-deleted persons on Rwanda. The `field_adult` half is different in kind — it is silent data loss, and any past merge of an adult who brought a child to a group has already caused it. Repairing those is not in this PR; it belongs with B-120, which scopes retroactive repair of what past merges destroyed.

## Verification

`php -l` and phpcs Drupal + DrupalPractice clean on all three files. SimpleTest runs in CI — `ddev simpletest` executes the main tree, not a worktree.

## Review

`/code-review medium` found three, all verified and all handled: the missing `field_adult` (this PR now covers it), the trace-contact shard comment (corrected), and an unrelated pre-existing crash in `hedley_patient_recalculate_shards_for_person_content()`, which is being fixed separately.

### DIFF
--- a/server/hedley/modules/custom/hedley_patient/hedley_patient.drush.inc
+++ b/server/hedley/modules/custom/hedley_patient/hedley_patient.drush.inc
@@ -242,6 +242,11 @@ function _consolidate_patients_execute($original, $duplicate) {
   // person who no longer exists.
   $messages[] = _consolidate_patient_newborn_references($original, $duplicate);
 
+  // A group participation names the adult who brings the child, and that is
+  // the reference the delete cascade DOES follow - so leaving it behind would
+  // not dangle, it would soft-delete another child's enrolment.
+  $messages[] = _consolidate_patient_group_participations($original, $duplicate);
+
   // A WhatsApp record names the person a report was shared about. It carries no
   // encounter indicator, so it is not a measurement and the loop above never
   // sees it either.
@@ -521,6 +526,84 @@ function _consolidate_patient_whatsapp_records($original, $duplicate) {
 
   return format_string('WhatsApp records: @count re-pointed to the original.', [
     '@count' => count($record_ids),
+  ]);
+}
+
+/**
+ * Re-points group participations the duplicate patient holds as the adult.
+ *
+ * A pmtct_participant enrols a child in a group, and names the adult who brings
+ * them in field_adult. That is the one person reference where leaving the
+ * duplicate in place does not merely dangle: field_adult IS followed by the
+ * person-delete cascade, so marking the duplicate deleted soft-deletes the
+ * participation - and the child, who has nothing to do with this merge, loses
+ * a group enrolment.
+ *
+ * Participations where the duplicate is the CHILD are a separate matter, left
+ * alone here: _consolidate_session_content_validate() refuses the merge unless
+ * the original already participates in every group the duplicate has
+ * measurements in, so the enrolment the child needs is already there.
+ *
+ * Shards are cleared so the presave hook recomputes them. A participant is a
+ * two-person node - hedley_device_assign_shards() gives it the union of both
+ * persons' shards - and it leaves an already-populated field alone, so without
+ * this the participation would keep the duplicate's shards and miss any health
+ * center only the original belongs to.
+ *
+ * @param int $original
+ *   Node ID of the 'original' patient.
+ * @param int $duplicate
+ *   Node ID of the 'duplicate' patient.
+ *
+ * @return string
+ *   Success message.
+ *
+ * @throws EntityMetadataWrapperException
+ */
+function _consolidate_patient_group_participations($original, $duplicate) {
+  $participant_ids = _hedley_patient_content_referencing_person($duplicate, [
+    'field_adult',
+  ], 'pmtct_participant');
+
+  // What the original already brings a child to, keyed child + group, so a
+  // participation it already holds is not created twice.
+  $existing = [];
+  $original_ids = _hedley_patient_content_referencing_person($original, [
+    'field_adult',
+  ], 'pmtct_participant');
+  foreach ($original_ids as $id) {
+    $wrapper = entity_metadata_wrapper('node', $id);
+    $key = $wrapper->field_person->getIdentifier() . ':' . $wrapper->field_clinic->getIdentifier();
+    $existing[$key] = TRUE;
+  }
+
+  $repointed = $dropped = 0;
+  foreach ($participant_ids as $id) {
+    $wrapper = entity_metadata_wrapper('node', $id);
+    $child = $wrapper->field_person->getIdentifier();
+    $key = $child . ':' . $wrapper->field_clinic->getIdentifier();
+
+    if ($child == $original || isset($existing[$key])) {
+      // Either the original is the child of this participation, so re-pointing
+      // would make it its own adult, or it already brings that child to that
+      // group. Redundant either way.
+      $wrapper->field_deleted->set(TRUE);
+      $wrapper->save();
+      $dropped++;
+      continue;
+    }
+
+    $wrapper->field_adult->set($original);
+    $wrapper->field_shards->set([]);
+    $wrapper->save();
+    // A later participation of the duplicate might now duplicate this one.
+    $existing[$key] = TRUE;
+    $repointed++;
+  }
+
+  return format_string('Group participations: @repointed re-pointed to the original, @dropped dropped as redundant.', [
+    '@repointed' => $repointed,
+    '@dropped' => $dropped,
   ]);
 }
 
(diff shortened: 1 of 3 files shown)


server/hedley/modules/custom/hedley_patient/hedley_patient.drush.inc
```diff
@@ -242,6 +242,16 @@ function _consolidate_patients_execute($original, $duplicate) {
...
 
+  // A WhatsApp record names the person a report was shared about. It carries no
+  // encounter indicator, so it is not a measurement and the loop above never
+  // sees it either.
+  $messages[] = _consolidate_patient_whatsapp_records($original, $duplicate);
+
+  // A trace contact IS a measurement, so the loop above re-pointed the person
+  // who reported it. It holds a second reference - the person who was contacted
+  // - which the loop does not touch.
+  $messages[] = _consolidate_patient_trace_contact_referrals($original, $duplicate);
```
anvmn, line 258: **medium — the commit claims to move "the last two person references a merge leaves behind", and that is wrong.**

`field_adult` is still unhandled. `hedley_general_get_person_association_fields()` lists six; the merge now covers `field_person`, `field_related_to`, `field_participating_patients`, `field_newborn` and `field_referred_person`. `grep field_adult` in this file returns nothing.

The failure is not a dangling reference — it is data loss. When the duplicate is an adult who is `field_adult` on **another child's** `pmtct_participant`, the merge leaves that reference, then sets `field_deleted = TRUE`. `hedley_person_content_affected_by_deletion()` queries `field_adult` and keeps bundle `pmtct_participant`, so the cascade soft-deletes it: the other child silently loses a group enrolment because someone else's duplicate was merged.

This is the same shape as the relationship gap, and it is not covered by `_consolidate_session_content_validate()`, which only checks the clinics the *duplicate* has measurements in — not participants where the duplicate is the adult for someone else.

Note a fix needs the two-person shard treatment: a `pmtct_participant` links two people, so its shards are the union (`hedley_general_two_person_node_shards`), and dedup matters because the original may already hold a participant for the same child and clinic.

anvmn (commented):
Findings from `/code-review medium` on this branch. I verified all three against the code before posting; analysis and proposed handling are in the conversation.

The reviewer confirmed the new code itself is sound: the bundle filter, the non-deleted filter (`whatsapp_record` has no `field_deleted`, and `hedley_general_apply_exclude_deleted` uses a LEFT JOIN + `isNull`, so those records are correctly included), the ordering before `field_deleted = TRUE`, and both docblocks' claims about the delete cascade.

**One finding cannot be anchored to a line, because its file is not in this diff.** `hedley_patient.module:172` — `hedley_patient_recalculate_shards_for_person_content()` keeps `field_person` in its `$association_fields`, so the query returns `whatsapp_record` nodes, and the loop then calls `$wrapper->field_shards->value()` on every node it loaded. `whatsapp_record` has no `field_shards` instance, so that throws `EntityMetadataWrapperException: Unknown data property field_shards`, aborting the person save. Trigger: a person who has any WhatsApp record has their health center changed — `hedley_patient_node_update` returns early only when the shards are unchanged. **Pre-existing and independent of this PR** (275 such records on Rwanda today), but this PR is what surfaced it, and my new docblock's "nothing to recompute" is true of the merge yet misleading about that loop.

server/hedley/modules/custom/hedley_patient/hedley_patient.drush.inc
```diff
@@ -479,6 +489,75 @@ function _consolidate_patient_newborn_references($original, $duplicate) {
...
+/**
+ * Re-points trace contacts that refer to the duplicate patient.
+ *
+ * An acute_illness_trace_contact is a measurement, so the sweep above already
+ * re-points its field_person - the person who reported the contact. It holds a
+ * second person reference the sweep does not touch: field_referred_person, the
+ * person who was contacted. When the duplicate is that person, the trace
+ * contact keeps naming them after the merge.
+ *
+ * No shard handling: a trace contact's shards derive from field_person, which
```
anvmn, line 619: **low — this docblock and `hedley_patient.module:160-163` say opposite things about trace-contact shards.**

Mine says they derive from `field_person`; the module comment says "a trace contact takes them from the encounter it was recorded at", and uses that to justify excluding `field_referred_person` from the recompute.

Mine is the correct one: `acute_illness_trace_contact` is not an individual or family encounter type and has no special case in `hedley_device_assign_shards()`, so it falls to `default:`, which reads `field_person` and calls `hedley_general_shards_by_person()`. The recompute loop agrees — it overwrites a trace contact's shards with the person's, since the bundle is not in the `pmtct_participant`/`relationship` special case.

No behavioural bug either way; the module comment should be corrected so the next person reasoning about this is not sent the wrong way.

PR #2030, opened 2026-08-03
Exports: stop one export from swallowing another

### LINKED ISSUE #2029: Exports: one view export request silently refuses every other
One export request silently refuses every other. Backlog: B-005.

## The mechanism

`hedley_view_export.module:171` queues every export under a hardcoded unique id of `0`:

```php
hedley_general_add_task_to_advanced_queue_by_id(HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE, 0, $arguments);
```

In `hedley_general_add_task_to_advanced_queue_by_id()` the title becomes `export_view_0` for **every** export, and the union `['title' => $task_title] + $data` means that fixed title wins over the human-readable one. The check then refuses a new task while any task with that title is `QUEUED`, or `PROCESSING` and under 20 minutes old.

So while one export is running, every other export request in the system is refused — and a multi-batch export returns to `QUEUED` between batches, so that window is its entire lifetime, with processing only happening on the five-minute `advancedqueue.sh` cron.

The refusal is silent three times over: the `FALSE` return is discarded; the *"We will soon email you a link"* message is shown regardless; and the empty file created for the export is orphaned, because `hedley_view_export_cron()` only removes files a queue item points at.

## Scale

Rare, and worth saying plainly: `file_managed` holds **3 exported files on Rwanda, newest 2025-06-23**, and none on Burundi. B-005 was parked on 2026-07-13 for exactly this reason and was revived deliberately, with the fix kept to the size the defect deserves.

## Fix

Give the queue id a meaning — user, view, display and a hash of the filters, since the same view of two health centers is two different exports. Check the return, delete the file nothing will collect, and say that the export was not started.

anvmn:
Fixes #2029.

Every export was queued under the title `export_view_0`, because the unique id passed to `hedley_general_add_task_to_advanced_queue_by_id()` was a hardcoded `0`. While any export was queued — and a multi-batch export returns to queued between batches, so that is its whole life — every other export request in the system was refused. Silently: the return was discarded and the same "we will soon email you a link" message was shown either way, so the second person to ask was told their export was coming and then never received an email.

## The change

The id now names what is being exported — user, view, display and a hash of the filters. The filters matter: the same view of health center A and of health center B are different exports. A refusal now says so, and deletes the empty file created for the request, which the cleanup cron cannot collect because it only walks files a queue item points at.

The message deliberately does not say *which* refusal it was. The two possibilities are a duplicate and a failed queue insert; the second already reports itself, and distinguishing them afterwards costs more than it is worth here.

23 lines, one file.

## Verification

`php -l` and phpcs Drupal + DrupalPractice clean. The dedup behaviour was exercised against the real queue on a local site — same helper, same table:

```
first request (user 7, view A):       queued
same user, same view again:           refused   <- as intended
different user, same view:            queued
same user, different view:            queued
--- old hardcoded 0 ---
user 7 asks:                          queued
unrelated user 9, different view:     refused   <- the bug
```

## Scope

This PR covers B-005 only. **B-087** — the worker abandoning a fully built CSV when the final-batch email fails — was originally in this branch and has since been parked: it needs the mail step retried in isolation, which is expensive to get right, on a feature that has produced three exports in its lifetime, the most recent in June 2025.

Note the branch name still carries `b005-b087` from that earlier scope, and the review threads above refer to a larger version of this change that was reset and rewritten.

### DIFF
--- a/server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
+++ b/server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
@@ -168,9 +168,29 @@ function hedley_view_export_add_item(string $type, string $view_name, string $di
     'range' => variable_get("hedley_view_export_range__{$view_name}", 500),
     'title' => $title,
   ];
-  hedley_general_add_task_to_advanced_queue_by_id(HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE, 0, $arguments);
-  // Display a message and a link back to the view.
-  drupal_set_message(t('Exporting view. We will soon email you a link to @email', ['@email' => $user->mail]));
+  // The unique id has to name what is being exported. It was a hardcoded 0, so
+  // every export shared the title export_view_0: while one was queued - and a
+  // multi-batch export is queued again between batches, so that is its whole
+  // life - every other export request in the system was refused.
+  $unique_id = implode('_', [
+    $user->uid,
+    $view_name,
+    $display,
+    md5(serialize([$exposed_inputs, $context])),
+  ]);
+  $queued = hedley_general_add_task_to_advanced_queue_by_id(HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE, $unique_id, $arguments);
+
+  if (!$queued) {
+    // Nothing will process the file that was created for this request, and the
+    // cleanup cron only removes files a queue item points at.
+    file_delete($file);
+    drupal_set_message(t('This export was not started. If you have just asked for the same one, it is already running.'), 'warning');
+  }
+  else {
+    drupal_set_message(t('Exporting view. We will soon email you a link to @email', ['@email' => $user->mail]));
+  }
+
+  // A link back to the view, whichever of the two messages was shown.
   return '<a href="javascript:history.back()">' . t('Go back') . '</a>';
 }
 


server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -258,23 +287,64 @@ function hedley_view_export_export_batch_worker($item) {
   }
 
   // Reaching here means the results count is less than the range. So this was
-  // the last batch.
-  // Send a link to the exported file to the user who created it.
+  // the last batch: the file holds every row it is ever going to hold. Record
+  // that before trying to send it, so that a failure to send does not cost the
+  // work of building it.
+  $item->data['export_built'] = TRUE;
```
anvmn: **blocker — the `export_built` flag is not persisted on the path that needs it most, so this commit does not actually prevent the duplicate rows it claims to.**

The assignment is in memory only. The single write-back of `$item->data` is `requeueItem()` inside `advancedqueue_process_item()`, which runs *after* the worker returns. If the process dies or the 120s `lease time` expires **during** `drupal_mail()` — a hanging SMTP relay being exactly the failure this PR is about — `_advancedqueue_release_stale_items()` calls `releaseItem()`, and I confirmed that only updates `expire` and `status`:

```php
$update = db_update('advancedqueue')
  ->fields(array(
    'expire' => 0,
    'status' => ADVANCEDQUEUE_STATUS_QUEUED,
  ))
```

No `data`. So the flag is lost, the item is reclaimed as a fresh batch, the last page is rendered again, and `file_put_contents(..., FILE_APPEND)` duplicates its rows in the CSV the user finally receives.

Fix: persist `data` before calling `drupal_mail()` — a direct `db_update` of the `data` column is enough and, unlike `requeueItem()`, does not also reset the status of an item still being processed.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -168,7 +168,29 @@ function hedley_view_export_add_item(string $type, string $view_name, string $di
...
+  // one of them was queued - and a multi-batch export is queued again between
+  // batches, so that is its whole lifetime - every other export request was
+  // refused. The refusal was silent, because the return was discarded and the
+  // message below was shown either way, so the second person to ask for an
+  // export was told it was on its way and then never received an email.
+  //
+  // Keeping the user and the display in the id preserves what the check was
+  // for - one person cannot flood the queue by asking twice for the same
+  // export - without letting unrelated exports block each other.
+  $unique_id = "{$user->uid}_{$view_name}_{$display}";
```
anvmn: **high — the id omits the filters, so distinct exports still block each other, and the failure is worse than a refusal.**

`{$user->uid}_{$view_name}_{$display}` treats "same user, same view, same display" as one task, but `exposed_inputs` and `context` are what make an export mean anything here.

Scenario: a manager exports a view filtered to health center A, then immediately re-runs it for health center B. The second request is refused, its file deleted, and the warning says a link is coming — so they receive the health-center-A file and read it as B's data. Silently wrong data is worse than the original silent refusal.

Fix: fold a hash of `$exposed_inputs` + `$context` into the id. The anti-flood property survives, because a genuine double-click has identical filters.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -258,23 +287,64 @@ function hedley_view_export_export_batch_worker($item) {
...
+    return $failure(format_string('No email address for user @uid, cannot deliver the export.', ['@uid' => $item->uid]));
+  }
+
   $params = [
     'export_uri' => $file->uri,
     'filename' => $file->filename,
   ];
   $mail_result = drupal_mail('hedley_view_export', 'export_ready', $account->mail, user_preferred_language($account), $params);
+
   if (empty($mail_result['result'])) {
```
anvmn, line 289: **medium — retrying the mail forever can spam duplicates and lock the user out of the export meanwhile.**

The queue is `retry after => 10` with `max attempts => 1000`, so a failing send is retried every ten seconds for roughly three hours. Two consequences:

- An empty `result` from `drupal_mail()` does not prove nothing was sent — a relay that accepts the message and then times out reports failure — so this can deliver the same link many times.
- The item stays QUEUED under the same title for the whole window, so the dedupe check added in this PR refuses every fresh request for that view, telling the user an export "is already running" that may never arrive.

Fix: bound the mail step specifically — count attempts in `$item->data` and fail permanently after a handful, rather than inheriting the 1000 meant for batches.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -168,7 +168,29 @@ function hedley_view_export_add_item(string $type, string $view_name, string $di
...
+  // message below was shown either way, so the second person to ask for an
+  // export was told it was on its way and then never received an email.
+  //
+  // Keeping the user and the display in the id preserves what the check was
+  // for - one person cannot flood the queue by asking twice for the same
+  // export - without letting unrelated exports block each other.
+  $unique_id = "{$user->uid}_{$view_name}_{$display}";
+  $queued = hedley_general_add_task_to_advanced_queue_by_id(HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE, $unique_id, $arguments);
+
+  if (!$queued) {
```
anvmn, line 183: **medium — `!$queued` conflates "already running" with "the insert failed".**

`hedley_general_add_task_to_advanced_queue()` returns the item id, and returns a falsy one when `createItem()` fails — having already shown its own "Failed to create AQ job" error. My branch then deletes the file and adds *"You already have an export of this view running. We will email a link when it is ready."*

Nothing is running, and no email will ever arrive. The user gets two contradictory messages, one of which is a promise that cannot be kept.

Fix: ask `hedley_general_advanced_queue_task_exists()` which case it is, and only claim a duplicate when there genuinely is one.

anvmn (commented):
Findings from `/code-review medium` on this branch. I verified all four against the contrib code before posting; my analysis is in the conversation and I have changed nothing yet.

The first one defeats the claim the second commit makes, so it is a blocker rather than a polish item.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -258,23 +311,110 @@ function hedley_view_export_export_batch_worker($item) {
...
+  if (empty($account->mail)) {
+    // No address to send to, and no retry will conjure one.
+    return $failure(format_string('No email address for user @uid, cannot deliver the export.', ['@uid' => $item->uid]));
+  }
+
   $params = [
     'export_uri' => $file->uri,
     'filename' => $file->filename,
   ];
   $mail_result = drupal_mail('hedley_view_export', 'export_ready', $account->mail, user_preferred_language($account), $params);
```
anvmn, line 288: **medium — the attempt cap does not cover the failure it was added for, and it is the same mistake I just fixed one line up.**

`mail_attempts` is incremented *after* `drupal_mail()` returns. If the mail system blocks past the 120s lease — an unresponsive SMTP relay, which is the case this whole change exists for — the worker never reaches the increment. `_advancedqueue_release_stale_items()` returns the item to QUEUED, the next claim takes the `export_built` early return, and `drupal_mail()` is called again with the counter unchanged.

So the hang path is bounded only by the queue's `max attempts` of 1000, at 10 seconds apart: hours of retries, the duplicate-check lock held throughout, and a relay that accepts DATA before timing out sends the link every time. Exactly what the commit message claims was eliminated.

Fix: increment and persist *before* calling `drupal_mail()`, for the same reason `export_built` is written first.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -168,7 +176,45 @@ function hedley_view_export_add_item(string $type, string $view_name, string $di
...
+
+  if (!$queued) {
+    // Nothing will process this file, so do not leave it behind: the cleanup
+    // cron only removes files a queue item points at.
+    file_delete($file);
+
+    // A refusal means one of two different things, and only one of them is
+    // worth promising an email for.
+    $task_title = HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE . '_' . $unique_id;
+    if (hedley_general_advanced_queue_task_exists(HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE, $task_title)) {
```
anvmn: **low — the refusal-reason recheck is a TOCTOU, and it becomes unnecessary once the branch below is fixed.**

`hedley_general_add_task_to_advanced_queue_by_id()` has already run this exact query. Re-running it can disagree: if the duplicate completed in between, its status is SUCCESS, which is not among the statuses checked, so a perfectly normal duplicate is reported with the red "could not start … tell an administrator" — while the user's export is in fact on its way. It flips the other way too, when a stuck PROCESSING item crosses the 20-minute threshold between the two calls.

See the next comment: once the insert failure is caught as an exception, a falsy return has only one possible meaning, and this second query can go entirely.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -168,7 +176,45 @@ function hedley_view_export_add_item(string $type, string $view_name, string $di
...
+  $queued = hedley_general_add_task_to_advanced_queue_by_id(HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE, $unique_id, $arguments);
+
+  if (!$queued) {
+    // Nothing will process this file, so do not leave it behind: the cleanup
+    // cron only removes files a queue item points at.
+    file_delete($file);
+
+    // A refusal means one of two different things, and only one of them is
+    // worth promising an email for.
+    $task_title = HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE . '_' . $unique_id;
```
anvmn: **low — this branch is unreachable, and the path it was written for leaks the file it was meant to clean up.**

`hedley_general_add_task_to_advanced_queue()` returns falsy only if `db_insert()->execute()` returns falsy, but in Drupal 7 a failing insert throws `PDOException`. I checked the chain: `createItem()` → `createUniqueItem()` → `db_insert`/`db_merge` execute.

So the real failure propagates out of `hedley_view_export_add_item()` **before** `file_delete($file)`, leaving an empty file in `private://hedley_view_export` that nothing points at — and `hedley_view_export_cron()` only deletes files reachable from an AQ row, so it stays there permanently.

Fix: wrap the queue call in try/catch and delete the file in the catch. That makes this branch do its job, and — with the exception path handled — a falsy return can only mean a duplicate, which removes the recheck above.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -258,23 +311,110 @@ function hedley_view_export_export_batch_worker($item) {
...
+ * @param RestfulAuthenticationManager $manager
+ *   The manager holding the original user, to switch back to.
+ * @param callable $failure
+ *   Returns a permanent AQ failure, for what a retry cannot mend.
+ *
+ * @return array
+ *   AQ result.
+ */
+function _hedley_view_export_notify_ready($item, $file, RestfulAuthenticationManager $manager, callable $failure) {
   $account = user_load($item->uid);
```
anvmn, line 283: **low — `user_load()` can return FALSE, and this reads a property off it.**

An account deleted between queueing and the final batch is plausible for a multi-hour export. `empty($account->mail)` on a bool is noise on every run before the correct permanent failure is returned. `if (!$account || empty($account->mail))` costs nothing.

(The pre-existing `$manager->setAccount($account)` at the top of the worker has the same exposure; out of scope for this diff, but worth knowing.)

anvmn (commented):
Second `/code-review medium` pass, after the first round of fixes. All four verified before posting.

Worth recording what it confirmed as well: the `unique_id` fix is real, the `export_built` early return is correctly placed after `file_load()`, persisting the flag before the mail attempt is right (`releaseItem()` writes only `status` and `expire`), `rows_count` is persisted before the retry reads it, `switchUserBack()` is called on every exit path, and the queue's `retry after => 10` does make `FAILURE_RETRY` requeue.

Note finding 2 and finding 3 have a single fix between them — see the comments.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -258,23 +316,112 @@ function hedley_view_export_export_batch_worker($item) {
...
+
+  if (!$account || empty($account->mail)) {
```
anvmn: **high — my `!$account` guard is in the wrong place, and the case it names kills the whole queue run.**

`RestfulAuthenticationManager::setAccount()` is declared `setAccount(\stdClass $account)`, and the worker calls it at line 193 with `user_load($item->uid)` — long before this guard at line 378. A deleted account makes `user_load()` return FALSE, so that call raises a **`TypeError`**.

`TypeError` is an `Error`, not an `Exception`, so `advancedqueue_process_item()`'s `catch (Exception $e)` does not catch it: the cron run dies mid-flight instead of failing this one item. Every other queued task behind it stops too.

So the scenario my own comment describes — a long export outliving the account that asked for it — still crashes, and my guard cannot be reached. It belongs at the top of the worker, before `setAccount()`. Note the crash is pre-existing; what is new is a comment claiming it is handled.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -168,7 +176,50 @@ function hedley_view_export_add_item(string $type, string $view_name, string $di
...
+    // otherwise leave the empty file behind for good: the cleanup cron only
+    // removes files a queue item points at, and there is no item.
+    file_delete($file);
+    watchdog_exception('hedley_view_export', $e);
+    drupal_set_message(t('Could not start the export. Please try again, and tell an administrator if it keeps happening.'), 'error');
+
+    return '<a href="javascript:history.back()">' . t('Go back') . '</a>';
+  }
+
+  if (!$queued) {
```
anvmn, line 183: **high — I was wrong last round: a falsy return does NOT have only one meaning, and the wrong branch makes a promise that will never be kept.**

`hedley_general_add_task_to_advanced_queue()` explicitly handles `createItem()` returning falsy — it sets its own "Failed to create AQ job" error and returns the falsy id. I reasoned that insert failures always throw, so falsy could only mean a duplicate. The defensive path exists regardless, and I cannot prove it unreachable.

If it is taken, the user now sees the red error AND "You already have this export running. We will email a link…", the file is deleted, and no email ever comes — precisely the silent-loss class this PR exists to remove.

Fix: check for the duplicate explicitly *before* adding, so the two cases are distinguished at the point where each is actually known, rather than inferred from a shared falsy value.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -258,23 +316,112 @@ function hedley_view_export_export_batch_worker($item) {
...
     'export_uri' => $file->uri,
     'filename' => $file->filename,
   ];
   $mail_result = drupal_mail('hedley_view_export', 'export_ready', $account->mail, user_preferred_language($account), $params);
+
   if (empty($mail_result['result'])) {
-    return $failure(format_string("Couldn't send export ready email to @email", ['@email' => $account->mail]));
+    $manager->switchUserBack();
+
+    return [
```
anvmn: **low — the retry reason I compose here is thrown away.**

On `FAILURE_RETRY`, `advancedqueue_process_item()` calls `requeueItem()`, which writes `expire`, `status`, `created` and `data` — not `result`. The old code returned a permanent FAILURE, and `deleteItem()` *does* persist `result`, so this is a regression in visibility: for the first five failures the AQ admin UI shows a bare "Retry" with no reason, and the reason only reappears on the final give-up.

Fix: `watchdog()` the reason where it is composed, rather than relying on a field the requeue does not write.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -258,23 +316,112 @@ function hedley_view_export_export_batch_worker($item) {
...
+function _hedley_view_export_notify_ready($item, $file, RestfulAuthenticationManager $manager, callable $failure) {
   $account = user_load($item->uid);
+
+  if (!$account || empty($account->mail)) {
+    // No address to send to, and no retry will conjure one. The account itself
+    // can be gone: a large export runs for hours.
+    return $failure(format_string('No email address for user @uid, cannot deliver the export.', ['@uid' => $item->uid]));
+  }
+
+  // Count this attempt BEFORE making it. Counting afterwards only catches a
```
anvmn: **low — the comment's premise is wrong.** `_advancedqueue_release_stale_items()` only flips the row back to QUEUED; it never kills the running worker. So "this process is killed when its lease expires" is not how AdvancedQueue behaves, and I should not have written it as though it were.

The consequence the reviewer draws — two workers inside `drupal_mail()` for the same item, both reading the same `mail_attempts`, last writer winning — is real in principle and would make the cap under-count. In practice production serialises queue processing (Jenkins concurrent builds are off, which is why B-058 part 2 and B-059 were closed as won't-fix), so it should not arise; but the comment should say what actually happens, and the counter should not claim a guarantee it does not give.

server/hedley/modules/custom/hedley_view_export/hedley_view_export.module
```diff
@@ -168,7 +176,50 @@ function hedley_view_export_add_item(string $type, string $view_name, string $di
...
+  $unique_id = "{$user->uid}_{$view_name}_{$display}_{$filters}";
+
+  try {
+    $queued = hedley_general_add_task_to_advanced_queue_by_id(HEDLEY_VIEW_EXPORT_EXPORT_VIEW_QUEUE, $unique_id, $arguments);
+  }
+  catch (Exception $e) {
+    // A failing insert throws rather than returning falsy, and the throw would
+    // otherwise leave the empty file behind for good: the cleanup cron only
+    // removes files a queue item points at, and there is no item.
+    file_delete($file);
```
anvmn: **low — both cleanup paths ignore what `file_delete()` returns.**

In D7 it returns FALSE, or a usage array, instead of deleting. Here the file has no AQ item pointing at it and `hedley_view_export_cron()` only walks files reachable from an AQ row, so a failed delete leaks the empty export permanently, with nothing that will ever collect it. The cron path at least warns via watchdog when a delete fails.

Fix: check the return and watchdog on failure, in both branches.

anvmn (commented):
Third `/code-review medium` pass. All five verified. Two of them correct conclusions I reached in the previous round, which is worth saying plainly.

It also re-confirmed the two core fixes: the per-export id removes the global `export_view_0` block, and `export_built` + persisting it before the mail attempt does prevent the last batch being appended twice.

PR #2031, opened 2026-08-04
Send backend error logs to GlitchTip instead of Rollbar

AronNovak:
Adds a `logs_glitchtip` module that reports watchdog events and uncaught exceptions over the Sentry protocol, replacing `logs_rollbar`, which an update hook disables (its code stays in the make file for now, since it implements hook_boot and removing it before every environment has run the update would break bootstrap). The DSN is deliberately not committed: set `GLITCHTIP_DSN` and `GLITCHTIP_ENVIRONMENT` in `server/config.sh`, or $conf['logs_glitchtip_dsn'] / `logs_glitchtip_environment` in settings.php — with neither set nothing is reported, so DDEV and CI stay silent. Pantheon environments are auto-detected; self-hosted country servers need the environment name set explicitly or reporting stays suppressed.

The Elm client still uses Rollbar and is untouched here — that is a separate follow-up.

### DIFF
--- a/server/default.config.sh
+++ b/server/default.config.sh
@@ -33,6 +33,15 @@ MYSQL_HOSTNAME="127.0.0.1"
 MYSQL_DB_NAME="drupal_elm_starter"
 
 
+# GlitchTip error reporting.
+# The DSN is an ingest key, so it is never committed. Leave it empty to keep
+# error reporting off, which is what local and CI environments want.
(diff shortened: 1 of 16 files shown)


anvmn (approved):
Copilot comments are on `logs_glitchtip module` which is generic, so, I don't think we should touch it.

Issue #2008, opened 2026-07-27
Child scoreboard: correct the existing kilogram-scaled birth weight records (#1981 part 2)


anvmn:
Follow-up to #1981, which had two parts. Part 1 — the gram range gate that stops new bad data — shipped in #1997. This is part 2: the records already stored.

## Why it still matters

The gate is forward-looking only. `hedley_ncda.module` computes `low_birth_weight = $birth_weight < 2000`, so every kilogram-scaled record already in the database keeps counting as a low birth weight, and any report reading that indicator stays unreliable until the stored values are dealt with.

Figures from the Rwanda site as measured in #1981 (14,048 birth weights total; Burundi had none):

| Range | Count | Status |
|---|---|---|
| under 10 | 1,817 | kilograms entered — mechanically recoverable (× 1,000) |
| 10–499 | 165 | implausible, not recoverable |
| 500–1,999 | 229 | genuinely low birth weight — leave alone |
| 2,000+ | 11,837 | fine |
| above 10,000 | 145 | implausible (max 350,022), not recoverable |

Of the 2,211 records currently flagged low birth weight, only 229 are genuine.

## What has to be decided

1. **The sub-10 group.** Multiplying by 1,000 is an assumption about what the nurse meant. It is a very safe assumption — the top values are 3.0, 2.8, 3.2, 3.5, 2.9, a textbook newborn distribution — but it is still a write to clinical data and wants sign-off before it runs.
2. **The 310 unrecoverable records** (165 in 10–499 and 145 above 10,000). There is no way to know what was intended. Options are to clear them, leave them, or flag them for a health-center review.
3. **Re-run the affected aggregates** after any correction, so `low_birth_weight` and the NCDA data reflect the corrected values.
4. **Re-measure first.** The counts above are from the #1981 investigation; take a fresh count before touching anything.

## Note carried over from #1997

Where a newborn exam already recorded the weight, the NCDA form does not ask for it — so an NCDA re-save can still carry a pre-existing bad value through. Correcting the stored records removes that path too.

## Related

- #1981 — the original report and the measurements above
- #1997 — part 1, the gram range gate
- #1980 — the wider missing-range-gate sweep

anvmn:
**Decision (@anvmn, 2026-08-05): convert.** The kilogram entries are to be multiplied into grams rather than rejected.

## What that covers, measured against a copy of the live database

| Band | Records | Treatment |
|---|---|---|
[...]

Issue #708, opened 2023-05-10
NCDA - Aggregated Scorecard: generate data for Demographics pane [4h]


anvmn:
![Selection_1841](https://github.com/TIP-Global-Health/eheza-app/assets/17882245/0d78aa80-df80-420c-9e52-7f605c487ac5)

anvmn:
### Analysis:

- For first 2 rows it's enought to pass child birth date.

- For birth weight, we need to:
   1. Load all (group, nutrition and well child) NCDA questioneers that child took.
   2. Load pregnancy summary for the child.

   Either one of NCDA questioneers or pregnancy summary should contain the birth weight.
   Note: low birth weight is considered to be bellow 2kg.

anvmn:
NCDA is on live for few months. Closing.

PR #2036, opened 2026-08-04
Child scoreboard: count low birth weight below 2500g, as the requirement says

### LINKED ISSUE #2035: Child scoreboard: aggregated scoreboard counts low birth weight below 2000g, but the requirement (and the app) say 2500g
## Summary

The aggregated NCDA scoreboard counts a child as born underweight below **2000 g**. The app uses **2500 g**, which is what the requirements specify and what WHO defines. Babies between 2000 g and 2499 g — most low-birth-weight babies — show "Born Underweight: Yes" on their own scorecard while being missing from the aggregated count.

This is live: `origin/main` carries the 2000.

## Where

* `server/hedley/modules/custom/hedley_ncda/hedley_ncda.module:530` — `$data['low_birth_weight'] = $birth_weight < 2000;`
* `client/src/elm/Pages/WellChild/ProgressReport/View.elm:1762,1783` — `birthWeight < 2500`
* `client/src/elm/Measurement/View.elm:3756` — `weight < 2500` puts the birth weight input in the red

Both sides resolve the weight identically: newborn-exam pregnancy summary `field_weight` first, else the first NCDA questionnaire's `field_birth_weight`. Only the threshold differs.

## Why 2500 is the specified number

* **#543** (*NCDA + Newborn Exam - Birthweight*, 2022-11-14): "The user will input the birthweight in grams. **If the weight is below 2500 grams the child is underweight.** There should be the RED callout next to it. In NCDA, this means that the scorecard will update accordingly."
* **#687** (*NCDA - Aggregated Scorecard - Demographics*, 2023-04-17) specifies the aggregated row as "Children born this month/year that had a **birth weight in the red**" — i.e. the same red callout, so the same 2500 g.

The 2000 has no requirement behind it. It entered in #708, the implementation task that replaced #687, via an analysis note reading "low birth weight is considered to be bellow 2kg", and no source upstream of that note says 2 kg. `2000` appears nowhere else in the codebase; `2500` is the threshold at all three client sites.

## Unreadable weights

The field is in grams, but some records hold kilograms (#1981, repair tracked in #2008). Values below 10 are kilograms typed into a grams field; values between 10 and 300 cannot be interpreted at all and cannot be repaired. All of them are currently counted as low birth weight, since they are below any threshold.

An unusable weight should read as *unknown*, not as *not underweight*. Both sides already model unknown: the aggregated decoder is `optional "low_birth_weight" (maybe bool) Nothing`, the aggregate counts only `Just True`, and the child's scorecard omits the "Born Underweight" row when the value is `Nothing`. Absence is also already a normal stored state, since a person with no recorded birth weight never had the key set.

## Fix

Raise the threshold to 2500 on the backend, and on both sides treat a birth weight below the plausible minimum (300 g, the same floor `getInputConstraintsBirthWeight` enforces at input) as unknown rather than as an answer.

## Deployment

`low_birth_weight` is **precomputed** into `person.field_ncda_data` by the Advanced Queue worker, and the scoreboard only reads that stored blob. The code change therefore has no effect on existing data until the aggregated data is regenerated per site:

```
drush scr profiles/hedley/modules/custom/hedley_ncda/scripts/generate-data-for-all.php
```

anvmn:
Fixes #2035

The aggregated scoreboard counted a child as born underweight below **2000 g**; the app uses **2500 g**. Babies of 2000-2499 g — most low-birth-weight babies — read "Born Underweight: Yes" on their own scorecard while being absent from the aggregated count.

2500 is the specified number, from the two issues that defined this indicator:

* **#543** (2022-11-14): "The user will input the birthweight in grams. **If the weight is below 2500 grams the child is underweight.** There should be the RED callout next to it. In NCDA, this means that the scorecard will update accordingly."
* **#687** (2023-04-17) specifies the aggregated row as "children born this month/year that had a **birth weight in the red**" — the same callout, so the same threshold.

The 2000 has no requirement behind it: it entered through #708, the implementation task that replaced #687, via an analysis note reading "low birth weight is considered to be bellow 2kg", with nothing upstream saying 2 kg.

## Unreadable weights now answer "unknown"

The field is in grams, but some records hold kilograms (#1981; repair tracked in #2008). Below 10 they are kilograms typed into a grams field; 10-300 cannot be interpreted at all and cannot be repaired. Every one of them used to count as low birth weight, being below any threshold.

They now read as unknown rather than as an answer, on both sides, using the same 300 g floor `getInputConstraintsBirthWeight` already enforces at input. Neither side needed a new state for this: the aggregated decoder is `optional "low_birth_weight" (maybe bool) Nothing`, the aggregate counts only `Just True`, the scorecard omits the row on `Nothing`, and a child with no recorded birth weight already produced exactly that.

**Visible change:** for those children the "Born Underweight" row disappears from the scorecard instead of reading "Yes". The aggregate keeps them in the newborns denominator (which is birth-date based) while excluding them from the numerator.

## One number, not three

`2500` was written out at three client sites and `2000` at one backend site. It is now `lowBirthWeightThreshold` in `Measurement.Utils` (used by the scorecard and by the red-callout check) and `HEDLEY_NCDA_LOW_BIRTH_WEIGHT_THRESHOLD` in `hedley_ncda`, each commented with the issues that set it and a pointer to the other side. That duplication is what let the two surfaces drift apart in the first place.

## Deployment

`low_birth_weight` is **precomputed** into `person.field_ncda_data` by the Advanced Queue worker, and the scoreboard only reads that stored blob — so this change reaches nothing already calculated until the data is recalculated.

`scripts/recalculate-low-birth-weight.php` does exactly that, and only for the children it applies to:

```
drush scr profiles/hedley/modules/custom/hedley_ncda/scripts/recalculate-low-birth-weight.php --dry_run
drush scr profiles/hedley/modules/custom/hedley_ncda/scripts/recalculate-low-birth-weight.php
```

Two groups calculate to something different: children between the old threshold and the new one (2000 to 2499 g), who were not counted as born underweight and now are, and children whose weight is too small to be read, who were counted and now count as unknown. Everyone else calculates to what is already stored.

That is why this is not `generate-data-for-all.php`. Recalculating a person ends in `node_save()`, which writes a revision, and device sync selects on `node.vid > base_revision` (`HedleyRestfulSync.class.php:142`) — so a full run would re-download the entire person table to every device in the field to deliver a change that applies to a few hundred of them.

The dry run reports what each affected child currently reads, without touching anything. Worth running first on Test, where the numbers can be checked against the app.

## Verification

* `elm make src/elm/Main.elm` — Success, 548 modules
* `elm-test` — 2968 passed
* `elm-format --validate` — clean
* `elm-review`, cold (`elm-stuff` removed first) — I found no errors!
* `php -l` — clean; `phpcs` Drupal and DrupalPractice with the full CI extension list — clean

**No behavioural test, deliberately.** The client logic is a `let` binding inside `viewChildIdentificationPane`, so it is not reachable from a unit test without extracting it — the same testability ceiling recorded for the birth-length gate in #2003. On the backend, nothing in the repo tests `hedley_ncda` aggregation; a guard would mean a new SimpleTest class standing up a person, a pregnancy summary and an aggregation run — a full install (~+2 to +9 min on the simpletest job) to protect a threshold constant. Agreed with @anvmn to accept lint plus review here. The 300 floor itself is already pinned by `birthWeightOutsideConstraintsTest`, which asserts 299 outside and 300 inside.


server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
```diff
@@ -526,8 +535,13 @@ function hedley_ncda_calculate_aggregated_data_for_person($person, $birth_date_f
...
     }
   }
-  if ($birth_weight) {
-    $data['low_birth_weight'] = $birth_weight < 2000;
+  // Birth weight is recorded in grams. A weight below the plausible minimum
+  // cannot be read: the smallest ones are kilograms typed into a grams field
+  // (issue #2008), and the rest cannot be interpreted at all. Leaving the
+  // indicator unset marks it as unknown, rather than claiming the child was
+  // not born underweight.
+  if ($birth_weight >= HEDLEY_NCDA_MINIMAL_PLAUSIBLE_BIRTH_WEIGHT) {
```
anvmn, line 550: **Finding 1 — no backfill for already-computed scorecard data** (medium/high, `/code-review medium`)

`low_birth_weight` is persisted per person in `field_ncda_data`, and recomputation only happens from `hedley_ncda_node_insert` / `hedley_ncda_node_update`. This branch adds no `hook_update_N` and no `.install` file (the module has none). A child born at 2300 g whose record was computed before this deploy stays `low_birth_weight: false` until someone touches one of their measurement nodes, while a child measured after the deploy gets `true` — so the scoreboard would report a mix of the 2000 and 2500 thresholds indefinitely.

**Assessment: valid, and already handled outside the code.** The regeneration requirement is documented in the PR body under “Deployment” and in #2035: `drush scr profiles/hedley/modules/custom/hedley_ncda/scripts/generate-data-for-all.php`, per site. The fair part of the finding is that a PR body is a poor home for something that has to happen at release time a month later — a comment on the constant naming that script would be durable. The suggested `hook_update_N` that enqueues every person is declined: this project does bulk recomputation through drush scripts, and queueing the whole person table is a much larger commitment than a threshold change warrants.

server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
```diff
@@ -526,8 +535,13 @@ function hedley_ncda_calculate_aggregated_data_for_person($person, $birth_date_f
...
+    $data['low_birth_weight'] = $birth_weight < HEDLEY_NCDA_LOW_BIRTH_WEIGHT_THRESHOLD;
```
anvmn, line 551: **Finding 2 — backend and client disagree on the fallback when the newborn-exam weight is unusable** (medium, `/code-review medium`)

(Anchored here; the line it is really about is the `if (empty($birth_weight))` fallback above, which is outside this diff.)

The client now returns `Nothing` for an implausible pregnancy-summary weight, and `bornUnderweighEntry` falls through via `Maybe.Extra.or` to `bornUnderweightByNCDA`. The backend does not: its NCDA fallback is gated on `empty($birth_weight)`, and a value like `3.2` is not empty, so `$birth_weight` stays `3.2`, the fallback never runs, and the indicator is left unset.

Scenario: newborn exam records `3.2` (kilograms typed into the grams field), the questionnaire records `2400`. The child’s scorecard says “Born Underweight: Yes”; the aggregated scorecard treats the child as unknown and never counts them.

**Assessment: confirmed, and it is a regression introduced by this PR** — sharper than the finding states. Before this change the two sides agreed (client `Just (3.2 < 2500)` = Yes, backend `3.2 < 2000` = true). The change split them, in exactly the dimension the PR sets out to fix.

Fix: gate the fallback on usability rather than emptiness, `if ($birth_weight < HEDLEY_NCDA_MINIMAL_PLAUSIBLE_BIRTH_WEIGHT)` — `NULL` becomes `0` there, so a missing weight still falls through. Checked the follow-on case: when the questionnaire weight is also unusable, both sides land on unknown.

client/src/elm/Pages/WellChild/ProgressReport/View.elm
```diff
@@ -1759,9 +1761,21 @@ viewChildIdentificationPane language site allNCDAQuestionnaires db ( childId, ch
             viewEntry Translate.ChildName child.name
 
         bornUnderweightByNewbornExam =
-            Maybe.andThen (\value -> Maybe.map (\(WeightInGrm birthWeight) -> birthWeight < 2500) value.birthWeight)
+            Maybe.andThen (\value -> Maybe.andThen resolveBornUnderweight value.birthWeight)
                 newbornExamPregnancySummary
 
+        resolveBornUnderweight (WeightInGrm birthWeight) =
+            if birthWeight < getInputConstraintsBirthWeight.minVal then
```
anvmn: **Finding 3 — the plausibility gate is one-sided** (low/medium, `/code-review medium`)

`getInputConstraintsBirthWeight` is `{ minVal = 300, maxVal = 7000 }`, and this gate uses only `minVal`. A stored `35000` passes both `birthWeight < 300 == False` here and `$birth_weight >= 300` on the backend, so both sides assert “not born underweight” and the child counts as a known-good newborn in the scoreboard denominator. The finding argues the rationale in the new comments — an uninterpretable weight should read as unknown, not as “not underweight” — applies equally above `maxVal`.

**Assessment: the supporting mechanism is refuted; the residual concern is narrower than stated.** The finding rests on out-of-range weights reaching the database because the check is “a warning popup, not a block”. It is a block on both entry paths:

- `Pages/WellChild/Activity/Update.elm:211-219` dispatches the save only when `outOfRange` is empty — *“Nothing is saved until they are within range.”*
- `Measurement/View.elm:2541-2557` diverts the NCDA save button to the popup instead of saving, deliberately on every step, because the weight is asked on the first step and saved on the last.

The cited evidence (`Measurement/Test.elm` exercising `WeightInGrm 350022`) tests the predicate, not what is persisted.

What survives: historical rows written before the range gate ships can hold anything, and such a row would read as “not underweight” on both sides. Both sides agree, so it creates no divergence. Floor-only was an explicit decision by @anvmn; declining, now with the added fact that new out-of-range entries are impossible from the release onward.

client/src/elm/Pages/WellChild/ProgressReport/View.elm
```diff
@@ -1759,9 +1761,21 @@ viewChildIdentificationPane language site allNCDAQuestionnaires db ( childId, ch
             viewEntry Translate.ChildName child.name
 
         bornUnderweightByNewbornExam =
-            Maybe.andThen (\value -> Maybe.map (\(WeightInGrm birthWeight) -> birthWeight < 2500) value.birthWeight)
+            Maybe.andThen (\value -> Maybe.andThen resolveBornUnderweight value.birthWeight)
                 newbornExamPregnancySummary
 
+        resolveBornUnderweight (WeightInGrm birthWeight) =
```
anvmn: **Finding 4 — no test covers the new branch** (low, `/code-review medium`)

(Anchored to the helper; the finding names `Pages/WellChild/ProgressReport/Test.elm`, which is not in this diff. That file only describes `generateUniversalInterventionsValues`.)

`resolveBornUnderweight` and the PHP gate are both untested, and the two `300`s are duplicated across languages with nothing tying them together. Since finding 2 shows the two implementations drifted within this very commit, a small test on each side would be cheap insurance.

**Assessment: the argument is fair, the remedy is not proportionate — declining, per @anvmn’s decision when this was scoped.** This helper is a `let` binding inside `viewChildIdentificationPane`, so testing it means first extracting it to `Measurement.Utils` — about ten lines plus a test, and it would guard only the client half. The backend half has no test infrastructure at all: nothing in the repo covers `hedley_ncda` aggregation, so a guard means a new SimpleTest class standing up a person, a questionnaire and an aggregation run, at the cost of a full install (~+2 to +9 min on the simpletest job) to protect a threshold constant.

Fixing finding 2 removes the drift this finding is arguing from, which is the more direct answer. The 300 floor is also already pinned indirectly by `birthWeightOutsideConstraintsTest`, which asserts 299 outside and 300 inside.

server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
anvmn: **Correction to my assessment above, and fixed in 14eac8d16.**

I called this "a regression introduced by this PR" that loses children from the count. That overstates it: the divergence is real in the code, but the data does not reach it.

@anvmn pointed out that the birth weight is taken only once, and the code agrees — `showNCDAQuestionsByNewbornExam` (`Measurement/Utils.elm:466-475`) is `Maybe.map (.birthWeight >> isNothing) newbornExamPregnancySummary |> Maybe.withDefault True`, and `Measurement/View.elm:3196` renders the birth weight input only when it is `True`. The questionnaire therefore asks for a weight **only when the newborn exam has none**, so the two sources are mutually exclusive by construction. My scenario needed both present at once, which the normal flow does not produce: if the newborn exam recorded `3.2`, the questionnaire never asked.

In the reachable cases the two sides already agreed — with a newborn exam weight, neither falls back; without one, `empty($birth_weight)` is true and the client is `Nothing`, so both do. The one path that could still produce both weights is a questionnaire filled before a newborn exam is completed, since the gate is evaluated when the questionnaire is filled.

So this is **latent, not live**. Fixed anyway in 14eac8d16, because the two implementations should state the same rule — that is what this PR is for — and the change costs one condition with no effect on reachable data.

Also void, from the same fact: my note about the backend picking the first non-empty `$ncdas` entry while the client takes `List.head`. With only one questionnaire ever carrying a weight, both can only pick the same one.

server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
anvmn: Addressed in 14eac8d16 to the extent it belongs in code: the threshold constant now says that changing it leaves already calculated data behind and that `scripts/generate-data-for-all.php` has to run afterwards on every site — a more durable home than a PR body that will be a month old by release time.

The `hook_update_N` half stays declined: bulk recomputation here goes through drush scripts, and queueing the whole person table is a larger commitment than a threshold change warrants.

client/src/elm/Measurement/View.elm
```diff
@@ -3753,7 +3753,7 @@ birthWeightInputsAndTasks language birthWeight setBirthWeightMsg =
         colorAlertIndication =
             Maybe.map
                 (\weight ->
-                    if weight < 2500 then
+                    if weight < lowBirthWeightThreshold then
```
anvmn: Why don't we check for lower limit here?

anvmn: You are right — it was inconsistent with the rest of the change. Fixed in 609d1698.

I had scoped the floor to the indicator (the two scorecard computations and the backend) and treated this as an input affordance where the range gate already speaks, since a weight below 300 cannot be saved. That is too thin: the callout does not say “something is off”, it asserts the child was born underweight — the same claim the scorecard makes. And it is reachable outside typing, because a legacy record holding `3.2` hydrates into the form and shows RED while the save gate simultaneously refuses the value as implausible.

Rather than add a third copy of the check, the rule now lives in one place. `Measurement.Utils.bornUnderweightByBirthWeight : Maybe WeightInGrm -> Maybe Bool` answers “was this child born underweight, as far as the weight recorded can say”, with `Nothing` for a weight that cannot be read, and it is asked by the callout and by both halves of the scorecard. The `let` binding in the scorecard is what left this site behind, so it is gone.

That also made the logic testable, which is the cheap version of the “no test covers the new branch” finding — `bornUnderweightByBirthWeightTest` in `Measurement/Test.elm` pins both ends (2499/2500 and 299/300, plus `3.2` and no weight at all). Discrimination checked: dropping the floor from the helper fails exactly the “a weight too small to read is not an answer” case, and nothing else.

Still floor only, no upper bound, as you decided. With the rule in one function that is now a one-line change in one place if you ever want it.

elm make 548 modules, elm-test 2974 passed, elm-format clean, elm-review cold clean.

server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
```diff
@@ -9,6 +9,17 @@ include_once 'hedley_ncda.features.inc';
 
 define('HEDLEY_NCDA_CALCULATE_AGGREGATED_DATA', 'hedley_ncda_calculate_ncda_data');
 
+// Birth weight, in grams, below which a child counts as born underweight.
+// This is what puts the birth weight input in the red on the client, and the
+// aggregated scorecard counts the children it applies to - issues #543, #687.
+// Changing this leaves already calculated data behind, so run
+// scripts/generate-data-for-all.php afterwards, on every site.
```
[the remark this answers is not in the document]
anvmn: Fixed in 9b62f9530. The comment was written before the targeted script existed and never updated — repointed at `recalculate-low-birth-weight.php`, and phrased so it says what has to happen (recalculate the children the threshold moves) rather than naming one script as the answer forever.

server/hedley/modules/custom/hedley_ncda/scripts/recalculate-low-birth-weight.php
```diff
@@ -0,0 +1,145 @@
...
+        ->condition('f.' . $source['column'], $previous_threshold, '>=')
+        ->condition('f.' . $source['column'], HEDLEY_NCDA_LOW_BIRTH_WEIGHT_THRESHOLD, '<')
+    );
+
+  $query
+    ->fields('fp', ['field_person_target_id'])
+    ->condition('f.deleted', 0)
+    ->condition('n.status', NODE_PUBLISHED)
+    ->condition('n.type', 'person')
+    ->condition($bands);
```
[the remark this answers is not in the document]
anvmn: Fixed in 9b62f9530 — the join now matches on `entity_type` as well as `entity_id`, and both `f.entity_type = 'node'` and `fp.deleted = 0` are named.

Worth recording what it changes in practice: **nothing, on real data.** With a copy of the live database in place I ran the query both ways and it selects the same **2,700** children. So this is defence against a future where these field names appear on another entity type, not a correction.

server/hedley/modules/custom/hedley_ncda/scripts/recalculate-low-birth-weight.php
```diff
@@ -0,0 +1,145 @@
...
+        $decoded = json_decode($items[0]['value'], TRUE);
+        if (!isset($decoded['low_birth_weight'])) {
+          $stored = 'unknown';
+        }
+        else {
+          $stored = $decoded['low_birth_weight'] ? 'born underweight' : 'not born underweight';
+        }
+      }
+
+      drush_print("  Person $person->nid currently reads: $stored");
```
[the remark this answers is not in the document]
anvmn: Not fixed, because it works — PHP's simple double-quoted syntax interpolates a single level of property access. Run on this codebase's PHP 7.4:

```
$person->nid = 4242;  $stored = "unknown";
"  Person $person->nid currently reads: $stored"
→ "  Person 4242 currently reads: unknown"
```

No error, correct value. Braces would be a style preference rather than a fix, so I have left the line as it is.

The script has since been run end to end against a copy of the live database, and this line printed 2,700 person ids correctly.

server/hedley/modules/custom/hedley_ncda/scripts/recalculate-low-birth-weight.php
```diff
@@ -0,0 +1,145 @@
...
+      $stored = 'not calculated yet';
+      $items = field_get_items('node', $person, 'field_ncda_data');
+      if ($items) {
+        $decoded = json_decode($items[0]['value'], TRUE);
+        if (!isset($decoded['low_birth_weight'])) {
+          $stored = 'unknown';
+        }
+        else {
+          $stored = $decoded['low_birth_weight'] ? 'born underweight' : 'not born underweight';
+        }
```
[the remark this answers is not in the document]
anvmn: Not fixed — `isset()` is precisely the construct that does not emit diagnostics for a bad offset, which is why the code reads the key through it. Every shape `json_decode` can return, on PHP 7.4 (what the sites run):

| decoded | `isset($decoded['low_birth_weight'])` | warning |
|---|---|---|
| `NULL` (invalid JSON) | not set | none |
| `5` | not set | none |
| `"str"` | not set | none |
| `[1,2]` | not set | none |
| `{"low_birth_weight":true}` | set | none |

An `is_array()` guard would be inert here.

PR #2041, opened 2026-08-05
Child scoreboard: convert birth weights that were recorded in kilograms

### LINKED ISSUE #2008: Child scoreboard: correct the existing kilogram-scaled birth weight records (#1981 part 2)

anvmn:
Fixes #2008

Stacked on #2036 — **base is `b149-low-birth-weight-threshold`, retarget to develop once that merges.** The recalculation this performs has to run with the corrected threshold and the plausibility floor already in place, or it would write answers that are wrong in a different way.

The birth weight is asked for and stored in grams, and a share of the records hold kilograms — a `3.4` where `3400` was meant. Those read far below any threshold, so the child counts as born underweight on their own scorecard and in the aggregated one, from a number that says nothing about them.

## Measured against a copy of the live database

| Band | Records | Treatment |
|---|---|---|
| `0 < v < 10` — kilograms | **2,029** | **2,013 converted**, `× 1000` |
| of those, converting above 7,000 g | **16** | **Left alone**, and named in the output |
| `10 <= v < 300` — between the scales | 71 | Left alone, permanently |

Range of the convertible values is 0.3 to 9.8 kg; none converts below the 300 g floor.

**The 16 are not birth weights in kilograms either.** They run from 7.1 to 9.8, and all but two are `child_scoreboard_ncda` — they read like the weight the child had on the day, entered in the wrong field. 7,000 g is what the app itself accepts on entry, so converting them would write values the input gate would refuse. The script prints their node ids so someone can decide what they actually were.

## Why it recalculates, rather than leaving that to the other script

Converting changes which children the indicator applies to. Once a `3.4` becomes `3400`, that child no longer matches any band `recalculate-low-birth-weight.php` looks for — while their stored answer is precisely the one the conversion just invalidated. Run the conversion first and the other script second, and those ~2,000 children are silently skipped.

So this recalculates each child **as soon as their weight is converted**, rather than collecting them for a second pass. A child is then either untouched or wholly done, and since a converted weight no longer matches what this script looks for either, **a run that stops early can simply be run again** — no bookkeeping, no order to remember. The deduplication a second pass would have bought is worth fourteen recalculations out of two thousand: 2,013 conversions belong to 1,999 distinct children.

It saves through `node_save` rather than writing the field tables, so the corrected weights reach devices; and it recalculates directly rather than relying on the queue, because B-153 records ~1,317 children whose stored data the current code already disagrees with. A data repair should not be built on a mechanism with evidence of losing work.

Side effect worth knowing: each `node_save` also enqueues the ordinary NCDA recalculation for that person, so the queue will redo work this script has already done. Harmless, and deduplicated per person.

## Verification

`php -l` clean; `phpcs` Drupal and DrupalPractice clean.

**Dry run against a copy of the live database**: reports 2,013 to convert and lists the 16 it holds back. 2,013 + 16 = 2,029, matching the row counts measured independently in SQL.

The conversion itself has not been run yet — that is the next step, on the local copy, before it goes near a real site.

## Running it

```
drush scr profiles/hedley/modules/custom/hedley_ncda/scripts/convert-kilogram-birth-weights.php --dry_run
drush scr profiles/hedley/modules/custom/hedley_ncda/scripts/convert-kilogram-birth-weights.php
```

Then `recalculate-low-birth-weight.php` for the children whose weight was already in grams. Order does not matter.

### DIFF
--- a/server/hedley/modules/custom/hedley_ncda/scripts/convert-kilogram-birth-weights.php
+++ b/server/hedley/modules/custom/hedley_ncda/scripts/convert-kilogram-birth-weights.php
@@ -1,0 +1,185 @@
+<?php
+
+/**
+ * @file
+ * Converts birth weights that were recorded in kilograms into grams.
+ *
+ * The birth weight is asked for and stored in grams, but a share of the
+ * records hold the weight in kilograms - a 3.4 where 3400 was meant. They read
+ * as far below any threshold, so the child counts as born underweight on their
+ * scorecard and in the aggregated one, while the weight itself is nonsense.
+ *
+ * Only weights that can be read as kilograms are converted. Two groups are
+ * left as they are:
+ * - Weights that would convert to more than a newborn can weigh. They are not
+ *   kilograms of birth weight either, and are most likely the weight the child
+ *   had on the day, entered in the wrong field. The script names them so they
+ *   can be looked at.
+ * - Weights between the two scales, which are neither grams nor kilograms and
+ *   which nothing says how to read.
+ *
+ * Converting changes which children the aggregated indicator applies to, so
+ * this recalculates every child it converts a weight for. That leaves it
+ * independent of recalculate-low-birth-weight.php, which covers the children
+ * whose weight was already in grams.
+ *
+ * Execution: drush scr profiles/hedley/modules/custom/hedley_ncda/scripts/
+ *   convert-kilogram-birth-weights.php [--dry_run] [--batch=50].
+ */
+
+if (!drupal_is_cli()) {
+  // Prevent execution from browser.
+  return;
+}
+
+// Report what would be converted, without converting it.
+$dry_run = (bool) drush_get_option('dry_run', FALSE);
+
+// Get the number of nodes to be processed at once.
+$batch = drush_get_option('batch', 50);
+
+// Get allowed memory limit.
+$memory_limit = drush_get_option('memory_limit', 800);
+
+// A weight below this can only be kilograms: it is far under what the lightest
+// newborn on record weighs in grams.
+$kilograms_ceiling = 10;
+
+// The heaviest a newborn can be, in grams. Mirrors
+// getInputConstraintsBirthWeight.maxVal on the client, which refuses anything
+// above it on entry.
+$grams_ceiling = 7000;
+
+// Birth weight is recorded at the newborn exam, and at the NCDA questionnaire
+// for children who did not have one.
+$sources = [
+  [
+    'field' => 'field_weight',
+    'table' => 'field_data_field_weight',
+    'column' => 'field_weight_value',
+    // The field carries every weight measurement, so it is the bundle that
+    // says which rows are birth weights.
+    'bundle' => HEDLEY_ACTIVITY_WELL_CHILD_PREGNANCY_SUMMARY_CONTENT_TYPE,
+  ],
+  [
+    'field' => 'field_birth_weight',
+    'table' => 'field_data_field_birth_weight',
+    'column' => 'field_birth_weight_value',
+    // The field exists only on the NCDA questionnaires.
+    'bundle' => NULL,
+  ],
+];
+
+$convert = [];
+$skipped = [];
+foreach ($sources as $source) {
+  $query = db_select($source['table'], 'f');
+  $query->join('node', 'n', 'n.nid = f.entity_id');
+  hedley_general_apply_exclude_deleted($query, 'n');
+
+  $query
+    ->fields('f', ['entity_id', $source['column']])
+    ->condition('f.deleted', 0)
+    ->condition('f.entity_type', 'node')
+    ->condition('n.status', NODE_PUBLISHED)
+    ->condition('f.' . $source['column'], 0, '>')
+    ->condition('f.' . $source['column'], $kilograms_ceiling, '<');
+
+  if (!empty($source['bundle'])) {
+    $query->condition('f.bundle', $source['bundle']);
+  }
+
+  foreach ($query->execute() as $row) {
+    $grams = $row->{$source['column']} * 1000;
+    $item = [
+      'nid' => $row->entity_id,
+      'field' => $source['field'],
+      'from' => $row->{$source['column']},
+      'to' => $grams,
+    ];
+
+    if ($grams > $grams_ceiling) {
+      $skipped[] = $item;
+      continue;
+    }
+
+    $convert[] = $item;
+  }
+}
+
+if (empty($convert) && empty($skipped)) {
+  drush_print('No birth weight is recorded in kilograms. Nothing to convert.');
+  return;
+}
+
+drush_print(count($convert) . ' birth weights are recorded in kilograms.');
+
+if (!empty($skipped)) {
+  drush_print(count($skipped) . ' more would convert to more than a newborn can weigh, and are left as they are:');
+  foreach ($skipped as $item) {
+    drush_print("  Measurement {$item['nid']}: {$item['from']} would become {$item['to']}");
+  }
+}
+
+if ($dry_run) {
+  drush_print('Dry run. Nothing was converted.');
+  return;
+}
+
+// Convert, remembering whose aggregated data the conversion invalidates.
+$persons = [];
+$converted = 0;
+foreach (array_chunk($convert, $batch) as $chunk) {
+  $nodes = node_load_multiple(array_column($chunk, 'nid'));
+  foreach ($chunk as $item) {
+    if (empty($nodes[$item['nid']])) {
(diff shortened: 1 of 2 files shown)


server/hedley/modules/custom/hedley_ncda/scripts/convert-kilogram-birth-weights.php
```diff
@@ -0,0 +1,225 @@
...
+foreach ($orphans as $item) {
+  $converted += (int) _hedley_ncda_convert_birth_weight($item);
+}
+
+foreach (array_chunk($convert, $batch, TRUE) as $chunk) {
+  foreach ($chunk as $person_id => $items) {
+    foreach ($items as $item) {
+      $converted += (int) _hedley_ncda_convert_birth_weight($item);
+    }
+
```
anvmn, line 200: **Finding 3 — "one recalculation covers them all" held only under an unguaranteed row order** (low, `/code-review medium`)

For a child with two kilogram-scaled questionnaires, the recalculation ran at the first conversion while the second was still in kilograms — and `hedley_ncda_calculate_aggregated_data_for_person()` takes the first non-empty `field_birth_weight` and breaks, with no `ORDER BY` tying this script's order to the module's. So the blob could be computed from half-converted data, and the `isset()` skip meant it was never recomputed.

**Assessment: confirmed. Scope: 14 children** — 2,013 conversions belong to 1,999 distinct children, measured against a copy of the live database. The review notes the queued job would paper over it; that is the mechanism B-153 documents losing work, so it is not something to rely on.

Fixed in 9f2efa4b3aad9c727f5da8420a6c28c77e11abcf by grouping the conversions by child: all of a child's weights are converted, then they are recalculated once. Resumability is unaffected, since a converted weight still matches nothing this script looks for.

server/hedley/modules/custom/hedley_ncda/scripts/convert-kilogram-birth-weights.php
```diff
@@ -0,0 +1,225 @@
...
+}
+
+if (!empty($too_heavy)) {
+  drush_print(count($too_heavy) . ' weights would convert to more than a newborn can weigh, and are left as they are:');
+  foreach ($too_heavy as $item) {
+    drush_print("  Measurement {$item['nid']}: {$item['from']} would become {$item['to']}");
+  }
+}
+
+if (!empty($unreadable)) {
```
anvmn, line 155: **Finding 6 — the unreadable middle band was dropped from all output** (low, `/code-review medium`)

The docstring said two groups are left alone and that the script names them, but the query stopped below the kilogram ceiling, so weights between the two scales were never counted or printed.

**Assessment: confirmed.** Fixed in 9f2efa4b3aad9c727f5da8420a6c28c77e11abcf — the selection now runs to the plausibility floor and the run reports three groups. Against a copy of the live database: **2,013** convertible weights across 1,999 children, **16** that would convert above what a newborn can weigh, **71** between the scales. Each number matches an independent SQL count.

server/hedley/modules/custom/hedley_ncda/scripts/recalculate-low-birth-weight.php
```diff
@@ -131,9 +153,10 @@
 
   $memory = round(memory_get_usage() / 1048576);
   if ($memory >= $memory_limit) {
-    drush_print(dt('Stopped before running out of memory, after @total of @count children. Run again to continue.', [
+    drush_print(dt('Stopped before running out of memory, after @total of @count children. Run again with --nid=@nid to continue.', [
       '@total' => $total,
       '@count' => $count,
```
anvmn, line 158: **Finding 1 — "Run again to continue" could not make progress** (medium, `/code-review medium`)

Recalculating a child does not change their birth weight, so a re-run reselected the same sorted list and stopped in the same place: the tail would never be reached, while the message kept saying it would. `generate-data-for-all.php` solves this with `--nid`; this had no equivalent.

**Assessment: confirmed, and the most important of the six.** Fixed in 9f2efa4b3aad9c727f5da8420a6c28c77e11abcf — the script takes `--nid`, filters the selection above it, and this message now names the id to pass.

Also fixed here (finding 4a): the query applied the deleted and published checks to the person only, never to the measurement carrying the weight, which the conversion script did do. Measured against a copy of the live database, adding it changes nothing — 2,700 children either way — so it is consistency rather than a correction.

server/hedley/modules/custom/hedley_ncda/scripts/convert-kilogram-birth-weights.php
```diff
@@ -0,0 +1,225 @@
...
+ * this recalculates each child once all of their weights are converted. That
+ * leaves it independent of recalculate-low-birth-weight.php, which covers the
+ * children whose weight was already in grams, and it means anything the script
+ * has finished is finished: a converted weight no longer matches what the
+ * script looks for, so a run that stops early can simply be run again.
+ *
+ * Note that saving a measurement queues the ordinary recalculation for its
+ * child as well, so a run leaves one queued job per child behind, repeating
+ * work this script has already done. They are harmless, but they occupy the
+ * queue for a while - the run says how many.
```
anvmn, line 31: **Finding 2 — each `node_save()` also queues a redundant recalculation** (medium, `/code-review medium`)

`hedley_ncda_node_update()` enqueues a job per child, deduplicated while queued, so a run leaves ~1,999 jobs repeating work this script has already done — behind a serialized worker, that occupies the queue for a long time and delays real sync work. Neither the docstring nor the output said so.

**Assessment: confirmed as an omission.** It was stated in the PR description but not where an operator would look. Both the docstring and the closing output now say it, with the number.

Not doing what the finding stops short of suggesting: deleting those queue rows afterwards. They are redundant rather than harmful, and deleting queue state carries more risk than the waste it would save.

anvmn:
**Finding 5 — `showNCDAQuestionsByNewbornExam` still uses `isJust`** (low, `/code-review medium`)

Posted here rather than inline: it points at `client/src/elm/Measurement/Utils.elm`, which belongs to the base branch and is not in this diff.

The claim: with unreadable weights now answering "unknown", a newborn exam holding `3.2` still counts as answered, so the NCDA form does not re-ask, and the child is permanently unknown with no in-app way for a nurse to supply a readable weight.

**Assessment: the mechanism is exactly right, the reach is two records — declining.** Of everything the conversion leaves behind, only **2** sit on pregnancy summaries; the other 69 between the scales and all 16 above 7,000 g are on NCDA questionnaires, which stay editable. Measured against a copy of the live database.

Changing a form gate to reach two records is not proportionate, and #2007's range gate stops new unreadable weights from being entered at all, so the group cannot grow.

**Same call on the second half of finding 4:** the recalculation script also selects children whose resolved birth weight comes from the other source, so a few are recalculated for nothing. The honest fix is comparing before saving in `hedley_ncda_calculate_aggregated_data_for_person()` — that would make every over-selection free, and it is worth doing on its own terms rather than by complicating this query.

PR #2045, opened 2026-08-06
Child scoreboard: keep one value per month by the month it was taken in

### LINKED ISSUE #2044: Child scoreboard: scoreboard drops measurements: the one-per-month rule buckets by months-before-today, not by month
## Summary

The aggregated NCDA scoreboard decoder keeps "one value per calendar month" by dropping the dates into a dictionary — but keys them on **how many months ago the measurement was**, not on the month it was taken in. The boundary therefore falls on the day of the month rather than the 1st, and the rule it is implementing is broken in both directions at once.

Live: present in `origin/main`.

## The code, and what it says it does

`server/elm/src/Backend/Scoreboard/Decoder.elm`. The docblock:

> Guiding rule is that patient should have only one severity value during calendar month. In case there are multiple severities, most severe one needs to be selected.

and the key, repeated at four sites:

```elm
List.map (\date -> ( diffMonths date currentDate, date )) dates |> Dict.fromList
```

With `currentDate = 2026-08-06`:

| Measurement | Calendar month | Bucket |
|---|---|---|
| 2026-07-02 | July | 1 |
| 2026-07-20 | July | 0 |
| 2026-08-01 | August | 0 |

## What it costs

**Measurements are silently dropped.** Any pair where the earlier date falls late in its month and the later one early in the next shares a bucket, and `Dict.fromList` keeps one. A child measured once a month loses half of them:

| Input | Kept today | Should be kept |
|---|---|---|
| May 28, Jun 3, Jul 20, Aug 1 | Jun 3, Aug 1 | all four |

The view matches by calendar month (`equalByYearAndMonth`), so the dropped ones are simply absent from their column — **the columns undercount**.

**And the stated rule is violated.** Two dates in the same month can land in different buckets, so both survive. In the severities case the cross-severity filters compare on that same key, so a child with a severe reading on July 2 and a normal one on July 25 is counted as **both severe and normal for July**.

Because the bucket depends on today's date, which values collide changes as the month advances — the same instability as #2033, in a different file.

## Reach

`sanitizeSingleValuePerMonth` is applied to every monthly row — ANC-Newborn, Universal Intervention rows 2 to 5, Nutrition Behaviour 2 to 4, Targeted Interventions 1 to 6, and all five WASH rows. `sainitzeCriterionBySeverities` covers stunting, underweight, wasting and MUAC, in three severities each. So most panes of the scoreboard are affected.

## Fix

Key on the month the measurement was taken in, which is what the comments already claim. The `currentDate` those functions take exists only to compute the broken key, so it goes with it — and with it the parameter threading that carried the date down from `updateBackend`, leaving the scoreboard's decoder and update functions shaped like their Reports and Completion siblings, which never took a date.

No recalculation is needed: this is decode-time arithmetic over the stored data.

anvmn:
Fixes #2044

The decoder's stated rule is one value per calendar month. It kept one value per *months-before-today* instead, so the boundary fell on the day of the month rather than the 1st — and the rule broke in both directions at once.

## What changes

The four keying sites move to the month the measurement was taken in:

```elm
List.map (\date -> ( calendarMonth date, date ))
```

Simulated against the real semantics of `Date.diff Months` and `AssocList.fromList`, with `currentDate = 2026-08-06`:

| Input | Kept before | Kept after |
|---|---|---|
| Jul 20, Aug 1 | **Aug 1 only** | both |
| Jul 2, Jul 25 | **both** | Jul 25 |
| May 28, Jun 3, Jul 20, Aug 1 | **Jun 3, Aug 1** | all four |

The third row is the everyday case: a child measured once a month loses half of those measurements, because each pair where the earlier date sits late in its month and the later one early in the next collapses into one bucket. The view matches by calendar month, so what is dropped is simply missing from its column.

The second row is the other half — two readings in one July surviving as separate values. In the severities case the cross-severity filters compare on the same key, so a child with a severe reading on July 2 and a normal one on July 25 was counted as **both severe and normal for July**, which is precisely what the rule exists to prevent. The simulation confirms that stops.

## Why the diff reaches further than four lines

`currentDate` existed in these functions only to compute the broken key. With the key gone the parameter is unused, and `NoUnused.Parameters` follows it up the chain: `decodeMonthlyValues` → the five pane decoders → `decodeNCDAData` → `decodePatientData` → `decodeSyncResponse` → `Backend.Scoreboard.Update.update` → `updateBackend`.

I removed it rather than silencing it with `_`, because the end state is the one that reads correctly: `Backend.Reports.Decoder.decodeSyncResponse` and `Backend.Completion.Decoder.decodeSyncResponse` never took a date, and neither did their `update` functions. The scoreboard was the outlier, and only because of this bug. The rest of the diff is mechanical and the compiler plus `elm-review` confirm it is complete.

## Deployment

**No recalculation needed.** This is decode-time arithmetic over the stored data, so it takes effect as soon as the rebuilt bundle is deployed — unlike #2035, which changes what is stored.

## Verification

* `elm make src/Main.elm` — Success, 84 modules
* `elm-format --validate` — clean
* `elm-review`, cold — I found no errors! (it drove the parameter removal)
* Simulation of old versus new bucketing, ported from `Date.diff Months` and `AssocList.fromList` semantics — the table above

**No test.** `server/elm` has no test suite at all: no test dependencies in `elm.json`, no `Test` modules, and CI runs only `elm-review` on it (`ci-scripts/test_elm_review.sh:27`). Stated rather than implied.

The regenerated `elm-main.js` is the second commit. Since #2034 and #2038 have merged, this bundle carries only this change plus the compiler-version churn already in develop's copy.

### DIFF
--- a/server/elm/src/Backend/Scoreboard/Decoder.elm
+++ b/server/elm/src/Backend/Scoreboard/Decoder.elm
@@ -5,9 +5,10 @@ import Backend.Components.Decoder exposing (decodeReportParams, decodeSelectedEn
 import Backend.Decoder exposing (decodeSite)
 import Backend.Scoreboard.Model exposing (ANCNewbornData, CriterionBySeverities, InfrastructureEnvironmentWashData, NCDAData, NutritionBehaviorData, NutritionCriterionsData, PatientData, RawVaccinationData, ScoreboardData, SyncResponse, TargetedInterventionsData, UniversalInterventionData, VaccinationProgressDict, VaccineType(..), emptyANCNewbornData, emptyInfrastructureEnvironmentWashData, emptyNCDAData, emptyNutritionBehaviorData, emptyNutritionCriterionsData, emptyTargetedInterventionsData, emptyUniversalInterventionData)
 import Backend.Scoreboard.Utils exposing (generateVaccinationProgressForVaccine)
+import Date
 import EverySet exposing (EverySet)
 import Gizra.Json exposing (decodeInt)
-import Gizra.NominalDate exposing (NominalDate, decodeYYYYMMDD, diffMonths)
+import Gizra.NominalDate exposing (NominalDate, decodeYYYYMMDD)
 import Json.Decode exposing (Decoder, bool, field, list, map, maybe, string, succeed)
 import Json.Decode.Pipeline exposing (hardcoded, optional, required)
 import Maybe.Extra exposing (isNothing)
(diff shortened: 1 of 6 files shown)


server/elm/src/Backend/Scoreboard/Decoder.elm
```diff
@@ -24,70 +25,71 @@ decodeScoreboardData =
...
+        |> map sanitizeCriterionBySeverities
 
 
 {-| Guiding rule is that patient should have only one severity value
 during calendar month.
 In case there are multuple severities, most severe one needs to be selected.
 -}
-sainitzeCriterionBySeverities : NominalDate -> CriterionBySeverities -> CriterionBySeverities
-sainitzeCriterionBySeverities currentDate data =
+calendarMonth : NominalDate -> ( Int, Int )
```
anvmn: **Finding 1 — the doc comment now documents the wrong function** (low, `/code-review medium`)

`{-| Guiding rule is that patient should have only one severity value during calendar month… -}` describes `sanitizeCriterionBySeverities`, but `calendarMonth` was inserted between the comment and it, so a rule about severity precedence now documents a two-field tuple helper.

**Assessment: confirmed, and mine** — the helper was added by a scripted edit that put it directly above the signature it was meant to precede, without noticing the signature carried a docblock. Moving `calendarMonth` above the comment restores it.

server/elm/src/Backend/Scoreboard/Decoder.elm
```diff
@@ -24,70 +25,71 @@ decodeScoreboardData =
...
+calendarMonth date =
+    ( Date.year date, Date.monthNumber date )
```
anvmn: **Finding 2 — the same notion is now defined twice** (low, `/code-review medium`)

`calendarMonth` re-implements what `Utils.NominalDate.equalByYearAndMonth` (`Utils/NominalDate.elm:11-14`) already expresses, in a different module.

**Assessment: confirmed, and it matters more here than the severity suggests.** This bug *was* the decoder's bucketing disagreeing with the view's matching, and the fix as written adds a second definition that happens to agree. Putting `calendarMonth` in `Utils/NominalDate.elm` and defining

```elm
equalByYearAndMonth first second =
    calendarMonth first == calendarMonth second
```

makes that agreement structural rather than coincidental — the decoder then keys on exactly what the view compares with. Behaviour is unchanged, since `Date.monthNumber` and `Date.month` are in bijection.

(The `Utils/NominalDate.elm` half is not in this diff, which is why this is anchored on the helper it replaces.)

anvmn:
**Review verification worth recording** (`/code-review medium`)

The review rebuilt `elm-main.js` from this branch's source with elm 0.19.2 and **byte-compared it against the committed bundle — identical**. That is the check the compiled-artifact rule actually needs: it proves the committed bundle is what this source produces, rather than taking the author's word that it was regenerated. I am folding it into my own routine for server-Elm changes.

It also confirmed the branch contains all of `origin/develop`, so the regeneration cannot regress recently merged work — relevant here, since #2034 and #2038 both rebuilt this same file.

No correctness findings. Every consumer of the affected lists was checked (`Pages/Scoreboard/View.elm` lines 288, 383, 478, 576, 779, 901, 1075): all filter with `equalByYearAndMonth` and none depends on list order or on which date within a month survives, so the key-type change carries no ordering regression.

server/elm/src/Backend/Scoreboard/Decoder.elm
anvmn: Fixed in 0deb3c4df3ec1e553d7605d26d70ed5e4e78ecd8 — `calendarMonth` is gone from this module entirely (see the reply on finding 2), so the comment sits directly above `sanitizeCriterionBySeverities` again.

server/elm/src/Backend/Scoreboard/Decoder.elm
anvmn: Fixed in 0deb3c4df3ec1e553d7605d26d70ed5e4e78ecd8, taking the structural version you proposed.

`calendarMonth` now lives in `Utils/NominalDate.elm` next to the comparison, and `equalByYearAndMonth` is defined in terms of it:

```elm
calendarMonth date =
    ( Date.year date, Date.monthNumber date )

equalByYearAndMonth first second =
    calendarMonth first == calendarMonth second
```

The decoder now keys on precisely what the view compares with, so the two cannot come apart — which is the invariant this bug broke in the first place. The decoder imports it and no longer needs `Date` at all.

Re-verified: elm make Success, elm-format clean, elm-review cold clean, bundle regenerated **and byte-compared against a fresh build of the committed source — identical**. That last check is yours: I had been asserting regeneration rather than proving it, and it is part of my routine for server-Elm changes now.

PR #2047, opened 2026-08-06
Vaccination: schedule vaccines the way the client does

### LINKED ISSUE #2046: Vaccination: server copy of the vaccination engine has drifted: every Rwandan child 18-24 months reads off track
## Summary

`server/elm/src/Pages/Scoreboard/Utils.elm` holds a copy of the client's vaccination scheduling engine, and it has drifted in three ways. Together they make the scoreboard's "up to date with immunization" row, and the same rows on the Reports postnatal-care page, wrong in both directions.

Live: present in `origin/main`.

## (a) DTP-standalone is scheduled for every site

`allVaccineTypes` takes no site and always includes `VaccineDTPStandalone`. Only Burundi schedules it.

The consumer (`Pages/Scoreboard/View.elm:617-641`) takes the **earliest outstanding dose** and marks the child off track unless it falls after the reference month. DTP-standalone is due at the later of 28 days after OPV-3 and 18 months of age — inside the scoreboard's 24-month window — and a Rwandan child never receives it. So from 18 months on it is permanently the earliest outstanding dose: **every Rwandan child aged 18 to 24 months with an OPV-3 record reads as off track, whatever they actually received.**

Children with no OPV-3 recorded are unaffected: the fallback schedules it 999 years out.

This affects the scoreboard only. `Pages/Reports/View.elm` passes the site-gated list from `Pages/Reports/Utils.elm`, which is already correct.

## (b) Rwanda's second IPV dose is missing

`getLastDoseForVaccine` takes no site and caps IPV at the first dose. Since #1426 Rwanda gives two. A child who has had IPV-1 is treated as finished, so children genuinely behind on IPV-2 are counted **up to date** — the opposite direction to (a), which is part of why the row looks plausible.

**The obvious fix for this is dangerous.** `getIntervalForVaccine` returns `( 0, Days )` for IPV in both copies. On the client that is a labelled placeholder:

> So far, there was only single IPV dose. Since #1426, at Rwanda site, we got second dose scheduled on the latter between age of 36 weeks, and 4 weeks after first dose was administered. This requirement is not reflected here. Instead, it's defined as special case at appropriate spots in code.

The server copied the placeholder without the comment and without the special cases. Adding the site to `getLastDoseForVaccine` alone would therefore schedule IPV-2 at `lastDoseDate + 0 days` — due the day dose 1 was given — putting **every** Rwandan child with an IPV-1 record off track. That is worse than the bug.

## (c) The OPV second-dose rule is missing

The client schedules OPV-2, where a birth dose was given, at the later of six weeks of age and four weeks after the birth dose. The server adds a flat four weeks, so it expects the dose about two weeks early and flags children behind before they are due.

## Fix, in an order that matters

1. `allVaccineTypes` takes `Site`: DTP-standalone for Burundi only, and still no HPV — this report stops at 24 months, which is what makes it legitimately different from the Reports list.
2. Thread `birthDate` into `nextVaccinationDataForVaccine` and port both special cases there.
3. Port the Rwanda IPV branch into `initialVaccinationDateByBirthDate`, for children with no IPV record yet.
4. Only then give `getLastDoseForVaccine` its site parameter — it is the switch that turns the second dose on, and is safe only once 2 and 3 exist.
5. Carry the client's comment across, so the placeholder is labelled in both copies.

No recalculation is needed: this is view-time arithmetic over the stored data.

## Not included

The DTP-standalone gate reads OPV-3 rather than DTP-3 in both the client and this copy. That is #1978's territory and needs a clinical decision, not a port.

anvmn:
Fixes #2046

The scoreboard runs its own copy of the client's vaccination engine, and it has drifted in three ways that pull the immunisation row in **both directions at once** — which is why the number looks plausible while being wrong per child.

## What was wrong

**DTP-standalone was scheduled for every site.** The consumer (`Pages/Scoreboard/View.elm:617-641`) takes the *earliest outstanding dose* and marks the child off track unless it falls after the reference month. DTP-standalone is due at the later of 28 days after OPV-3 and 18 months of age — inside the 24-month window — and Rwanda never gives it. So **every Rwandan child aged 18 to 24 months with an OPV-3 record read as off track**, whatever they had actually received. Children without OPV-3 were unaffected; their fallback is 999 years out.

**Rwanda's second IPV dose was missing**, so a child who had IPV-1 was treated as finished — the opposite error.

**OPV-2 used a flat four-week interval**, where the client waits for the later of six weeks of age and four weeks after the birth dose, so children read behind about two weeks early.

## ⚠️ Why the order of these changes matters

`getIntervalForVaccine` returns `( 0, Days )` for IPV **in both copies**. On the client that is a placeholder, and it says so:

> This requirement is not reflected here. Instead, it's defined as special case at appropriate spots in code.

The server had copied the placeholder without the comment and without the special cases. So the obvious one-line fix — give `getLastDoseForVaccine` a site so Rwanda expects a second dose — would have scheduled IPV-2 at `lastDoseDate + 0 days`, **due the day the first dose was given**, putting every Rwandan child with an IPV-1 record off track. Strictly worse than the bug it fixes.

The dose count is therefore turned on **last**, after the special cases and the initial date exist. The client's comment is carried across so the placeholder is labelled in both copies.

## The two vaccine lists stay separate

`Pages/Reports/Utils.allVaccineTypes` includes `VaccineHPV`; this one does not, because this report covers children to 24 months and HPV is given at 12 years — the docblock said so already. They are not two copies of one list, and unifying them would put a 12-year vaccine into an under-2 report. The scoreboard's copy gains only the site gate, and the docblock now says why they differ.

## Verification

Simulated old against new for a child born 2025-01-01:

**Rwanda IPV dose 2**, showing what the naive fix would have done:

| IPV-1 given | Naive port | This PR |
|---|---|---|
| 2025-04-09 (14w) | **2025-04-09** — same day | 2025-09-10 |
| 2025-07-30 (30w) | **2025-07-30** — same day | 2025-09-10 |
| 2025-10-08 (40w) | **2025-10-08** — same day | 2025-11-05 |

**OPV dose 2 with a birth dose:**

| OPV-1 given | Before | After |
|---|---|---|
| 2025-01-02 | 2025-01-30 | 2025-02-12 |
| 2025-01-11 | 2025-02-08 | 2025-02-12 |

**`allVaccineTypes`:** Rwanda loses `dtp_sa`, Burundi keeps it, neither gains HPV.

Also: `elm make` Success · `elm-format --validate` clean · `elm-review` cold — I found no errors! · and the regenerated bundle **byte-compared against a fresh build of the committed source — identical**.

**No test.** `server/elm` has no test suite: no test dependencies, no `Test` modules, and CI runs only `elm-review` on it.

## Not included, and related

The DTP-standalone gate reads OPV-3 rather than DTP-3, in this copy and the client's alike, though the comment beside it says "All 3 doses of DTP" and the PHP engine checks `dtp dose-3`. It has no GitHub issue — it is carried in the improvement backlog as B-055 — and it needs a clinical decision rather than a port, since it changes when a dose is considered due. Fixing it here alone would put this copy out of step with the client again.

**The PHP completion engine has the same missing IPV dose, and the same trap** (`hedley_reports.module`, `'+0 days'` for ipv, an OPV special case, no IPV one). It is a separate PR, and it should follow this one so it can cite the rule rather than re-derive it. Its own wrinkle: the next-dose block is duplicated at two call sites there, so the rule has to be extracted before it is added, or it lands twice.

### DIFF
--- a/server/elm/src/Pages/Scoreboard/Utils.elm
+++ b/server/elm/src/Pages/Scoreboard/Utils.elm
@@ -30,7 +30,7 @@ generateFutureVaccinationsData site birthDate vaccinationProgress =
                 nextVaccinationData =
                     case latestVaccinationDataForVaccine vaccinationProgress vaccineType of
                         Just ( lastDoseAdministered, lastDoseDate ) ->
-                            nextVaccinationDataForVaccine site vaccineType initialOpvAdministered lastDoseDate lastDoseAdministered
+                            nextVaccinationDataForVaccine site birthDate vaccineType initialOpvAdministered lastDoseDate lastDoseAdministered
 
                         Nothing ->
                             let
@@ -71,9 +71,9 @@ latestVaccinationDataForVaccine vaccinationsData vaccineType =
             )
 
 
-nextVaccinationDataForVaccine : Site -> VaccineType -> Bool -> NominalDate -> VaccineDose -> Maybe ( VaccineDose, NominalDate )
-nextVaccinationDataForVaccine site vaccineType initialOpvAdministered lastDoseDate lastDoseAdministered =
-    if getLastDoseForVaccine initialOpvAdministered vaccineType == lastDoseAdministered then
+nextVaccinationDataForVaccine : Site -> NominalDate -> VaccineType -> Bool -> NominalDate -> VaccineDose -> Maybe ( VaccineDose, NominalDate )
+nextVaccinationDataForVaccine site birthDate vaccineType initialOpvAdministered lastDoseDate lastDoseAdministered =
+    if getLastDoseForVaccine site initialOpvAdministered vaccineType == lastDoseAdministered then
         Nothing
 
     else
@@ -83,13 +83,39 @@ nextVaccinationDataForVaccine site vaccineType initialOpvAdministered lastDoseDa
                     let
                         ( interval, unit ) =
                             getIntervalForVaccine site vaccineType
+
+                        byInterval =
+                            Date.add unit interval lastDoseDate
                     in
-                    ( dose, Date.add unit interval lastDoseDate )
+                    if vaccineType == VaccineOPV && initialOpvAdministered && dose == VaccineDoseSecond then
+                        -- The initial OPV dose is given before the child is two
+                        -- weeks old, and the second at six weeks, so the interval
+                        -- alone does not say when it is due.
+                        ( dose, laterOf byInterval (Date.add Weeks 6 birthDate) )
+
+                    else if site == SiteRwanda && vaccineType == VaccineIPV && dose == VaccineDoseSecond then
+                        -- The interval for IPV is zero, since for most sites
+                        -- there is only one dose. Rwanda's second dose is due on
+                        -- the later of 36 weeks of age and 28 days after the
+                        -- first, which is what says when it is due here.
+                        ( dose, laterOf (Date.add Days 28 lastDoseDate) (Date.add Weeks 36 birthDate) )
+
+                    else
+                        ( dose, byInterval )
                 )
 
 
-getLastDoseForVaccine : Bool -> VaccineType -> VaccineDose
-getLastDoseForVaccine initialOpvAdministered vaccineType =
+laterOf : NominalDate -> NominalDate -> NominalDate
+laterOf first second =
+    if Date.compare first second == GT then
+        first
+
+    else
+        second
+
+
+getLastDoseForVaccine : Site -> Bool -> VaccineType -> VaccineDose
+getLastDoseForVaccine site initialOpvAdministered vaccineType =
     case vaccineType of
         VaccineBCG ->
             VaccineDoseFirst
@@ -114,7 +140,12 @@ getLastDoseForVaccine initialOpvAdministered vaccineType =
             VaccineDoseSecond
 
         VaccineIPV ->
-            VaccineDoseFirst
+            case site of
+                SiteRwanda ->
+                    VaccineDoseSecond
+
+                _ ->
+                    VaccineDoseFirst
 
         VaccineMR ->
             VaccineDoseSecond
@@ -163,6 +194,12 @@ getIntervalForVaccine site vaccineType =
         VaccineRotarix ->
             ( 4, Weeks )
 
+        -- So far, there was only single IPV dose.
+        -- Since https://github.com/TIP-Global-Health/eheza-app/issues/1426,
+        -- at Rwanda site, we got second dose scheduled on the latter
+        -- between age of 36 weeks, and 4 weeks after first dose was administered.
+        -- This requirement is not reflected here. Instead, it's defined as
+        -- special case at appropriate spots in code (which use getIntervalForVaccine).
         VaccineIPV ->
             ( 0, Days )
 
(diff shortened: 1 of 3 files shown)


server/elm/src/Pages/Scoreboard/Utils.elm
```diff
@@ -245,8 +282,19 @@ initialVaccinationDateByBirthDate site birthDate initialOpvAdministered vaccinat
...
 
         VaccineIPV ->
-            Date.add Weeks 14 birthDate
-                |> Date.add unit (dosesInterval * interval)
+            case ( site, vaccineDose ) of
+                ( SiteRwanda, VaccineDoseSecond ) ->
+                    -- The later of 36 weeks of age and 28 days after the first
+                    -- dose. Where the first dose is not known, 36 weeks is the
+                    -- earliest the second is allowed.
+                    Dict.get VaccineIPV vaccinationProgress
```
anvmn, line 298: **Finding 3 — the DTP-standalone gate reads OPV-3, not DTP-3** (low, pre-existing, `/code-review medium`)

(Anchored on the nearest line in this diff; the branch it concerns is the `VaccineDTPStandalone` case a little above, which this PR does not touch.)

That branch looks up `VaccineOPV / VaccineDoseThird`, while its own comment says *"All 3 dosed of DTP were given"* and the PHP engine checks `dtp dose-3` (`hedley_reports.module:5779`) — a later, independent implementation matching the comment's intent.

**Assessment: real, and the two scenarios are worth recording.**

* A Burundi child with OPV 1-3 but only two DTP doses gets DTP-standalone scheduled at 18 months and reads off track from that month, though it is not yet due.
* A Burundi child with all three DTP doses but no third OPV dose falls into the `Maybe.withDefault (Date.add Years 999 birthDate)` branch and reads on track indefinitely.

**One correction to the framing.** The finding says this PR makes the lookup load-bearing. For Burundi it was equally determinative before — DTP-standalone was already in their list and already scheduled through this branch. What changes is that Rwanda no longer reaches this code at all, so the PR removes an effect rather than creating one.

**Declining here, deliberately.** The same lookup exists in the client, so correcting this copy alone would re-introduce the drift this PR removes. And it changes when a dose is considered due, which is a clinical decision rather than a port. It is carried in the improvement backlog as B-055, with the PHP engine as the evidence for the intent; these two scenarios are being added to that entry.

anvmn:
**Findings 1 and 2 — the PHP copy still has the same drift, and the same trap** (medium and low, `/code-review medium`)

Posted here rather than inline: both concern `server/hedley/modules/custom/hedley_reports/hedley_reports.module`, which this PR does not touch.

**Both confirmed, and both are the next PR rather than a gap in this one.** They are carried in the improvement backlog as B-141 and were scoped before this PR was written; the "Not included, and related" section above names them.

Finding 1 is exact: `hedley_reports_get_last_dose_for_vaccine()` returns `'dose-1'` for ipv with no `$site`, so at `:5817` the loop `continue`s and `well_child_ipv_immunisation` is never expected — completion is over-reported for every Rwanda child from 36 weeks who is missing dose 2. Independently reaching the same conclusion from the other side is useful corroboration.

Finding 2 is the trap, and the reason the PHP fix is sequenced after this PR rather than bundled into it: `hedley_reports_get_interval_for_vaccine()` returns `'+0 days'` for ipv and the special-case block at `:5828-5835` covers only OPV, so correcting the dose count alone would expect IPV-2 from the day after dose 1. Exactly the failure this PR was ordered to avoid in Elm.

One thing the findings do not mention, which makes the PHP change bigger than it looks: **that next-dose block is duplicated.** `hedley_reports_get_last_dose_for_vaccine()` is called at `:5815` and `:6463`, each followed by its own byte-identical copy of the interval, OPV special case and `$expected[]` logic. The rule has to be extracted into one function before it is added, or the fix creates two more copies of it — taking this single rule to six implementations.

That is recorded in the backlog entry, so the PHP PR starts from it rather than rediscovering it.

PR #2051, opened 2026-08-06
Vaccination: wait for three doses of DTP, not three of polio

### LINKED ISSUE #2050: Vaccination: DTP standalone booster waits for three doses of polio instead of three of DTP
## Summary

The DTP standalone booster is due after three doses of DTP. The Elm code waits for three doses of **polio** instead — a different vaccine, which in Burundi has four doses of its own.

Burundi only: it is the only site that schedules this vaccine.

## The requirement says DTP

The code came in with `29a578fde` ("Logic for suggesting DTP (standalone vaccine)"), from PR #927, for issue **#926**:

> In addition there is an extra (4th dose of DTP only) at 18 months
>
> **Add new vaccine called DTP for one dose at 18 months. It should only appear after 3 doses of Pentavalent and with a minimum 28 day interval from the last Pentavalent dose.**

**#920** supplies the naming and the schedule: in Burundi the DTP-HepB-Hib combo is called **Pentavalent** — what the code calls `VaccineDTP` — and its table gives Pentavalent 3 doses while Oral Polio has 4. The two are not interchangeable.

Everything except the code agrees:

| Source | Says |
|---|---|
| Requirement #926 | after 3 doses of Pentavalent, 28 days from the last one |
| The comment above the code | "All 3 doses of DTP were given… 28 days since third dose" |
| The variable it binds | `fourWeeksAfterThirdDTPDose` |
| The PHP engine (`hedley_reports.module:5779`) | `$vaccination_progress['dtp']['dose-3']` |
| **The Elm code** | **`VaccineOPV`, dose 3** |

## What it costs

Children vaccinated on schedule are unaffected — the 18-month floor decides the date, so the lookup does not change the answer. It bites catch-up children:

* **Three DTP doses but polio incomplete** → the lookup finds nothing, the fallback is "never", and the booster is **never due**. The child is scored on track indefinitely and the dose is never offered.
* **Polio dose 3 earlier than DTP dose 3** → the booster is offered **inside** the 28 day window after the third DTP dose that the requirement asks to wait for.

## Fix

`VaccineOPV` becomes `VaccineDTP`, in both Elm copies — `client/src/elm/Measurement/Utils.elm` and `server/elm/src/Pages/Scoreboard/Utils.elm`. They must move together, or the scoreboard and the app disagree again, which is what #2046 has just finished repairing.

The PHP engine already reads `dtp` and does not change.

anvmn:
Fixes #2050

Stacked on #2047 — **base is `b131-server-vaccination-engine-drift`, retarget to develop once that merges.** It touches the same function, and this is the correction that PR's review surfaced for the third time.

The DTP standalone booster is due after three doses of DTP. The code waited for three doses of **polio** — a different vaccine, which in Burundi has four doses of its own.

## This was gated on a clinician, and should not have been

It sat in the improvement backlog from Round 1 as needing clinical sign-off, on the grounds that reading polio might be a deliberate proxy. It is not a judgement call: the requirement is written down.

`29a578fde` ("Logic for suggesting DTP (standalone vaccine)") came from PR #927, for issue **#926**:

> **Add new vaccine called DTP for one dose at 18 months. It should only appear after 3 doses of Pentavalent and with a minimum 28 day interval from the last Pentavalent dose.**

**#920** gives the naming — Burundi calls the DTP-HepB-Hib combo **Pentavalent**, which is `VaccineDTP` — and its schedule table gives Pentavalent 3 doses against Oral Polio's 4. They are not interchangeable, and @anvmn's comment on #926 confirms the mapping at the time.

Four sources agree; the code was alone:

| Source | Says |
|---|---|
| Requirement #926 | after 3 doses of Pentavalent |
| The comment above the code | "All 3 doses of DTP were given" |
| The variable it binds | `fourWeeksAfterThirdDTPDose` |
| The PHP engine (`hedley_reports.module:5779`) | `$vaccination_progress['dtp']['dose-3']` |
| The Elm code | `VaccineOPV`, dose 3 |

## Who it affects

Burundi only, and only children catching up. On schedule, 18 months of age decides the date and the lookup makes no difference.

* **Three DTP doses, polio incomplete** → the lookup finds nothing, the fallback is "never", the booster is never due. The child reads on track for a dose nobody will offer.
* **Polio dose 3 earlier than DTP dose 3** → the booster is offered inside the 28-day window the requirement asks to wait out.

## Both copies, together

`client/src/elm/Measurement/Utils.elm` and `server/elm/src/Pages/Scoreboard/Utils.elm`. Moving one alone would re-create the client/scoreboard drift #2046 has just finished repairing. The PHP already reads `dtp` and does not change.

## Verification

`elm make src/elm/Main.elm` — Success, 548 modules · `elm-test` — 2974 passed · `elm make src/Main.elm` (server) — Success, 84 · `elm-format --validate` both files — clean · `elm-review` cold, both apps — I found no errors! · regenerated bundle **byte-compared against a fresh build of the committed source — identical**.

No new test: the change is a one-word lookup with the intent stated in the comment beside it, and `server/elm` has no test suite at all.

## Which reports move

Both consumers of `generateFutureVaccinationsData`, not only the scoreboard:

* **Aggregated scoreboard**, Universal Intervention row 1 (`Pages/Scoreboard/View.elm:619`).
* **Reports, "up to date with immunization"** (`Pages/Reports/View.elm:3127`), whose buckets run to 24 months — so its `19To24Months` row moves for Burundi with this change, and its `15WeeksTo10Months`, `10To19Months` and `19To24Months` rows move for Rwanda with #2047's IPV change.

Worth comparing both reports before and after, rather than the scoreboard alone.

### DIFF
--- a/client/src/elm/Measurement/Utils.elm
+++ b/client/src/elm/Measurement/Utils.elm
@@ -9536,7 +9536,10 @@ initialVaccinationDateByBirthDate site birthDate initialOpvAdministered vaccinat
             -- All 3 dosed of DTP were given, it has passed
             -- at least 28 days since third dose, and, child
             -- is at last 18 months old.
-            Dict.get VaccineOPV vaccinationProgress
+            -- Burundi calls the DTP combo Pentavalent, and issue #926 asks for
+            -- this dose after three of them - not after three doses of polio,
+            -- which is a different vaccine with a fourth dose of its own.
+            Dict.get VaccineDTP vaccinationProgress
                 |> Maybe.andThen (Dict.get VaccineDoseThird)
                 |> Maybe.map
                     (\thirdDoseDate ->
--- a/server/elm/src/Pages/Scoreboard/Utils.elm
+++ b/server/elm/src/Pages/Scoreboard/Utils.elm
@@ -252,7 +252,10 @@ initialVaccinationDateByBirthDate site birthDate initialOpvAdministered vaccinat
             -- All 3 dosed of DTP were given, it has passed
             -- at least 28 days since third dose, and, child
             -- is at last 18 months old.
-            Dict.get VaccineOPV vaccinationProgress
+            -- Burundi calls the DTP combo Pentavalent, and issue #926 asks for
+            -- this dose after three of them - not after three doses of polio,
+            -- which is a different vaccine with a fourth dose of its own.
+            Dict.get VaccineDTP vaccinationProgress
                 |> Maybe.andThen (Dict.get VaccineDoseThird)
                 |> Maybe.map
                     (\thirdDoseDate ->
--- a/server/hedley/modules/custom/hedley_general/js/elm-main.js
+++ b/server/hedley/modules/custom/hedley_general/js/elm-main.js
@@ -21178,7 +21178,7 @@ var $author$project$Pages$Scoreboard$Utils$initialVaccinationDateByBirthDate = F
 						A2(
 							$elm$core$Maybe$andThen,
 							$pzp1997$assoc_list$AssocList$get($author$project$Backend$Scoreboard$Model$VaccineDoseThird),
-							A2($pzp1997$assoc_list$AssocList$get, $author$project$Backend$Scoreboard$Model$VaccineOPV, vaccinationProgress))));
+							A2($pzp1997$assoc_list$AssocList$get, $author$project$Backend$Scoreboard$Model$VaccineDTP, vaccinationProgress))));
 			case 'VaccinePCV13':
 				return A3(
 					$justinmimbs$date$Date$add,


anvmn:
## Review triage (`/code-review medium`)

All four findings concern the base branch (#2047), not this one-word change, so they are here rather than inline — the lines are outside this diff.

**Finding 1 — the PHP copy is left behind on Rwanda IPV. Already done, in a sibling PR.** The review diffed `develop...b055`, which contains #2047 but not #2049, where that fix lives: `hedley_reports_resolve_expected_immunisations()` extracts the duplicated next-dose block and adds the Rwanda IPV dose-2 rule, and `hedley_reports_get_last_dose_for_vaccine()` gains `$site`. The scenario described — a Rwanda child at 10 months with one IPV dose reading off track in Elm but complete in PHP — is exactly what #2049 closes. Both need to merge for the three copies to agree.

**Finding 2 — the Reports page is a second consumer, unmentioned. Fair, and added to the PR body.** `Pages/Reports/View.elm:3127` passes `Pages.Reports.Utils.allVaccineTypes site` into the same `generateFutureVaccinationsData`, and its buckets run to 24 months, so its `19To24Months` row moves for Burundi with this change and three of its rows move for Rwanda with #2047. QA should compare both reports, not the scoreboard alone.

**Finding 3 — a child with more doses than the schedule expects can never read up to date. Confirmed, and reachable today.** Counted on a copy of the live Rwanda database, children whose recorded dose count exceeds the schedule: **DTP 12, PCV13 11, OPV 10, IPV 3, MR 2** — about 38 children reading off track every month with no way to clear it, and it only accumulates. Worst for IPV, whose `( 0, Days )` interval puts the phantom dose on the day the last one was given.

Not fixed here: unrelated to this change, pre-existing in both Elm copies, and the guard has to land in both together or it re-creates the drift #2046 repaired. Filed as **B-154** in the improvement backlog with those numbers and your suggested guard.

**Finding 4 — my rationale comment pointed at the wrong codebase. Fixed in 4f1e6a2b4**, on #2047 where the code lives. I had written that the branch is kept "because the client asks this same question for any dose", which reads as though the client calls this function; it has its own copy, and nothing in `server/elm` reaches the branch. It now says that, and that it is answered for symmetry with the client rather than because anything here asks.

The divergence you noted alongside it — the client returns `never` for Rwanda IPV doses 3 and up where this fall-through gives 14 weeks — is the same class as finding 3, and is covered by B-154 rather than by special-casing.

PR #2055, opened 2026-08-09
Deploy: let a failed deploy say so

### LINKED ISSUE #2054: Deploy: the deploy chain can fail and still report success
## Summary

Three places in `server/RoboFile.php` where a deploy can fail and report success. Each is small; together they mean the output you watch during a release cannot be trusted to mean the code arrived, or arrived anywhere in particular.

## (a) The push result is tested against the wrong variable

`:139-140`:

```php
$push_status = $this->_exec("cd $pantheonDirectory && git pull && git add . && git commit -am 'Site update' && git push")->getExitCode();
if ($client_sync_result != 0) {
  throw new Exception('Failed to push to Pantheon');
}
```

`$push_status` is assigned and never read. The condition re-tests `$client_sync_result` — the rsync exit code from `:108`, already checked at `:109`, so zero by the time we reach here. A failed `git pull`, `commit` or `push` passes, and `deployPantheonSync()` then runs `cc`, `updb` and `fra` **against an environment that never received the code**.

There is also no `git pull` before the rsync: checkout at `:71`, rsyncs at `:96` and `:108`, and the only pull at `:139` against an already-dirty tree. On a two-machine deploy the pull fails exactly where the failure is swallowed.

## (b) `deployPantheonSync` discards its result

`:169-186` ends `->run();` with the Result thrown away and the method returning void, so `ddev robo deploy:pantheon:sync live` **exits 0 even when terminus, `updb` or `fra` failed**. The shell `&&` stops later steps, but nothing surfaces it.

## (c) An unset `PANTHEON_NAME` deploys to a different real site

`:155-165` falls back to `const PANTHEON_NAME = 'eheza-app'` when the variable is unset, while `deployPantheon()` refuses to run without it (`:32-35`).

**`eheza-app` is a real Pantheon site**, not a placeholder. So the fallback does not fail against a nonexistent target — it runs `cc`, `updb`, `fra` and `uli` against the wrong live site, with the errors hidden by (b).

## Fix

* Test `$push_status`, and pull as part of the checkout rather than after the rsync.
* Return the sync result, and raise when a step failed.
* Require `PANTHEON_NAME` in the sync as the deploy already does, and **delete the constant** — it is the mechanism of (c), and its only other use was a check that could never fire (`empty('eheza-app')`).

anvmn:
Fixes #2054

Three places in `server/RoboFile.php` where a deploy could fail and still report success. Each is small; together they mean the output you watch during a release does not reliably say that the code arrived, or where.

## What changes

**The push result is actually tested.** `$push_status` was assigned and never read — the throw beside it re-tested `$client_sync_result`, the rsync code from `:108` that `:109` had already checked and which is therefore zero by the time it is reached. A failed `git pull`, `commit` or `push` went straight on to `cc`, `updb` and `fra` against an environment that never received the code.

**The pull happens with the checkout, not after the rsync.** It ran at `:139` against an already-dirty tree, which is precisely where a two-machine deploy fails — and it failed inside the statement whose result was being ignored.

**The sync hands back what happened.** `deployPantheonSync` ended `->run();` with the Result discarded and returned void, so the command exited 0 even when terminus, `updb` or `fra` failed. It now raises on failure and returns the result.

**`PANTHEON_NAME` is required, and the constant is gone.** The sync fell back to `const PANTHEON_NAME = 'eheza-app'` when the variable was unset. **That is a real site of ours** — I confirmed it in the Pantheon site list — so the fallback did not fail against a nonexistent target, it ran `cc`, `updb`, `fra` and `uli` against a different live environment, with the errors hidden by the previous defect. The constant is deleted rather than left in place: it is the mechanism of this defect, and its only other use was `empty('eheza-app')`, a check that could never fire.

The deploy skill's reference doc documented the fallback, so it is updated too.

## Verification, and what it cannot cover

`php -l` clean; `phpcs` Drupal and DrupalPractice clean.

I cannot exercise a real deploy, and did not fake one. What I did run is the branching of all three defects with the shell stubbed:

```
                                          BEFORE                     AFTER
(a) git push fails, rsync succeeded       continues to cc/updb/fra   throws
(a) everything succeeds                   continues to cc/updb/fra   continues to cc/updb/fra
(b) updb fails during sync                caller sees success        throws
(b) all steps succeed                     caller sees success        returns success
(c) PANTHEON_NAME unset                   eheza-app (another site)   throws
(c) PANTHEON_NAME=vhw                     vhw                        vhw
```

Worth one reviewer's eye in particular: **removing the constant would fatal at runtime if any use were left behind**, since `self::PANTHEON_NAME` becomes an undefined class constant — and `php -l` does not catch that. I grepped the file and the repo; the only remaining references are to the environment variable, and the one dead `self::` use was removed with it.

### DIFF
--- a/.claude/skills/_deploy-common/deploy-reference.md
+++ b/.claude/skills/_deploy-common/deploy-reference.md
@@ -22,10 +22,10 @@ and `server/RoboFile.php`.
 2. **Aborts if the eheza-app working tree is dirty.**
 3. **Aborts if the Pantheon clone (`.pantheon-<name>`) is dirty.**
 4. **Aborts if `pantheon.upstream.yml` is missing or has no `php_version:`.**
-5. Checks out the target branch in the Pantheon clone (`master` for Dev, else the multidev branch name).
+5. Checks out the target branch in the Pantheon clone (`master` for Dev, else the multidev branch name), then pulls it. Checkout and pull fail with different messages, so a pull that fails for its own reasons — no network to the code server, no upstream, a conflict — does not read as a missing branch.
 6. rsyncs `www/.` (server) and `client/dist/.` (app) into the clone, excluding `.git`, `.ddev`, `client`, etc.
 7. Prints `git status`, then asks **"Commit changes and deploy?"** (interactive — needs a human).
-8. On confirm: `git pull && git add . && git commit -am 'Site update' && git push` to Pantheon.
+8. On confirm: `git add .`, commit if there is anything to commit, then `git push` to Pantheon. Nothing to commit is not a failure — re-running a deploy after fixing something on the Pantheon side finds no change to make, and the push still runs in case an earlier run committed without pushing.
 9. Calls `deployPantheonSync(<env>, FALSE)` → runs `cc all` (×2), `updb -y`, `fra -y`, `cc all`, `uli` on that env.
 
 `$branchName == 'master'` maps to the Pantheon **`dev`** environment.
--- a/.claude/skills/_deploy-common/deploy-reference.md
+++ b/.claude/skills/_deploy-common/deploy-reference.md
@@ -18,7 +18,7 @@ and `server/RoboFile.php`.
 ## What the robo commands actually do (`server/RoboFile.php`)
 
 ### `deployPantheon($branchName = 'master')` — `ddev robo deploy:pantheon [env]`
-1. Resolves `PANTHEON_NAME` from the env var (falls back to the `PANTHEON_NAME` const, `eheza-app`, if unset — which is why the env var must be set correctly).
+1. Resolves `PANTHEON_NAME` from the env var, and refuses to run without it. It used to fall back to a `PANTHEON_NAME` const of `eheza-app` — which is itself a real site, so an unset variable deployed to someone else's environment.
 2. **Aborts if the eheza-app working tree is dirty.**
 3. **Aborts if the Pantheon clone (`.pantheon-<name>`) is dirty.**
 4. **Aborts if `pantheon.upstream.yml` is missing or has no `php_version:`.**
--- a/server/RoboFile.php
+++ b/server/RoboFile.php
@@ -56,12 +56,23 @@ class RoboFile extends Tasks {
     }
 
     $result = $this
-      ->taskExec("cd $pantheonDirectory && git checkout $branchName && git pull")
+      ->taskExec("cd $pantheonDirectory && git checkout $branchName")
       ->printOutput(FALSE)
       ->run();
 
     if (!$result->wasSuccessful()) {
       throw new Exception("Specified branch $branchName does not exist.");
+    }
+
+    // Before the rsync, so that the tree is clean when it runs. Its own check,
+    // because a pull fails for reasons that have nothing to do with the branch
+    // existing - no network to the code server, no upstream, a conflict.
+    $result = $this
+      ->taskExec("cd $pantheonDirectory && git pull")
+      ->run();
+
+    if (!$result->wasSuccessful()) {
+      throw new Exception("Failed to pull $branchName in the Pantheon directory ($pantheonDirectory)");
     }
 
     $rsyncExclude = [
@@ -124,7 +135,11 @@ class RoboFile extends Tasks {
       return;
     }
 
-    $push_status = $this->_exec("cd $pantheonDirectory && git add . && git commit -am 'Site update' && git push")->getExitCode();
+    // `git commit` exits non-zero when there is nothing to commit, which is an
+    // ordinary case: re-running the deploy after fixing something on the
+    // Pantheon side finds the rsync has produced no change. Push regardless -
+    // an earlier run may have committed without getting as far as pushing.
+    $push_status = $this->_exec("cd $pantheonDirectory && git add . && { git diff --cached --quiet || git commit -m 'Site update'; } && git push")->getExitCode();
     if ($push_status != 0) {
       throw new Exception('Failed to push to Pantheon');
     }
@@ -142,12 +157,17 @@ class RoboFile extends Tasks {
    *   Determine if a deploy should be done by terminus. That is, for example
    *   should TEST environment be updated from DEV.
    *
+   * @return \Robo\Result
+   *   The result of the deploy steps.
+   *
+   * @throws \Exception
+   *   If PANTHEON_NAME is not set, or any of the deploy steps failed.
    * @throws \Robo\Exception\TaskException
    */
   public function deployPantheonSync(string $env = 'test', bool $doDeploy = TRUE) {
-    // Required, as deployPantheon() requires it. The constant below is another
-    // real site, so falling back to it would run these steps against someone
-    // else's environment.
+    // Required, as deployPantheon() requires it. This used to fall back to a
+    // constant naming another of our sites, so an unset variable deployed to
+    // someone else's environment rather than failing.
     $pantheonName = getenv('PANTHEON_NAME');
     if (!$pantheonName) {
       throw new Exception('Please specify PANTHEON_NAME in your DDEV local config, so it is clear which site is being deployed to.');
--- a/server/RoboFile.php
+++ b/server/RoboFile.php
@@ -9,14 +9,6 @@ use Symfony\Component\Yaml\Yaml;
 class RoboFile extends Tasks {
 
   /**
-   * The Pantheon name.
-   *
-   * You need to fill this information for Robo to know what's the name of your
-   * site.
-   */
-  const PANTHEON_NAME = 'eheza-app';
-
-  /**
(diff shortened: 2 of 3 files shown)


server/RoboFile.php
```diff
@@ -136,8 +124,8 @@ public function deployPantheon($branchName = 'master') {
       return;
     }
 
-    $push_status = $this->_exec("cd $pantheonDirectory && git pull && git add . && git commit -am 'Site update' && git push")->getExitCode();
-    if ($client_sync_result != 0) {
+    $push_status = $this->_exec("cd $pantheonDirectory && git add . && git commit -am 'Site update' && git push")->getExitCode();
```
anvmn: **Finding 1 — a no-change redeploy now hard-aborts** (medium, `/code-review medium`)

**Confirmed, and it is a regression this PR introduces — it must be fixed before merge.**

`git commit` exits **1** when there is nothing to commit (verified by running it, not assumed), so the corrected `$push_status != 0` check throws `Failed to push to Pantheon` and `deployPantheonSync()` never runs. Before this PR the same exit code was harmless, because the check tested a variable that was always zero.

The scenario is one the `deploy-multidev` skill documents in its own preflight: the first deploy pushes the code and then fails on terminus auth; the operator authenticates and re-runs to get the post-deploy steps; the rsync now produces no diff, the commit exits 1, and the command throws. It will throw on every retry until some file changes, leaving the environment on new code with a stale cache and registry.

That is a worse failure than the one this PR fixes — a deploy that cannot be completed, rather than one that lies about having completed.

Fix: treat "nothing to commit" as success, and still push (there may be an earlier commit that never got pushed).

server/RoboFile.php
```diff
@@ -68,7 +56,7 @@ public function deployPantheon($branchName = 'master') {
     }
 
     $result = $this
-      ->taskExec("cd $pantheonDirectory && git checkout $branchName")
+      ->taskExec("cd $pantheonDirectory && git checkout $branchName && git pull")
```
anvmn: **Finding 2 — a failed `git pull` is reported as "branch does not exist"** (low/medium, `/code-review medium`)

**Confirmed, and mine.** Moving the pull into the checkout's `taskExec` put it under a single exit-code check whose message is `Specified branch $branchName does not exist.` So a network or SSH failure to the Pantheon code server, a branch with no upstream, or a merge conflict all surface as a missing branch — and `deploy-multidev/SKILL.md:57` teaches operators exactly that reading, sending them to create an environment that already exists.

The review also notes the pull runs under `printOutput(FALSE)` with no TTY, so a merge wanting an editor or a credential prompt fails or hangs with its output swallowed.

Fix: two `taskExec` calls with distinct messages.

server/RoboFile.php
```diff
@@ -136,8 +124,8 @@ public function deployPantheon($branchName = 'master') {
       return;
     }
 
-    $push_status = $this->_exec("cd $pantheonDirectory && git pull && git add . && git commit -am 'Site update' && git push")->getExitCode();
-    if ($client_sync_result != 0) {
+    $push_status = $this->_exec("cd $pantheonDirectory && git add . && git commit -am 'Site update' && git push")->getExitCode();
+    if ($push_status != 0) {
```
anvmn, line 143: **Finding 3 — the pull is now separated from the push by the interactive confirm** (low, `/code-review medium`)

**Real, and I am declining it as proposed.**

The observation holds: the pull happens at checkout, then rsync, `git status` and a `Commit changes and deploy?` prompt that can wait minutes for a human, so the remote can advance in between and the push is rejected non-fast-forward.

But the suggested `git pull --ff-only` immediately before the push would not help. By that point there is a local commit, so a fast-forward is impossible and the pull fails instead of the push. Making it work means `--rebase`, which is a real decision about how a deploy checkout should behave, not a tidy-up — and rebasing a Pantheon checkout during an interactive deploy deserves its own thought rather than being folded into a fix for something else.

Recovery today is already sound: the next run pulls at checkout. Recording rather than fixing blind.

.claude/skills/_deploy-common/deploy-reference.md
```diff
@@ -18,7 +18,7 @@ and `server/RoboFile.php`.
 ## What the robo commands actually do (`server/RoboFile.php`)
 
 ### `deployPantheon($branchName = 'master')` — `ddev robo deploy:pantheon [env]`
-1. Resolves `PANTHEON_NAME` from the env var (falls back to the `PANTHEON_NAME` const, `eheza-app`, if unset — which is why the env var must be set correctly).
+1. Resolves `PANTHEON_NAME` from the env var, and refuses to run without it. It used to fall back to a `PANTHEON_NAME` const of `eheza-app` — which is itself a real site, so an unset variable deployed to someone else's environment.
```
anvmn, line 21: **Finding 4 — this PR left the deploy docs half-updated** (low, `/code-review medium`)

**Confirmed, and mine.** I updated this line for the removed constant and stopped there. Step 8 still reads ``git pull && git add . && git commit -am 'Site update' && git push``, and step 5 still describes the checkout without the pull — both stale as of this PR, in the file the two deploy skills read as their source of truth.

`deploy-multidev/SKILL.md:57`, which explains the "Specified branch does not exist" message, is stale for the same reason (see finding 2).

Fix: move the pull in the docs where it moved in the code, and correct the SKILL.md explanation once finding 2 is addressed.

server/RoboFile.php
```diff
@@ -182,8 +171,14 @@ public function deployPantheonSync(string $env = 'test', bool $doDeploy = TRUE)
       // clear caches again so the reverted config takes effect.
       ->exec("terminus remote:drush $pantheonTerminusEnvironment -- fra -y")
       ->exec("terminus remote:drush $pantheonTerminusEnvironment -- cc all")
-      ->exec("terminus remote:drush $pantheonTerminusEnvironment -- uli")
-      ->run();
+      ->exec("terminus remote:drush $pantheonTerminusEnvironment -- uli");
+
+    $result = $task->run();
+    if (!$result->wasSuccessful()) {
+      throw new Exception("Deploy steps failed on $pantheonTerminusEnvironment");
```
anvmn, line 198: **Finding 5 — the docblock no longer matches the method** (low, `/code-review medium`)

**Confirmed, and mine.** `deployPantheonSync()` now throws plain `\Exception` in two places and returns a `\Robo\Result`, while the docblock still declares only `@throws \Robo\Exception\TaskException` and no `@return`.

CI will not catch it: `ci-scripts/test_coder.sh` sniffs `server/hedley/modules/custom/*` only, so `RoboFile.php` is never linted. Worth fixing precisely because nothing else will.

server/RoboFile.php
anvmn: Fixed in 7cb7bd082e0a7050cd736dfb9b7a4fb481ea7072.

```
git add . && { git diff --cached --quiet || git commit -m 'Site update'; } && git push
```

Nothing to commit is no longer a failure, and the push still runs — an earlier run may have committed without getting that far.

Proved against real git rather than reasoned about, in a throwaway repo with a bare remote:

| case | old form | new form |
|---|---|---|
| a change to commit | exit 0 | exit 0 |
| nothing to commit | **exit 1** | exit 0 |
| committed but never pushed | — | exit 0, and the commit reaches the remote |

The last row is the one that matters for the retry story: a run that committed and then failed on terminus auth is completed by the next run rather than blocked by it.

server/RoboFile.php
anvmn: Fixed in 7cb7bd082e0a7050cd736dfb9b7a4fb481ea7072 — the checkout and the pull are separate `taskExec` calls with separate messages now, the pull's being `Failed to pull <branch> in the Pantheon directory`.

`deploy-multidev/SKILL.md` is updated too: its preflight explained the "Specified branch does not exist" message as the only outcome, and now names both.

.claude/skills/_deploy-common/deploy-reference.md
anvmn: Fixed in 7cb7bd082e0a7050cd736dfb9b7a4fb481ea7072. Step 5 now says the checkout pulls and that the two fail differently; step 8 describes the conditional commit rather than the old `git pull && git add . && …` chain. `deploy-multidev/SKILL.md:57` is corrected as part of finding 2.

server/RoboFile.php
anvmn: Fixed in 7cb7bd082e0a7050cd736dfb9b7a4fb481ea7072 — the docblock now declares `@return \Robo\Result` and the `\Exception` thrown for an unset `PANTHEON_NAME` or a failed step, alongside the existing `@throws \Robo\Exception\TaskException`.

Worth keeping the reason you gave: nothing would have caught this, since `ci-scripts/test_coder.sh` sniffs `server/hedley/modules/custom/*` only and never lints `RoboFile.php`.

PR #2057, opened 2026-08-09
Build: make gulp publish fail when the build fails

### LINKED ISSUE #2056: Build: gulp publish exits 0 and ships an error page when the build fails
## Problem

`gulp publish` reports success when the Elm compile fails, and packages the compiler error page as the app.

The `elm` task pipes through `plumber()`, which exists for the watch loop: during `gulp` (dev) a compile error must not kill the watcher, and `gulp-elm` writes the error into the output HTML so the developer sees it in the browser. `publish` reuses the same task, so the same swallowing applies to a production build.

The `minify` task has the equivalent problem: its uglify error handler logs to the console and lets the stream continue.

## Reproduced

Introduced a deliberate type error in `client/src/elm/Pages/Device/View.elm` and ran `gulp publish` on `develop`:

```
gulp exit=0
dist/index.html: 440 bytes
<!DOCTYPE HTML><html><body><pre>-- TYPE MISMATCH ---------- src/elm/Pages/Device/View.elm
```

A green run, and the deployable artifact is a page displaying the compile error instead of the application. `ddev robo deploy:pantheon` would push that.

## Fix

Have the shared tasks distinguish the two callers. `publish` sets a flag first; when it is set, the `elm` task runs without `plumber()` and without an error handler so the error propagates, and the uglify handler destroys the stream. The watch path is untouched.

After the fix, same broken file:

```
gulp exit=1
dist/index.html: not created
'elm' errored after 757 ms
'build' errored after 777 ms
'publish' errored after 800 ms
```

and with the file restored, `publish` completes normally (exit 0, full 1380-byte `dist/index.html`).

## Related

A failed `publish` stops before `deploy:revert`, leaving the tracked `client/src/elm/Config.elm` holding deployment config. That is pre-existing, but this change makes the failure path reachable, so it is now worth acting on.

anvmn:
Fixes #2056

`gulp publish` exits 0 when the build fails, and writes the compiler error page to `dist/index.html` as the app.

The cause is tasks shared by two callers with opposite needs. `elm` and `styles` both pipe through `plumber()` and both catch their compiler's error, write it into `serve/index.html` and continue — that is what the watch loop needs, so a typo does not kill the watcher and the error shows up in the browser. `publish` runs the same tasks and inherits the swallowing.

Both tasks now attach `plumber` and their error handler only when not publishing. `publish` sets the flag as its first step and clears it as its last, so a graph that runs publish before a serving task in one process leaves the watch loop intact.

## Verified by running it

Deliberate type error in `client/src/elm/Pages/Device/View.elm`, `gulp publish`, on `develop`:

```
gulp exit=0
dist/index.html: 440 bytes
<!DOCTYPE HTML><html><body><pre>-- TYPE MISMATCH ---------- src/elm/Pages/Device/View.elm
```

On this branch, four runs:

| run | expected | got |
|---|---|---|
| broken Elm + `publish` | fail | exit 1, no dist |
| broken SCSS + `publish` | fail | exit 1, no dist |
| broken Elm + `build` (watch path) | still writes the error page | exit 0, error page intact |
| clean + `publish` | succeed | exit 0, real 1380-byte dist |

The third row is the discrimination test: the dev loop keeps the behaviour the comments say it must.

## Review history

The first commit fixed only the Elm path and also changed the uglify error handler. Review found the SCSS path unfixed — the same failure, reached through SASS — and that the uglify change was inert, since `this.destroy()` there acts on a mid-pipeline stream rather than the one the task returns. A parse error in a minified file ends identically with or without it, on this branch and on develop. That hunk is gone; the SCSS path is fixed.

## Not in this PR

A failing `publish` stops before `deploy:revert`, leaving the tracked `client/src/elm/Config.elm` holding deployment config. This change is what makes that path reachable for Elm errors. It belongs with the deploy-safety work (B-146) rather than here, and is recorded as an inline comment.

### DIFF
--- a/client/gulpfile.js
+++ b/client/gulpfile.js
@@ -14,6 +14,11 @@ var $ = require("gulp-load-plugins")();
 // "del" is used to clean out directories and such
 var del = require("del");
 
+// Publishing must fail when the build does. The watch loop must not: it keeps
+// running so that the next save can fix the error, and shows the error in the
+// browser meanwhile. The same tasks serve both, so they ask which they are in.
+var isPublishing = false;
+
 var rename = require("gulp-rename");
 // BrowserSync isn't a gulp package, and needs to be loaded manually
 var browserSync = require("browser-sync");
@@ -272,6 +277,11 @@ gulp.task("minify", gulp.series(
         mangle: false
       }))).on('error', function(err) {
         console.error(err);
+
+        if (isPublishing) {
+          // A partially minified dist is not something to ship quietly.
+          this.destroy(err);
+        }
       })
 
     // Minify CSS
@@ -309,6 +319,18 @@ gulp.task("deploy", function() {
 });
 
 gulp.task('elm', gulp.series('version', function elmCompile() {
+  if (isPublishing) {
+    // No plumber and no error handler, so a compile error ends the task and
+    // with it the publish. Swallowing it here would leave the error page that
+    // the watch loop writes to be packaged into dist and shipped as the app.
+    return gulp.src('src/elm/Main.elm')
+      .pipe(elm({
+        'debug': false,
+        'warn': false
+      }))
+      .pipe(gulp.dest('serve'));
+  }
+
   return gulp.src('src/elm/Main.elm')
     .pipe(plumber())
     .pipe(elm({
@@ -489,7 +511,13 @@ gulp.task("deploy:revert", function(cb) {
 
 // Builds your site with the "build" command and then runs all the optimizations on
 // it and outputs it to "./dist"
+gulp.task("publish:mark", function markPublishing(cb) {
+  isPublishing = true;
+  cb();
+});
+
 gulp.task("publish", gulp.series(
+  "publish:mark",
   "deploy:config",
   gulp.parallel("build", "clean:prod"),
   gulp.parallel("minify", "cname", "images"),
--- a/client/gulpfile.js
+++ b/client/gulpfile.js
@@ -61,12 +61,20 @@ gulp.task("clean:prod", function() {
 });
 
 // Compiles the SASS files and moves them into the "assets/stylesheets" directory
-gulp.task("styles", function() {
+gulp.task("styles", function styleCompile() {
   // Looks at the style.scss file for what to include and creates a style.css file
-  return gulp.src("src/assets/scss/style.scss")
-    .pipe(plumber())
-    .pipe(sass({silenceDeprecations: ['legacy-js-api', 'import']}))
-    .on('error', function(err) {
+  var stream = gulp.src("src/assets/scss/style.scss");
+
+  // Publishing attaches neither plumber nor the handler below, so a SASS error
+  // travels down the stream and fails the task.
+  if (!isPublishing) {
+    stream = stream.pipe(plumber());
+  }
+
+  stream = stream.pipe(sass({silenceDeprecations: ['legacy-js-api', 'import']}));
+
+  if (!isPublishing) {
+    stream = stream.on('error', function(err) {
       browserSync.notify("SASS error");
 
       console.error(err.message);
@@ -79,7 +87,10 @@ gulp.task("styles", function() {
 
       // No need to continue processing.
       this.emit('end');
-    })
+    });
+  }
+
+  return stream
     // AutoPrefix your CSS so it works between browsers
     .pipe($.autoprefixer("last 1 version", {
       cascade: true
@@ -277,11 +288,6 @@ gulp.task("minify", gulp.series(
         mangle: false
       }))).on('error', function(err) {
         console.error(err);
-
-        if (isPublishing) {
-          // A partially minified dist is not something to ship quietly.
-          this.destroy(err);
-        }
       })
 
     // Minify CSS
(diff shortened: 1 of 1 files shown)


client/gulpfile.js
```diff
@@ -272,6 +277,11 @@ gulp.task("minify", gulp.series(
         mangle: false
       }))).on('error', function(err) {
         console.error(err);
+
+        if (isPublishing) {
```
anvmn: **This hunk is inert — it should come out or be replaced.**

`this.destroy(err)` here acts on a mid-pipeline stream, not the stream the task returns, so it changes nothing observable.

Verified by running: appended a parse error to `serve/uuid.js`, then ran the minify task on this branch and on `develop`. Identical outcome both times:

```
exit=1
The following tasks did not complete: minify, minifyFiles
Did you forget to signal async completion?
```

So a uglify error already fails today — but by stalling the task and letting the event loop drain, not by failing cleanly. This PR neither improves nor worsens that. Suggest dropping the hunk and tracking the stall separately, or routing the error to the task callback so it fails with a real message.

client/gulpfile.js
```diff
@@ -489,7 +511,13 @@ gulp.task("deploy:revert", function(cb) {
 
 // Builds your site with the "build" command and then runs all the optimizations on
 // it and outputs it to "./dist"
+gulp.task("publish:mark", function markPublishing(cb) {
+  isPublishing = true;
```
anvmn, line 520: `isPublishing` is set once and never cleared. Any graph that runs publish before a watch/serve task in the same process (`gulp publish watch`, or a task composed after publish later) leaves it set, so the watch loop's elm task takes the no-plumber path and dies on the first compile error — the behaviour the comment above says the watch loop must keep.

Not reachable via the documented `ddev gulp publish` workflow, so low. Clearing the flag at the end of the series removes the coupling.

client/gulpfile.js
```diff
@@ -489,7 +511,13 @@ gulp.task("deploy:revert", function(cb) {
...
+  cb();
+});
+
 gulp.task("publish", gulp.series(
+  "publish:mark",
```
anvmn, line 532: Confirmed by running, and noted in the PR description: the failing run's task list aborts at `build` and `deploy:revert` never starts, so the tracked `client/src/elm/Config.elm` is left holding production deploy config.

This PR is what makes the path reachable for Elm errors, so it is worth deciding here rather than deferring: either restore the config whether or not the series succeeds, or take it with B-146(c).

anvmn (commented):
Review findings (medium), all four verified by running them in a worktree rather than by reading. One of them shows a hunk in this PR is inert, and one shows the PR only half-delivers its own goal — details inline and below.

anvmn:
**Out-of-diff finding (the most important one) — `client/gulpfile.js:64-82`, the `styles` task.**

This PR only half-delivers its own goal. The SASS path has the identical shape to the Elm one and is untouched: `styles` logs the error, writes `<pre>error</pre>` over `serve/index.html`, and calls `this.emit('end')` regardless of `isPublishing`. `minify` re-runs `styles` before packaging.

Verified by running, on this branch:

```
$ printf '\n.broken { color: ; @@@ }\n' >> src/assets/scss/style.scss
$ gulp publish
exit=0
dist/index.html: 230 bytes
<!DOCTYPE HTML><html><body><pre>src/assets/scss/style.scss
Error: Expected expression.
```

Exit 0, and the deployable artifact is an error page — exactly the failure this PR exists to prevent, reached through SASS instead of Elm.

Posting as a general comment because the line is outside the diff.

anvmn:
Pushed `21e703a69`, which takes three of the four findings.

**SCSS (the one that mattered)** — `styles` had the identical shape to `elm`, so a SASS error still published an error page. Both tasks now attach `plumber` and their error handler only when not publishing, written the same way in both places.

**Uglify hunk** — removed. Verified it was inert: a parse error in a minified file ends identically with or without `this.destroy(err)`, on this branch and on develop.

**Flag reset** — cleared at the end of the series.

**Config.elm on a failed publish** — not taken here. B-146 was parked, and this belongs with it rather than with the build-failure fix.

Re-verified all four paths by running them:

| run | expected | got |
|---|---|---|
| broken Elm + `publish` | fail | exit 1, no dist |
| broken SCSS + `publish` | fail | exit 1, no dist |
| broken Elm + `build` (watch path) | still writes the error page | exit 0, error page intact |
| clean + `publish` | succeed | exit 0, real 1380-byte dist |

The third row is the discrimination test — it confirms the dev loop keeps the behaviour the comments say it must.

client/gulpfile.js
anvmn: Confirmed and removed in `21e703a69`.

I ran it rather than reasoning about it: a parse error in a non-excluded JS file ends identically on this branch and on develop — `exit=1`, `The following tasks did not complete: minify, minifyFiles / Did you forget to signal async completion?`. So the hunk bought nothing. The predicted indefinite hang does not occur either, because nothing in `publish` holds the event loop open; the task stalls and the process exits when the loop drains. Failing cleanly there would mean routing the error to the task callback, which is a separate change from this branch's purpose.

client/gulpfile.js
anvmn: Fixed in `21e703a69` — a `publish:unmark` step clears the flag at the end of the series.

See the thread on line 522: that reset does not cover the case where an earlier step fails, which is worth its own answer.

client/gulpfile.js
anvmn: Confirmed, and deliberately not taken here.

B-146, the deploy-safety cluster this belongs to, was parked. Since this PR is what makes the path reachable for Elm errors, I did not want it to disappear silently — hence the inline note and the section in the PR description. If it bites anyone, the fix is to restore the config whether or not the series completes.

client/gulpfile.js
```diff
@@ -489,9 +516,23 @@ gulp.task("deploy:revert", function(cb) {
 
 // Builds your site with the "build" command and then runs all the optimizations on
 // it and outputs it to "./dist"
+gulp.task("publish:mark", function markPublishing(cb) {
+  isPublishing = true;
+  cb();
+});
```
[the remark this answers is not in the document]
anvmn: True as written, and it is the reason I graded my own version of this finding low rather than fixing it properly.

Two things narrow it. A failed `publish` ends the task graph, and gulp exits — so there is normally nothing left in the process for the stale flag to affect. And in the case the comment names, `gulp publish watch`, the CLI runs the two in parallel by default, so `watch` is already running under a set flag from the moment `publish:mark` completes; resetting at the end would not have protected it either.

Making it airtight means scoping the flag rather than resetting it — deciding from the environment or a separate task graph, so no task's behaviour depends on what ran before it in the same process. That is a larger change than this branch, and I would rather do it deliberately than bolt a completion callback onto the series. Happy to take it if the maintainers prefer it here.

PR #2059, opened 2026-08-09
Person records: stop a person edit from erasing recorded GPS coordinates

### LINKED ISSUE #2058: Person records: editing a person erases previously recorded GPS coordinates
## Problem

Any edit to a person — a corrected name is enough — erases the GPS coordinates recorded when that person was registered.

The edit form has no coordinate inputs, so `validatePerson` reports `Nothing` for both (`client/src/elm/Backend/Person/Form.elm:452-453`, the two entries between `village` and `saveGPSLocation`, matching the field order in `Backend/Person/Model.elm:31-32`). `PatchPerson` re-attaches coordinates only when a fresh reading was requested (`Backend/Update.elm:4632`, and that checkbox defaults unchecked), the encoder sends them regardless as `maybe string` (`Backend/Person/Encoder.elm:43-44`), and the backend applies the resulting `null`.

Nothing reads these fields yet on the server side, which is why the loss went unnoticed: what is destroyed is data collected for future use, and no screen can restore it.

## It has already happened

The `gps_coordinates` feature flag is on at `vhw` and `tip-somalia`, and unset at `ihangane`. On vhw.live:

- 18,331 persons, 10,599 currently carrying coordinates
- 10,676 distinct persons appear in `field_revision_field_latitude`
- the 77 in the difference are all still-existing person nodes that had coordinates and have none now
- 76 of those 77 were edited after creation, which is the mechanism

tip-somalia has the flag on but only one coordinate record.

## Fix

Three parts:

1. **Backend guard.** In `HedleyRestfulSync`, drop null coordinates from a person `PATCH`. This protects every device the moment it deploys, including the many in the field running older builds that will keep sending nulls for months. An edit carrying a fresh reading still updates the fields, and creation is untouched.
2. **Client.** `PatchPerson` carries over the stored coordinates instead of the form's `Nothing`.
3. **Repair.** The erased values are still in the node revisions, so a one-shot script can restore them.

anvmn:
Fixes #2058

Any edit to a person erases the GPS coordinates recorded at registration. The edit form has no coordinate inputs, so `validatePerson` reports `Nothing` for both, `PatchPerson` re-attaches them only when a fresh reading was requested, and the encoder sends the `Nothing` as `null` regardless. The backend applies it. A corrected name destroys a household's location, and no screen can put it back.

On vhw.live it has already happened to **77 persons** — they appear in `field_revision_field_latitude` but have no current value, they are all still-existing person nodes, and 76 of the 77 were edited after creation. Another 10,599 carry coordinates today and are one edit away.

## Three parts

**1. Backend guard** (`HedleyRestfulSync`) — null coordinates are dropped from a person `PATCH`. This is the half that matters first: devices in the field run whatever build they last received, so a client-only fix leaves every un-updated device erasing coordinates for months, while this takes effect for all of them the moment it deploys. It sits in the existing per-type `$ignored` idiom, two lines below where `save_gps_location` is filtered for the same kind of reason. An edit that *does* carry a fresh reading still updates the fields, and `POST` is untouched, so registration still records them.

**2. Client** (`Backend/Update.elm`) — `PatchPerson` carries over the stored coordinates instead of the form's `Nothing`. It goes here rather than in the form because the validator reporting `Nothing` is defensible — the form genuinely has no coordinate inputs — while treating "the form knows nothing" as "clear the field" is not.

**3. Repair script** — the erased values are still in the node revisions, so `restore-wiped-gps-coordinates.php` restores the most recent revision holding both coordinates, to persons that hold neither now. Latitude and longitude are taken from the *same* revision, so the restored pair is one reading from one moment; a person with only one of the two is reported and skipped rather than guessed at.

```
drush scr profiles/hedley/modules/custom/hedley_person/scripts/restore-wiped-gps-coordinates.php --dry_run
```

Only vhw needs it. The flag is on at `vhw` and `tip-somalia`, unset at `ihangane`; tip-somalia has one coordinate record in total.

## Verified by running

**The repair script, end to end on a local copy** — recorded coordinates on a person, wiped them the way an edit does (`node_save` with the fields emptied), then:

| step | result |
|---|---|
| dry run | `Person 28590 would be restored to -1.9440000, 30.0619000`, current rows still 0 |
| real run | `Restored coordinates for 1 of 1 persons` |
| stored values afterwards | `-1.9440000` / `30.0619000` — the originals |
| second dry run | `No person is missing coordinates that a revision still holds` (idempotent) |

The test person was returned to having no coordinates afterwards.

**Its selection against real data** — the script's query, run against vhw.live, selects exactly the **77** persons independently identified as having lost coordinates. (The script also excludes soft-deleted persons, so it may restore slightly fewer.)

**Static checks** — `php -l` clean on both PHP files; phpcs `Drupal` and `DrupalPractice` with the full CI extension list both exit 0 with no output, and a deliberately bad file makes the same invocation exit 2 with 12 findings, so the standards really are loaded. `elm make src/elm/Main.elm` compiles 548 modules; `elm-format --validate` clean; `elm-review` reports one pre-existing unused import in the touched file, identical on develop, and nothing new.

## Not covered

The sibling defect recorded with this one: a stale GPS reading from a previous household can display as current and be saved onto the next person, because geolocation errors send no reply to Elm and `model.coordinates` is not cleared per request. Different bug, same feature — left out deliberately.

### DIFF
--- a/client/src/elm/Backend/Update.elm
+++ b/client/src/elm/Backend/Update.elm
@@ -4438,7 +4438,7 @@ updateIndexedDb language currentDate currentTime coordinates zscores site featur
                     -- Adding GPS coordinates.
                     personWithCoordinates =
                         if gpsCoordinatesEnabled features && person.saveGPSLocation then
-                            updatePersonWithCooridnates person coordinates
+                            updatePersonWithCoordinates person coordinates
 
                         else
                             person
@@ -4648,7 +4648,7 @@ updateIndexedDb language currentDate currentTime coordinates zscores site featur
                 -- fix this leaves what is stored alone.
                 personWithCoordinates =
                     if gpsCoordinatesEnabled features && person.saveGPSLocation then
-                        updatePersonWithCooridnates personWithStoredCoordinates coordinates
+                        updatePersonWithCoordinates personWithStoredCoordinates coordinates
 
                     else
                         personWithStoredCoordinates
@@ -10469,8 +10469,8 @@ generateTuberculosisEncounterCompletedMsgs currentDate after id =
         |> Maybe.withDefault []
 
 
-updatePersonWithCooridnates : Person -> Maybe App.Model.GPSCoordinates -> Person
-updatePersonWithCooridnates person =
+updatePersonWithCoordinates : Person -> Maybe App.Model.GPSCoordinates -> Person
+updatePersonWithCoordinates person =
     Maybe.map
         (\coordinates ->
             { person
--- a/client/src/elm/Backend/Update.elm
+++ b/client/src/elm/Backend/Update.elm
@@ -4627,26 +4627,31 @@ updateIndexedDb language currentDate currentTime coordinates zscores site featur
 
         PatchPerson origin personId person ->
             let
-                -- Adding GPS coordinates.
+                -- The edit form has no coordinate inputs, so it always reports
+                -- Nothing for them. Patching that in would erase the location
+                -- recorded at registration, so the stored one is carried over.
+                personWithStoredCoordinates =
+                    Dict.get personId model.people
+                        |> Maybe.andThen RemoteData.toMaybe
+                        |> Maybe.map
+                            (\stored ->
+                                { person
+                                    | registrationLatitude = stored.registrationLatitude
+                                    , registrationLongitude = stored.registrationLongitude
+                                }
+                            )
+                        |> Maybe.withDefault person
+
+                -- Adding GPS coordinates. Laid over the stored ones, rather
+                -- than offered instead of them, because a reading that was
+                -- asked for is not always a reading that arrived: without a
+                -- fix this leaves what is stored alone.
                 personWithCoordinates =
                     if gpsCoordinatesEnabled features && person.saveGPSLocation then
-                        updatePersonWithCooridnates person coordinates
+                        updatePersonWithCooridnates personWithStoredCoordinates coordinates
 
                     else
-                        -- The edit form has no coordinate inputs, so it always
-                        -- reports Nothing for them. Patching that in would
-                        -- erase the location recorded at registration, so the
-                        -- stored one is carried over instead.
-                        Dict.get personId model.people
-                            |> Maybe.andThen RemoteData.toMaybe
-                            |> Maybe.map
-                                (\stored ->
-                                    { person
-                                        | registrationLatitude = stored.registrationLatitude
-                                        , registrationLongitude = stored.registrationLongitude
-                                    }
-                                )
-                            |> Maybe.withDefault person
+                        personWithStoredCoordinates
             in
             ( { model | postPerson = Loading }
             , sw.patchFull personEndpoint personId personWithCoordinates
--- a/client/src/elm/Backend/Update.elm
+++ b/client/src/elm/Backend/Update.elm
@@ -4633,7 +4633,20 @@ updateIndexedDb language currentDate currentTime coordinates zscores site featur
                         updatePersonWithCooridnates person coordinates
 
                     else
-                        person
+                        -- The edit form has no coordinate inputs, so it always
+                        -- reports Nothing for them. Patching that in would
+                        -- erase the location recorded at registration, so the
+                        -- stored one is carried over instead.
+                        Dict.get personId model.people
+                            |> Maybe.andThen RemoteData.toMaybe
+                            |> Maybe.map
+                                (\stored ->
+                                    { person
+                                        | registrationLatitude = stored.registrationLatitude
+                                        , registrationLongitude = stored.registrationLongitude
+                                    }
+                                )
+                            |> Maybe.withDefault person
             in
             ( { model | postPerson = Loading }
             , sw.patchFull personEndpoint personId personWithCoordinates
--- a/server/hedley/modules/custom/hedley_person/scripts/restore-wiped-gps-coordinates.php
+++ b/server/hedley/modules/custom/hedley_person/scripts/restore-wiped-gps-coordinates.php
@@ -1,0 +1,154 @@
+<?php
+
+/**
+ * @file
+ * Restores GPS coordinates that editing a person erased.
+ *
+ * The person edit form has no coordinate inputs, so it reported Nothing for
+ * them, and the resulting patch wrote that over whatever the device recorded
+ * at registration. One ordinary edit - a corrected name - was enough. The
+ * fields have no other writer and no screen that can restore them.
+ *
+ * What the edit did not touch is the node's revisions, so the erased values
+ * are still there to be read back. This restores the most recent revision
+ * that held both coordinates, to persons that hold neither now.
+ *
+ * Restoring a person writes a new revision, which every device holding that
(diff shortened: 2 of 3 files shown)


client/src/elm/Backend/Update.elm
```diff
@@ -4633,7 +4633,20 @@ updateIndexedDb language currentDate currentTime coordinates zscores site featur
                         updatePersonWithCooridnates person coordinates
 
                     else
-                        person
+                        -- The edit form has no coordinate inputs, so it always
+                        -- reports Nothing for them. Patching that in would
+                        -- erase the location recorded at registration, so the
+                        -- stored one is carried over instead.
+                        Dict.get personId model.people
```
anvmn: **The carry-over only guards the `else` branch, so one path still sends the destructive nulls.**

`updatePersonWithCooridnates` (`Backend/Update.elm:10454-10463`) ends in `Maybe.withDefault person` — with `coordinates = Nothing` it returns the record untouched.

So if the feature is on and the health worker ticks "save GPS location" while a fix is available, but the subscription drops before Save, the true branch is taken, nothing fresh is applied, and the form's `registrationLatitude = Nothing` is patched through. The stored coordinates are erased and only the backend guard prevents the loss.

The shape that closes it is to carry the stored values over first and apply a fresh reading on top when there is one, rather than treating the two as alternatives.

server/hedley/modules/custom/hedley_person/scripts/restore-wiped-gps-coordinates.php
```diff
@@ -0,0 +1,154 @@
...
+
+// Persons whose revisions hold a latitude, and whose current data holds none.
+// Longitude is not named here: it is read per person below, since a person
+// missing only one of the two is a case to report rather than to guess at.
+$query = db_select('field_revision_field_latitude', 'r');
+$query->join('node', 'n', 'n.nid = r.entity_id');
+// Field tables keep the rows of deleted field instances and are keyed by
+// entity type as well, so both are named rather than left to the fact that
+// this field is only on nodes today.
+$query->leftJoin('field_data_field_latitude', 'd', 'd.entity_id = r.entity_id AND d.entity_type = r.entity_type AND d.deleted = r.deleted');
```
anvmn, line 50: The selection keys only on a missing **latitude** (`isNull('d.entity_id')` against `field_data_field_latitude`), but the write below sets **both** fields.

Two consequences: a person currently holding a longitude but no latitude is selected, and its present longitude is overwritten with the old revision's value; and a person that lost only its longitude is never found. The docblock's "persons that hold neither now" is not what the query expresses.

Unlikely in practice, since the client nulls both together — but this script is the only writer that could create the half-state it warns about.

server/hedley/modules/custom/hedley_person/scripts/restore-wiped-gps-coordinates.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    if ($dry_run) {
+      drush_print("  Person $person_id would be restored to {$revision['field_latitude_value']}, $longitude.");
+      $total++;
+      continue;
+    }
+
+    $person = node_load($person_id);
+    if (!$person) {
```
anvmn, line 121: A failed `node_load` is skipped without printing or counting, so the closing "Restored coordinates for $total of $count persons" under-reports silently, and an operator cannot tell a failed load from a person never reached. That line is their only completion signal.

anvmn (commented):
Review findings (medium). No blocking problems; three real issues, the first of which is a hole in the client half of this fix. All verified against the code rather than inferred.

anvmn:
Pushed `71ac44bec` for the first finding.

The carry-over guarded only the branch where no reading was asked for. Since `updatePersonWithCooridnates` ends in `Maybe.withDefault person`, a ticked checkbox with no fix available fell through with the form's `Nothing` intact and erased the stored location anyway. The stored values are now carried over first, and a fresh reading is laid over them when there is one — so the two are no longer alternatives.

The other two are not being changed, by decision:

- **Selection keys on latitude only** — latitude and longitude are always written together, so the half-state the finding describes does not arise. (It does mean the script's "no longitude alongside it" skip is belt-and-braces rather than a path that fires; it reports rather than guesses, so it stays.)
- **Silent skip on a failed `node_load`** — the codebase does not check node loads elsewhere, and this is not the place to start a new convention.

Re-verified after the change: `elm make src/elm/Main.elm` compiles 548 modules, `elm-format --validate` clean, `elm-review` unchanged at one pre-existing unused import in the touched file.

client/src/elm/Backend/Update.elm
anvmn: Confirmed and fixed in `71ac44bec`.

You are right that `Maybe.withDefault person` made the true branch a pass-through: a reading that is asked for is not always a reading that arrives. The stored values are now carried over first and a fresh reading is laid over them, so the two are no longer alternatives and the ticked-checkbox-without-a-fix path keeps what is stored.

Re-verified after the change: 548 modules compile, `elm-format --validate` clean, `elm-review` unchanged.

server/hedley/modules/custom/hedley_person/scripts/restore-wiped-gps-coordinates.php
anvmn: Not changing this one, by decision: latitude and longitude are always written together, so a person holds both or neither and the half-state does not arise.

One consequence worth naming — it also makes the script's "no longitude alongside it" skip belt-and-braces rather than a branch that fires. It stays, since it reports rather than guesses, but it is there for data that arrived some other way, not for anything our client can produce.

server/hedley/modules/custom/hedley_person/scripts/restore-wiped-gps-coordinates.php
anvmn: Real, but declined: the codebase does not check the result of a node load anywhere else, and this script is not the place to introduce that convention.

client/src/elm/Backend/Update.elm
```diff
@@ -4627,13 +4627,31 @@ updateIndexedDb language currentDate currentTime coordinates zscores site featur
...
+                        |> Maybe.withDefault person
+
+                -- Adding GPS coordinates. Laid over the stored ones, rather
+                -- than offered instead of them, because a reading that was
+                -- asked for is not always a reading that arrived: without a
+                -- fix this leaves what is stored alone.
                 personWithCoordinates =
                     if gpsCoordinatesEnabled features && person.saveGPSLocation then
-                        updatePersonWithCooridnates person coordinates
+                        updatePersonWithCooridnates personWithStoredCoordinates coordinates
```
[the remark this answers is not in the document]
anvmn: Done in `2f64ab868`.

Four occurrences, all in this file — the annotation, the definition and the two call sites — so the rename is contained and the compiler verifies it. Renamed here rather than deferred, since this branch adds one of those call sites and nobody grepping for `Coordinates` was going to find the helper.

Verified after the rename: 548 modules compile, `elm-format --validate` clean, `elm-review` unchanged.

PR #2063, opened 2026-08-10
Child scoreboard: let the report cancel, and refer on diarrhea

### LINKED ISSUE #2062: Child scoreboard: ending a ChildScoreboard encounter from the progress report cannot be cancelled, and never refers a child with diarrhea
Two defects on the same screen, both from how `Pages/ChildScoreboard/ProgressReport/View.elm` fills in the shared progress-report config.

## 1. The confirmation dialog cannot be cancelled

```elm
{ showEndEncounterDialog = model.showAIEncounterPopup
, setEndEncounterDialogStateMsg = always ShowAIEncounterPopup
```

The shared dialog's Cancel button sends `setEndEncounterDialogStateMsg False` (`Pages/WellChild/ProgressReport/View.elm:635`). `always` discards that `False`, and `ShowAIEncounterPopup` sets `showAIEncounterPopup = True` unconditionally — the very flag the dialog's visibility reads. So Cancel re-asserts what opened the dialog.

The only way out is to confirm and end the encounter. A health worker who opens it to check something has no way back.

## 2. Diarrhea does not refer when the encounter is ended from the report

The encounter page routes on the NCDA answer (`Pages/ChildScoreboard/Encounter/View.elm`):

```elm
if childGotDiarrhea then showAIEncounterPopupMsg else closeEncounterMsg assembled.id
```

which warns the worker and starts an acute illness encounter. The progress report's end path never asks. Its `TriggerAcuteIllnessEncounter` handler exists (`ProgressReport/Model.elm:35`, `Update.elm:45-50`) but no view references it — which is how an unused-constructor sweep surfaced this.

So a child recorded with diarrhea is referred when the encounter is ended from the Encounter tab and silently not referred when it is ended from the progress report. Same child, same data, different clinical outcome depending on the screen.

## Scope

Checked all five bindings of that config field. Two other pages also pass `always NoOp`, but pair it with a hardcoded `showEndEncounterDialog = False`, so their dialog never opens and the message is never sent. ChildScoreboard is the only site combining `always` with a live flag. It is a one-off, not a family — though those two are one edit away from the same trap.

## Fix

Give the page a real `SetEndEncounterDialogState Bool`, and move the diarrhea question into a shared helper both end paths call, so where an encounter was ended from stops deciding whether the child is referred.

anvmn:
Fixes #2062

Two defects on one screen, both from how the ChildScoreboard progress report fills in the shared progress-report config.

## The dialog could not be cancelled

It passed `setEndEncounterDialogStateMsg = always ShowAIEncounterPopup` while reading `showEndEncounterDialog = model.showAIEncounterPopup`. The shared dialog's Cancel sends `False`; `always` discarded it and set the flag the dialog's own visibility reads, so Cancel re-opened the dialog and confirming was the only way out.

The page now has a real `SetEndEncounterDialogState Bool` and its own `showEndEncounterDialog` flag, which also lets `showAIEncounterPopup` go back to meaning the popup.

## Diarrhea did not refer from the report

Ending an encounter from the encounter page asks whether diarrhea was recorded on the NCDA questionnaire, warns the worker, and starts an acute illness encounter. Ending it from the progress report never asked — its `TriggerAcuteIllnessEncounter` handler existed but no view referenced it, which is how an unused-constructor sweep found this.

So the same child, with the same data, was referred or not depending on which screen the encounter was ended from. Both paths now ask through one helper — `childGotDiarrhea` in `Pages/ChildScoreboard/Encounter/Utils.elm` — and the report shows the same warning before it refers.

## Scope check

I checked all five bindings of `setEndEncounterDialogStateMsg` before writing anything:

| site | msg | dialog flag | status |
| --- | --- | --- | --- |
| `Nutrition/ProgressReport:71` | `SetEndEncounterDialogState` | `model.showEndEncounterDialog` | correct |
| `WellChild/ProgressReport:172` | `SetEndEncounterDialogState` | `model.showEndEncounterDialog` | correct |
| `PatientRecord/View:175` | `always NoOp` | `False` | inert |
| `ProgressReport/View:54` | `always NoOp` | `False` | inert |
| `ChildScoreboard/ProgressReport:71` | `always ShowAIEncounterPopup` | `model.showAIEncounterPopup` | **broken** |

Three sites discard the `Bool`, but two pair it with a hardcoded `False`, so their dialog never opens and the message is never sent. Only ChildScoreboard combines `always` with a live flag. A one-off, not a family — the two inert sites are left alone rather than churned, though they are one edit away from the same trap.

## Verified

`elm make src/elm/Main.elm` compiles 548 modules. `elm-format --validate` clean. `elm-test` 2,974 passed — re-run after replacing the symlinked `elm-stuff` with a real directory, since a symlinked one silently compiles the main tree and makes the pass vacuous; the generated project's `source-directories` were confirmed to resolve into the worktree. `elm-review` reports 10 findings under `ChildScoreboard`, all the pre-existing `Backend.Entities` / `Backend.Measurement.Model` unused-import pattern that also affects files this branch never touches; each of the four touched files already carried that import on develop. Three imports in `Encounter/View.elm` became unused when the inline check moved to the helper, and were removed.

One structural note: the report's view now returns `div [] [ report, viewModal popup ]` rather than the bare report, because the popup needs a sibling. The report's own root keeps its `page-report` class, and no CSS selector depends on it being a direct child.

### DIFF
--- a/client/src/elm/Pages/ChildScoreboard/ProgressReport/Update.elm
+++ b/client/src/elm/Pages/ChildScoreboard/ProgressReport/Update.elm
@@ -37,7 +37,9 @@ update msg model =
             )
 
         ShowAIEncounterPopup ->
-            ( { model | showAIEncounterPopup = True }
+            -- The confirmation dialog has done its job by the time we get
+            -- here, and two stacked modals would be one too many.
+            ( { model | showAIEncounterPopup = True, showEndEncounterDialog = False }
             , Cmd.none
             , []
             )
@@ -51,6 +53,9 @@ update msg model =
 
         SetDiagnosisMode mode ->
             ( { model | diagnosisMode = mode }, Cmd.none, [] )
+
+        SetEndEncounterDialogState isOpen ->
+            ( { model | showEndEncounterDialog = isOpen }, Cmd.none, [] )
 
         MsgReportToWhatsAppDialog subMsg ->
             let
--- a/client/src/elm/Pages/ChildScoreboard/ProgressReport/Update.elm
+++ b/client/src/elm/Pages/ChildScoreboard/ProgressReport/Update.elm
@@ -2,6 +2,7 @@ module Pages.ChildScoreboard.ProgressReport.Update exposing (update)
 
 import App.Model
 import Backend.ChildScoreboardEncounter.Model
+import Backend.Entities exposing (..)
 import Backend.IndividualEncounterParticipant.Model exposing (IndividualParticipantInitiator(..))
 import Backend.Model
 import Components.ReportToWhatsAppDialog.Model
@@ -21,13 +22,12 @@ update msg model =
             )
 
         CloseEncounter id ->
-            ( model
+            -- The dialog is cleared as well: this page model is kept per
+            -- encounter and outlives the encounter, so a flag left set here
+            -- greets whoever opens the report next.
+            ( { model | showEndEncounterDialog = False }
             , Cmd.none
-            , [ Backend.ChildScoreboardEncounter.Model.CloseChildScoreboardEncounter
-                    |> Backend.Model.MsgChildScoreboardEncounter id
-                    |> App.Model.MsgIndexedDb
-              , App.Model.SetActivePage PinCodePage
-              ]
+            , closeEncounterMsgs id ++ [ App.Model.SetActivePage PinCodePage ]
             )
 
         SetActivePage page ->
@@ -36,20 +36,23 @@ update msg model =
             , [ App.Model.SetActivePage page ]
             )
 
-        ShowAIEncounterPopup ->
-            -- The confirmation dialog has done its job by the time we get
-            -- here, and two stacked modals would be one too many.
+        ShowAIEncounterPopup id ->
+            -- The encounter is closed here, because that is what was just
+            -- confirmed. The popup that follows is the referral, not a second
+            -- chance to decide, and the dialog goes so the two do not stack.
             ( { model | showAIEncounterPopup = True, showEndEncounterDialog = False }
             , Cmd.none
-            , []
+            , closeEncounterMsgs id
             )
 
         TriggerAcuteIllnessEncounter assembled ->
+            -- The encounter was closed when the popup opened, so this only
+            -- navigates. Going through CloseEncounter would append its own
+            -- SetActivePage PinCodePage after this one, and the last one wins.
             ( { model | showAIEncounterPopup = False }
             , Cmd.none
             , [ App.Model.SetActivePage <| UserPage (AcuteIllnessParticipantPage InitiatorParticipantsPage assembled.participant.person) ]
             )
-                |> sequenceExtra update [ CloseEncounter assembled.id ]
 
         SetDiagnosisMode mode ->
             ( { model | diagnosisMode = mode }, Cmd.none, [] )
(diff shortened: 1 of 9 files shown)


client/src/elm/Pages/ChildScoreboard/ProgressReport/Update.elm
```diff
@@ -37,7 +37,9 @@ update msg model =
             )
 
         ShowAIEncounterPopup ->
-            ( { model | showAIEncounterPopup = True }
+            -- The confirmation dialog has done its job by the time we get
+            -- here, and two stacked modals would be one too many.
+            ( { model | showAIEncounterPopup = True, showEndEncounterDialog = False }
```
anvmn, line 43: **The referral navigates to the PIN code page, not the acute illness encounter.**

`TriggerAcuteIllnessEncounter` returns `[SetActivePage AcuteIllnessParticipantPage]` and then `sequenceExtra update [ CloseEncounter assembled.id ]`. `sequenceExtra` concatenates with `msgsSoFar ++ newMsgs` (`Gizra/Update.elm:64`), and `CloseEncounter` contributes `SetActivePage PinCodePage`, so the list is:

```
[ SetActivePage AcuteIllnessParticipantPage
, MsgIndexedDb CloseChildScoreboardEncounter
, SetActivePage PinCodePage
]
```

`App.Update.updateConfigured` folds that in order with `sequence update`, and each `SetActivePage` assigns `model.activePage`. Last one wins: **PinCodePage**.

So a nurse who records diarrhea, ends the encounter from the report and clicks Continue closes the encounter and lands on the PIN screen. The warning promises a referral that does not happen.

This branch is what makes the path reachable from the report, but the same shape is already at `Pages/ChildScoreboard/Encounter/Update.elm:40` — meaning the referral has never worked on either screen. `Pages/WellChild/Encounter/Update.elm:44` shows the correct shape: it emits the navigation itself and never routes through a msg that navigates elsewhere.

client/src/elm/Pages/ChildScoreboard/ProgressReport/Update.elm
```diff
@@ -52,6 +54,9 @@ update msg model =
         SetDiagnosisMode mode ->
             ( { model | diagnosisMode = mode }, Cmd.none, [] )
 
+        SetEndEncounterDialogState isOpen ->
```
anvmn, line 60: `CloseEncounter` does not reset the new `showEndEncounterDialog` flag. The page model lives in `App.Model.childScoreboardReportPages`, keyed by encounter id, and is never cleared on navigation — so ending the encounter from the report leaves the flag `True`, and re-opening that report in the same session shows the "End encounter?" dialog on arrival, over an already-closed encounter.

Both siblings reset it: `Nutrition/ProgressReport/Update.elm:20` and `WellChild/ProgressReport/Update.elm:20`.

anvmn (commented):
Review findings (medium). The first one matters: the referral this PR wires up does not actually arrive at the acute illness encounter, and that defect turns out to be pre-existing on the encounter page too — so the diarrhea referral has never worked anywhere. Traced rather than inferred; details inline.

anvmn:
Third finding, posted here because its line sits outside the diff — `ProgressReport/View.elm:72`, the `closeEncounterMsg` branch.

When diarrhea is recorded, confirming the end-encounter dialog closes nothing. The user answers "End encounter?" / "Once you end the encounter…", and the only effect is that the warning popup opens. That popup offers Continue and no dismiss, so a user who instead navigates back via the header has confirmed ending the encounter while it stays open.

Behaviourally this matches the encounter page, but there no confirmation step exists — so only the report makes the confirm-then-nothing-happened state reachable. Low severity, and it disappears if the referral is fixed to close and navigate in one step.

PR #2065, opened 2026-08-10
Dashboard stats: stop counting an absence as a session attended

### LINKED ISSUE #2064: Dashboard stats: missed-session statistics count an absence as attendance
## Problem

`hedley_activity_get_persons_sessions_with_measurements()` (`hedley_activity.module:463`) selects published, non-deleted nodes that carry both `field_person` and `field_session`. There is no bundle condition, so the `attendance` bundle qualifies — it carries both fields.

`hedley_stats_get_session_attendance_stats_by_health_center()` (`hedley_stats.module:2838-2851`) reads that result as attendance:

```php
if (isset($persons_sessions[$child_id][$session_id])) { $attended = TRUE; break; }
...
if (!$attended) { $missed_session[] = $expected_person; }
```

An attendance node exists whether or not the person came: un-checking a checked-in mother dispatches `SetCheckedIn attendanceId motherId False`, which **updates** the node to `attended = FALSE` rather than deleting it. So the record stating that a child was absent is what proves they were present, and the session is not counted as missed.

## Scale

On ihangane.live: 187,676 attendance nodes, of which **6,748 record `field_attended = 0`**. Each is a session a child is recorded as having missed that the statistics count as attended.

## What should count

A session counts for a person when a measurement was taken there. The attendance record is not a measurement — it says only whether the person was expected. That rules out two cases the current query gets wrong:

- marked absent, nothing measured — counted today, should not be
- marked present, nothing measured — counted today, should not be either, because the session did not actually run for that child

A child with any real measurement in the session still counts, whatever the attendance record says.

## Fix

Exclude the `attendance` bundle from the query. The function's name then describes what it does, and the caller's use of it as "attended" follows from the rule above.

Note this moves published statistics: missed sessions will rise at every health centre once deployed. It is a correction, but it will look like a regression on a dashboard.

anvmn:
Fixes #2064

The query behind the missed-session statistics selected any published node carrying both `field_person` and `field_session`, with no bundle condition. The `attendance` record carries both — and it exists whether or not the person came, because marking someone absent **updates** it to `attended = FALSE` rather than deleting it.

So the record stating that a child was absent was the thing proving they attended, and that session was never counted as missed. On ihangane.live, **6,748 of 187,676 attendance records say `attended = FALSE`**.

## What counts as attending

A measurement means the person was there — and so does the attendance record, but only when it says so. Its value, not its existence, answers the question:

| attended | measurements | before | after |
| --- | --- | --- | --- |
| yes | yes | counted | counted |
| no | yes | counted | counted — someone was clearly there |
| yes | no | counted | counted — the record says they came |
| no | no | **counted** (the bug) | not counted |

```php
hedley_general_join_field_to_query($query, 'node', 'field_attended', FALSE);
$query->condition(
  db_or()
    ->condition('node.type', HEDLEY_ACTIVITY_ATTENDANCE_CONTENT_TYPE, '<>')
    ->condition('field_attended.field_attended_value', 1)
);
```

The join is optional because only the `attendance` bundle carries that field, so every other measurement passes on the first condition.

## Why not exclude attendance altogether

The first pass did exactly that, on the reasoning that a session where nothing was measured did not really run. Review found it contradicts how the codebase already counts group attendance: `hedley_stats_get_total_encounters_by_patients_at_clinic()` (`hedley_stats.module:3004-3012`) counts attendance with `attended = TRUE`, commented *"Only participants who actually attended the sessions."* The same child would have been one FBF encounter there and a missed session here.

Measured on the 50 most recent sessions on ihangane.live: **141** attendance records say `attended = 1`, and **39 of them (28%)** have no measurement for that person in that session. The common case, not the corner one — so the two statistics now agree instead.

## Test

The existing coverage was a smoke check that computed `$count_missed_sessions` and `$count_completed_programs` and asserted on **neither** — the semantics were entirely uncovered, which is how this survived. Rather than add a test class (each costs a full install in CI), assertions are folded into `testStatsCalculations`: one session, three children — weighed, ticked with nothing taken, marked absent. The first two count, the third does not. The third assertion fails on develop.

## Verification status

`php -l` clean; phpcs `Drupal` and `DrupalPractice` with the full CI extension list both exit 0 with no output.

**The SimpleTest has not been run locally** — the DDEV router is unhealthy on this machine, so `ddev simpletest` cannot start. CI's `test_simpletest_linux` job is the first execution of these assertions.

## Known gap, not taken here

The new assertions exercise the helper, not `hedley_stats_get_session_attendance_stats_by_health_center()` where the defect actually surfaced, so a regression one level up would still pass. Writing that blind — without being able to run it — risks a red CI cycle on period arithmetic, so it is raised rather than guessed at.

### DIFF
--- a/server/hedley/modules/custom/hedley_activity/hedley_activity.module
+++ b/server/hedley/modules/custom/hedley_activity/hedley_activity.module
@@ -489,8 +489,17 @@ function hedley_activity_get_persons_sessions_with_measurements(array $sessions_
   // the record itself turned an absence into proof of attendance.
   //
   // The join is optional because only the attendance bundle carries the
-  // field; every other measurement passes on the first condition.
-  hedley_general_join_field_to_query($query, 'node', 'field_attended', FALSE);
+  // field; every other measurement passes on the first condition. It is made
+  // by hand rather than through the usual helper, which would also add the
+  // value to the SELECT list and so to the DISTINCT below - a person with
+  // both a measurement and an attendance record would come back as two rows
+  // where one will do, in the query that exists to be the cheap bulk path.
+  $query->leftJoin(
+    'field_data_field_attended',
+    'field_attended',
+    'node.nid = field_attended.entity_id AND field_attended.entity_type = :attended_entity_type',
+    [':attended_entity_type' => 'node']
+  );
   $query->condition(
     db_or()
       ->condition('node.type', HEDLEY_ACTIVITY_ATTENDANCE_CONTENT_TYPE, '<>')
--- a/server/hedley/modules/custom/hedley_activity/hedley_activity.module
+++ b/server/hedley/modules/custom/hedley_activity/hedley_activity.module
@@ -477,6 +477,15 @@ function hedley_activity_get_persons_sessions_with_measurements(array $sessions_
   $session_field_name = $session_field . '.' . $session_field . '_target_id';
   $query->condition($session_field_name, $sessions_ids, 'IN');
 
+  // A session counts for a person when something was measured there. The
+  // attendance record does not count: it carries a person and a session like
+  // any measurement does, but it says only whether the person was expected to
+  // come, and it exists either way - marking someone absent updates it to
+  // attended = FALSE rather than removing it. So counting it made a record of
+  // absence into proof of presence, and a tick with nothing behind it into a
+  // session that ran.
+  $query->condition('node.type', HEDLEY_ACTIVITY_ATTENDANCE_CONTENT_TYPE, '<>');
+
   $query->distinct();
 
   $result = $query->execute();
--- a/server/hedley/modules/custom/hedley_activity/hedley_activity.module
+++ b/server/hedley/modules/custom/hedley_activity/hedley_activity.module
@@ -452,8 +452,7 @@ function hedley_activity_node_by_peer_in_session(\EntityMetadataWrapper $wrapper
  *
  * Attending is evidenced either by a measurement taken during the session, or
  * by an attendance record that says they came. An attendance record saying
- * they did not is no evidence at all, which is what the name of this function
- * fails to convey - it predates the distinction.
+ * they did not is no evidence at all.
  *
  * Resolves attendance for many persons at once with a single query, rather
  * than a COUNT query per person, so callers can determine it in memory.
@@ -465,7 +464,7 @@ function hedley_activity_node_by_peer_in_session(\EntityMetadataWrapper $wrapper
  *   A map keyed by person node ID; each value is a set of session node IDs (as
  *   array keys) that person attended.
  */
-function hedley_activity_get_persons_sessions_with_measurements(array $sessions_ids) {
+function hedley_activity_get_persons_sessions_attended(array $sessions_ids) {
   $persons_sessions = [];
   if (empty($sessions_ids)) {
     return $persons_sessions;
--- a/server/hedley/modules/custom/hedley_activity/hedley_activity.module
+++ b/server/hedley/modules/custom/hedley_activity/hedley_activity.module
@@ -488,18 +488,13 @@ function hedley_activity_get_persons_sessions_with_measurements(array $sessions_
   // and its value, not its presence, is what answers the question. Counting
   // the record itself turned an absence into proof of attendance.
   //
-  // The join is optional because only the attendance bundle carries the
-  // field; every other measurement passes on the first condition. It is made
-  // by hand rather than through the usual helper, which would also add the
-  // value to the SELECT list and so to the DISTINCT below - a person with
-  // both a measurement and an attendance record would come back as two rows
-  // where one will do, in the query that exists to be the cheap bulk path.
-  $query->leftJoin(
-    'field_data_field_attended',
-    'field_attended',
-    'node.nid = field_attended.entity_id AND field_attended.entity_type = :attended_entity_type',
-    [':attended_entity_type' => 'node']
-  );
+  // The join is left, because only the attendance bundle carries the field;
+  // every other measurement passes on the first condition. It is made by hand
+  // rather than through the usual helper, which would also add the value to
+  // the SELECT list and so to the DISTINCT below - a person with both a
+  // measurement and an attendance record would come back as two rows where
+  // one will do, in the query that exists to be the cheap bulk path.
+  $query->leftJoin('field_data_field_attended', 'field_attended', 'node.nid = field_attended.entity_id');
   $query->condition(
     db_or()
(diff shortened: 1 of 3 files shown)


server/hedley/modules/custom/hedley_activity/hedley_activity.module
```diff
@@ -477,6 +477,15 @@ function hedley_activity_get_persons_sessions_with_measurements(array $sessions_
...
   $query->condition($session_field_name, $sessions_ids, 'IN');
 
+  // A session counts for a person when something was measured there. The
+  // attendance record does not count: it carries a person and a session like
+  // any measurement does, but it says only whether the person was expected to
+  // come, and it exists either way - marking someone absent updates it to
+  // attended = FALSE rather than removing it. So counting it made a record of
+  // absence into proof of presence, and a tick with nothing behind it into a
+  // session that ran.
+  $query->condition('node.type', HEDLEY_ACTIVITY_ATTENDANCE_CONTENT_TYPE, '<>');
```
anvmn: **This contradicts how the codebase counts attendance elsewhere, and the overlap is not rare.**

`hedley_stats_get_total_encounters_by_patients_at_clinic()` (`hedley_stats.module:3004-3012`) selects `type = attendance` with `field_attended = TRUE`, commented *"Only participants who actually attended the sessions."*

So after this change, a child ticked present with nothing measured is **one FBF encounter** in Total Encounters and **a missed session** on the attendance card, at the same time.

Measured on ihangane.live, over the 50 most recent sessions:

| | |
| --- | --- |
| attendance records with `attended = 1` | 141 |
| of those, with **no** measurement for that person in that session | **39 (28%)** |

So it is a common case. The narrow fix for the reported bug alone would keep attendance rows whose `field_attended` is TRUE.

This is deliberate — the rule chosen was that a session counts only when something was measured, so a tick with nothing behind it is not a session that ran. Recording the consequence here so it is a decision on the record rather than a surprise on a dashboard.

server/hedley/modules/custom/hedley_activity/hedley_activity.module
```diff
@@ -477,6 +477,15 @@ function hedley_activity_get_persons_sessions_with_measurements(array $sessions_
   $session_field_name = $session_field . '.' . $session_field . '_target_id';
   $query->condition($session_field_name, $sessions_ids, 'IN');
 
+  // A session counts for a person when something was measured there. The
+  // attendance record does not count: it carries a person and a session like
+  // any measurement does, but it says only whether the person was expected to
```
anvmn: The comment says the attendance record *"says only whether the person was expected to come"*. That is wrong: `field_attended` records whether they **came** — which is exactly why the record's value, not its existence, is what the original bug turned on. Worth rewording whichever way the finding above is decided.

anvmn (commented):
Review findings (medium). The first one is a decision, not a defect: it shows this change puts the missed-session statistic in contradiction with the total-encounters statistic, and I have measured how often that bites. Holding for the maintainer's call rather than acting on it.

anvmn:
Two test findings, posted here because their lines sit outside the diff hunks.

**`HedleyStatsCalculation.test:193` — the new test covers the helper, not the statistic.** Test 3b asserts on `hedley_activity_get_persons_sessions_with_measurements()` directly and creates no PMTCT participant, so `hedley_stats_get_session_attendance_stats_by_health_center()` — where an absent child actually surfaced as attended — is never called. A regression reintroduced one level up, in the `isset($persons_sessions[...])` loop or in a future caller, would still pass. Part 3 already builds the scaffolding an end-to-end assertion needs.

**`HedleyStatsCalculation.test:142` — the pre-existing smoke check asserts on nothing.** It computes `$count_missed_sessions` and `$count_completed_programs` at lines 155-168 and then part 4 begins. Those attendance-only children flip from attended to missed under this change and the suite stays silent. Pre-existing, but this PR's docblock now advertises part 3 as covering what counts as attending, so it is the natural place to close it — and doing so would make the contradiction in the first finding visible in the test output rather than only on a dashboard.

server/hedley/modules/custom/hedley_activity/hedley_activity.module
anvmn: Taken in `172fcef71` — narrowed to `field_attended = TRUE`.

The query now keeps an attendance record when it says the person came, and drops it otherwise:

```php
hedley_general_join_field_to_query($query, 'node', 'field_attended', FALSE);
$query->condition(
  db_or()
    ->condition('node.type', HEDLEY_ACTIVITY_ATTENDANCE_CONTENT_TYPE, '<>')
    ->condition('field_attended.field_attended_value', 1)
);
```

The join is optional because only the `attendance` bundle carries that field — every other measurement passes on the first condition.

That fixes the reported defect (the 6,748 absences) and leaves this statistic agreeing with `hedley_stats_get_total_encounters_by_patients_at_clinic()`, so the 28% of ticked-but-unmeasured attendances stay counted in both rather than counted in one and missed in the other.

server/hedley/modules/custom/hedley_activity/hedley_activity.module
anvmn: Fixed in the same commit. The comment now says the record reports whether they came, and that its value rather than its existence is what answers the question — which is the distinction the bug turned on.

The function's docblock says the same, and adds that an attendance record saying they did not come is no evidence at all.

anvmn:
Second review pass — three low findings, two taken in `e7be40da9`, one refuted by data.

**Taken — the calling-side comment contradicted the shipped rule.** It still described what the *first* commit did ("a session where nothing was taken did not run for that person, whatever the attendance record says"), which the narrowing replaced. It now says what the code does and matches the docblock. Good catch: a comment asserting the opposite of the query is worse than none.

**Taken — the join widened the DISTINCT.** `hedley_general_join_field_to_query()` also does `addField(...)`, so `field_attended` joined the SELECT list and therefore the `distinct()`. A person with both a measurement and an attendance record for one session came back as two rows instead of one — harmless to the result, which keys into a set, but wasteful in the query that exists to be the cheap bulk path. Now joined by hand, so the DISTINCT stays over person and session.

**Refuted — attendance rows with no `field_attended`.** The field instance is `required = 0`, so the concern was that such rows would flip from attended to missed. Measured on every site:

| site | attendance nodes | with a `field_attended` row |
| --- | --- | --- |
| ihangane | 187,676 | **187,676** |
| vhw | 75 | **75** |
| tip-somalia | 0 | 0 |

No such row exists anywhere, so nothing shifts and there is nothing to defend against. Adding a NULL-means-attended arm would be guarding a case the data does not contain.

PR #2067, opened 2026-08-11
CI: stop passing when the tests never ran

### LINKED ISSUE #2066: CI: CI reports success when no tests ran
## Problem

`.ddev/commands/web/simpletest` cannot fail for a run that never happened:

```bash
php ./www/scripts/run-tests.sh ... 2>&1 | tee /tmp/simpletest-result.txt

grep -E -i "([1-9]+ fail)|(Fatal error)|..." /tmp/simpletest-result.txt && exit 1

drush pm-disable simpletest -y

exit 0
```

The runner's exit code dies in the pipe to `tee`; the wrapper has no `set -e` and no `pipefail`; `drush en simpletest -y` is unchecked; and the only gate is a grep for failure text followed by an unconditional `exit 0`. Nothing distinguishes **no failures** from **no tests**.

The job is green when the runner segfaults, when the runner is missing, when the simpletest module is not enabled, when the output is empty, and when `drush en` fails.

`test_simpletest_linux` is the only server-side test job in CI, gating the pending release. A broken site build leaves it green on every pull request.

`ci-scripts/install_drupal.sh:11` is how a build gets broken without anyone noticing: `ddev restart || ddev logs` under `set -e` means a failed restart followed by a succeeding `ddev logs` leaves the step green, and the whole Drupal install runs as ddev post-start hooks.

Two more gates that cannot fail:

- `client/playwright.config.ts` has no `forbidOnly`, so a committed `test.only` shrinks an e2e shard to one test and the job stays green.
- The three `store_artifacts` steps in `.circleci/config.yml` default to `when: on_success`, so failure screenshots are dropped by the run that produced them.

## What is not broken

For runs that complete, the grep does its job — 3 fails, 0 fails with 2 exceptions, a `Fatal error`, and trailing-zero counts like 10 all turn the job red. The hole is exclusively the never-ran case.

anvmn:
Fixes #2066

The simpletest wrapper fails when the suite does not run. `set -o pipefail` carries the runner's exit code out of the pipe to `tee`, that status is checked, `drush en simpletest -y` is checked, and a run reports at least one pass before the script exits 0. The failure grep is unchanged — it was already correct for runs that complete — and the module is disabled from a trap, so a failing run leaves the site as it found it.

A failed `ddev restart` fails `ci-scripts/install_drupal.sh` rather than passing through the `ddev logs` fallback.

`forbidOnly: !!process.env.CI` stops a committed `test.only` from shrinking an e2e shard while the job stays green.

The three `store_artifacts` steps carry `when: always`, so failure screenshots survive the run that produced them. Note that `trace: 'on-first-retry'` with `retries: 0` means traces are never captured, so this covers screenshots only.

## Not in this PR

Three of the issue's smaller gates, each of which can turn CI red for pre-existing reasons and wants its own diff:

- `lint_elm_review` installs registry-latest `elm-review` at run time instead of the pinned version.
- `test_shell.sh` checks neither the extensionless install and deploy chain nor any `.ddev/commands/*`, including the wrapper this PR fixes.
- Nothing rebuilds the committed `server/elm` bundle and compares it against source.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,6 +145,9 @@ jobs:
           no_output_timeout: 15m
       - store_artifacts:
           path: client/test-results
+          # Screenshots are captured only on failure, so the default
+          # on_success would drop them exactly when they are needed.
+          when: always
           destination: playwright-results
   e2e_playwright_2:
     machine:
@@ -199,6 +202,9 @@ jobs:
           no_output_timeout: 15m
       - store_artifacts:
           path: client/test-results
+          # Screenshots are captured only on failure, so the default
+          # on_success would drop them exactly when they are needed.
+          when: always
           destination: playwright-results
   e2e_reporting:
     machine:
@@ -253,4 +259,7 @@ jobs:
           no_output_timeout: 15m
       - store_artifacts:
           path: client/test-results
+          # Screenshots are captured only on failure, so the default
+          # on_success would drop them exactly when they are needed.
+          when: always
           destination: playwright-results
--- a/.ddev/commands/web/simpletest
+++ b/.ddev/commands/web/simpletest
@@ -4,11 +4,37 @@
 ## Usage: simpletest
 ## Example: "ddev simpletest"
 #
-drush en simpletest -y
+
+# The runner's exit code reaches us through the pipe into tee, and every step
+# below is checked. Without this the job reported success for a runner that
+# never started - a broken site build left the only server-side test job green
+# on every pull request, with nothing to distinguish no failures from no tests.
+set -o pipefail
+
+if ! drush en simpletest -y; then
+  echo "Could not enable the simpletest module, so no tests were run."
+  exit 1
+fi
+
 cd ..
 php ./www/scripts/run-tests.sh --php "$(which php)" --concurrency 4 --verbose --color --url http://eheza-app.ddev.site:8081 Hedley 2>&1 | tee /tmp/simpletest-result.txt
+runner_status=$?
 
+if [ "$runner_status" -ne 0 ]; then
+  echo "The test runner exited with status $runner_status."
+  exit 1
+fi
+
+# Reports a failure, a fatal, or an exception. The trailing-zero alternatives
+# catch counts like 10 and 20, which the leading [1-9] would otherwise miss.
 grep -E -i "([1-9]+ fail)|(Fatal error)|([1-9]+ exception)|([0-9]+0 fail)|([0-9]+0 exception)" /tmp/simpletest-result.txt && exit 1
+
+# Says nothing failed, which is also what it says when nothing ran. A suite
+# that passed has passes to show for it.
+if ! grep -E -q "[1-9][0-9]* passes" /tmp/simpletest-result.txt; then
+  echo "No passing tests were reported, so the suite did not run."
+  exit 1
+fi
 
 drush pm-disable simpletest -y
 
--- a/.ddev/commands/web/simpletest
+++ b/.ddev/commands/web/simpletest
@@ -16,6 +16,10 @@ if ! drush en simpletest -y; then
   exit 1
 fi
 
+# Every path below can exit, and leaving the module enabled changes the state
+# of the developer's site.
+trap 'drush pm-disable simpletest -y' EXIT
+
 cd ..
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -145,6 +145,9 @@ jobs:
           no_output_timeout: 15m
       - store_artifacts:
           path: client/test-results
+          # Screenshots are captured only on failure, so the default
+          # on_success would drop them exactly when they are needed.
```
anvmn, line 149: **Half refuted by running it, half still unproven — worth knowing which is which.**

The concern was that `when` is not a documented key of `store_artifacts` (the docs list only `path` and `destination`), so config processing might reject it and **fail every pipeline**. It does not: CircleCI accepted this commit's config and spawned all ten jobs. That branch is dead.

What remains unproven is the other branch — whether `when: always` actually changes the behaviour or is silently ignored, in which case the step is still skipped after a failing `Run E2E tests` and the screenshots are still lost. A green run cannot distinguish those.

Proving it needs one deliberately failing e2e run, which costs ~25 minutes of CI and a temporary broken test. I have not done that. If it turns out to be a no-op the fallback is a `run` step with `when: always` that copies the results aside, which is documented for `run`.

Separately: `trace: 'on-first-retry'` with `retries: 0` means traces are never captured at all, so at best this recovers screenshots.

.ddev/commands/web/simpletest
```diff
@@ -4,12 +4,42 @@
...
 grep -E -i "([1-9]+ fail)|(Fatal error)|([1-9]+ exception)|([0-9]+0 fail)|([0-9]+0 exception)" /tmp/simpletest-result.txt && exit 1
 
-drush pm-disable simpletest -y
+# Says nothing failed, which is also what it says when nothing ran. A suite
+# that passed has passes to show for it. The singular is spelled differently -
+# simpletest formats the count with format_plural, so exactly one reads
+# "1 pass" - and a run of a single narrow class must not look like no run.
+if ! grep -E -q "[1-9][0-9]* pass(es)?" /tmp/simpletest-result.txt; then
+  echo "No passing tests were reported, so the suite did not run."
+  exit 1
```
anvmn, line 42: Taken in `a74cf3e12`, and this one was a real bug in my fix.

`_simpletest_format_summary_line()` builds the count with `format_plural($n, '1 pass', '@count passes')` (`simpletest.module:105`), so a run of a single class prints **"1 pass"** — no "passes" anywhere. My gate would have failed CI for a suite that had succeeded.

Now `[1-9][0-9]* pass(es)?`, and the harness carries the case:

```
SINGLE class, 1 pass (singular)          exit=0
```

Also taken: the module is now disabled from a `trap ... EXIT` rather than a final line, so the newly-added exit paths leave a developer's site as they found it.

anvmn (commented):
Review findings (medium), five of them. Three taken, one reverted as a genuine regression I introduced, and one settled by running it. Detail inline and below.

anvmn:
Two more findings, both outside the diff hunks.

**`ci-scripts/install_ddev.sh:21` — reverted, and the reviewer was right that I broke it.** `docker network create ddev_default || { ddev logs; exit 1; }` turns a benign, idempotent condition into a hard failure: if the network already exists — a re-run on the same machine, an SSH rerun, or a future ddev that creates it itself — `docker network create` exits non-zero and the install now dies on a healthy environment. That fallback was tolerating an expected condition, not hiding a failure, which is the opposite of the `install_drupal.sh` case this PR is actually about.

It was also scope creep. B-157 called that line dormant, and it was; I changed it for symmetry and introduced a regression doing so. Reverted in `a74cf3e12`.

**The gate proves *a* class passed, not that the *expected suite* ran — accepted, not taken.** `run-tests.sh` builds its list from discovered tests, so if a `.test` file drops out of a module's `.info` the Hedley group silently shrinks: one class passes, the runner exits 0, job green. That is a weaker version of the same ambiguity this PR closes, and the reviewer is right that a minimum expected count would close it properly.

I have not taken it because any threshold is a judgement call — a hard floor goes stale as the suite grows, and comparing summary lines against discovered `.test` classes is a bigger change than the rest of this diff put together. Worth its own item rather than a number invented here.

Verified as correct by the same pass, for the record: `runner_status=$?` does capture the pipeline status under `set -o pipefail`; `run-tests.sh` propagates child failures via `exit($status)`; `circleci_wait.sh` forwards exit codes so the new failures reach CircleCI; `SC2181` is excluded in `test_shell.sh` and `.ddev/commands` is unlinted anyway; and `forbidOnly: !!process.env.CI` works because CircleCI sets `CI=true`.

PR #2071, opened 2026-08-11
Reports: stop marking a visit incomplete when the medication was given on schedule

### LINKED ISSUE #2070: Reports: completion report counts a medication as missed when it was given within the last six months
Vitamin A, mebendezole and albendazole are given at six-month intervals. The Completion report expects them at every well-child encounter regardless of when they were last given, so a visit is marked incomplete for a child whose medication schedule was followed.

The app itself applies the interval correctly — `expectMedicationTask` (`client/src/elm/Pages/WellChild/Activity/Utils.elm:1344-1360`) does not offer the task within six months of an administration, so the nurse cannot record it. The report then counts its absence against her. "Pediatric Care" completion is therefore lowest for exactly the children whose protocol was followed.

The report searches earlier encounters for the last administration, finds it, and then asks that encounter for its date:

```php
$administration_note = $measurement->field_administration_note[LANGUAGE_NONE][0]['value'];
if ($administration_note == 'administered-today') {
  $last_administration_date_obj = $previous_encounter_data['start_date_obj'];
  break;
}
```

The earlier encounters carry `encounter`, `encounter_type` and `measurements_by_type` (`hedley_reports.module:4811-4815`) and nothing else. `start_date_obj` is added afterwards, in a `foreach ($encounters_data as $index => $encounter_data)` that holds a copy, so it never reaches the array that `array_slice()` reads from. The key is missing and the lookup yields NULL.

The branch below reads that NULL as "never administered", though the loop above has just proved otherwise:

```php
if (is_null($last_administration_date_obj)) {
  // Medication was not administered previously - add it as expected.
  $expected[] = $expected_activity;
}
```

This is the only place in the module that reads such a key from an earlier encounter; every other read is of `measurements_by_type`, `encounter` or `encounter_type`, which are present.

anvmn:
Fixes #2070

A well-child encounter carries its own `start_date_obj` from the moment it is built, so the search through earlier encounters can read the date of the one that administered a medication. The six-month comparison then matches the interval the app applies in `expectMedicationTask`, and a visit is no longer counted incomplete for a child whose medication schedule was followed.

Adding six months works on a copy of that date, so the encounter's own date stays where it belongs.

The two other completion functions that share this loop shape — child scoreboard and NCD — never read a date from an earlier encounter, so they are unaffected.

### DIFF
--- a/server/hedley/modules/custom/hedley_reports/hedley_reports.module
+++ b/server/hedley/modules/custom/hedley_reports/hedley_reports.module
@@ -4808,10 +4808,13 @@ function hedley_reports_generate_completion_data_for_well_child($participant, $e
       $measurements_by_type[$measurement->type] = $measurement;
     }
 
+    $start_date = explode(' ', $encounter->field_scheduled_date[LANGUAGE_NONE][0]['value'])[0];
+
     $encounters_data[] = [
       'encounter' => $encounter,
       'encounter_type' => $encounter_type,
       'measurements_by_type' => $measurements_by_type,
+      'start_date_obj' => new DateTime($start_date),
     ];
   }
 
@@ -4824,10 +4827,7 @@ function hedley_reports_generate_completion_data_for_well_child($participant, $e
     $encounter = $encounter_data['encounter'];
     $measurements_by_type = $encounter_data['measurements_by_type'];
 
-    // Resolve encounter start date.
-    $start_date = explode(' ', $encounter->field_scheduled_date[LANGUAGE_NONE][0]['value'])[0];
-    $start_date_obj = new DateTime($start_date);
-    $encounter_data['start_date_obj'] = $start_date_obj;
+    $start_date_obj = $encounter_data['start_date_obj'];
 
     // Try to resolve age in months at a time encounter was performed.
     $encounter_data['age_in_months'] = hedley_reports_resolve_age_in_months_for_individual_encounter($encounter, $start_date_obj);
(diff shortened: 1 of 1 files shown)


server/hedley/modules/custom/hedley_reports/hedley_reports.module
```diff
@@ -4848,7 +4849,7 @@ function hedley_reports_generate_completion_data_for_well_child($participant, $e
     hedley_reports_well_child_add_next_steps_activities($encounter_data, $birth_date_obj, $gender, $expected);
 
     $completion_data = [
-      'start_date' => $start_date,
+      'start_date' => $encounter_data['start_date'],
```
anvmn, line 4852: **Confirmed — this branch broke the saved start date, and the fix is in `1b9d1102b`.**

Moving `$start_date` into the first loop left the second loop reading whatever value survived it. Encounters load sorted ascending, so every encounter's `field_reports_data` would have been stamped with the newest encounter's date, silently, and rewritten that way on the next recompute.

Demonstrated on the loop shape alone — three encounters on 2025-01-10, 2025-04-15 and 2025-10-02:

```
date NOT on the card:  2025-10-02, 2025-10-02, 2025-10-02
date ON the card:      2025-01-10, 2025-04-15, 2025-10-02
```

The encounter now carries `start_date` beside `start_date_obj`, and this line reads it from the encounter rather than from a variable the two loops share.

That is the second time on this branch that an edit compiled and did the wrong thing, both times because a text anchor matched in more than one completion function. Both remaining edits were applied by line number with the target line asserted first.

anvmn (commented):
Review found a regression this branch introduced, and it was right. Fixed in `1b9d1102b`.

PR #2075, opened 2026-08-12
Vaccination: stop the date picker coming up empty when a vaccine becomes due today

### LINKED ISSUE #2074: Vaccination: the date picker is empty when a vaccine becomes due today
On the day a dose first becomes due, the immunisation form asks whether the vaccine was administered previously, and the date picker behind that question is unusable: the month list is empty and every day is disabled. The nurse cannot pick a date, and the dose is recorded as administered previously, dated today, with no administered-today note.

The question appears from the due date onward, since it is hidden only while the due date is later than today (`client/src/elm/Measurement/Utils.elm:1450`):

```elm
if Date.compare expectedOnDate currentDate == GT then
```

The picker's range runs from the dose's due date to yesterday (`:1560-1566`):

```elm
{ dateFrom = dateFrom                    -- the dose's due date
, dateTo = Date.add Days -1 currentDate  -- yesterday
, dateDefault = Just dateFrom
}
```

When the due date is today, that range starts after it ends. `DateSelector.Selector` renders such a range as an empty month list with every day cell disabled.

The default is not inert. Opening the picker writes it into the form (`Pages/WellChild/Activity/Update.elm:900`, and the same in ChildScoreboard and Prenatal):

```elm
defaultSelection =
    Maybe.Extra.or form.vaccinationUpdateDate (Maybe.andThen .dateDefault state)
```

That path does not pass through the widget's `Date.clamp`, which only guards dates the widget emits. Save requires nothing more than a date being present, so the dose saves with today's date through the path meant for doses given earlier.

All three immunisation programs build their picker through the same function, so all three behave this way. The trigger is an ordinary calendar coincidence: a child reaching eligibility age today, or a last dose plus its interval landing exactly today.

anvmn:
Fixes #2074

The date picker for a dose administered previously runs from the day the dose became due to yesterday. A dose that becomes due today was not due on any earlier day, so the range is that one day:

```elm
dateTo =
    Date.max (Date.add Days -1 currentDate) dateFrom
```

The calendar has a day to show, and the date the form takes when the picker opens is the day the dose became due.

Ending the range at yesterday would offer a day the dose could not have been given on — and for a first dose of BCG or OPV, whose due date is the birth date, a day before the child was born.

Ranges that already made sense are unchanged: a dose due last month still offers from that date to yesterday, and one due yesterday still offers that single day.

All three immunisation programs — WellChild, ChildScoreboard and Prenatal — build this picker through the same function.

## Not in this PR

Opening the picker writes its default into the form before the nurse chooses anything. That applies to every vaccine rather than to this case, and changing it would alter the interaction well beyond the defect.

### DIFF
--- a/client/src/elm/Measurement/Utils.elm
+++ b/client/src/elm/Measurement/Utils.elm
@@ -1558,12 +1558,22 @@ vaccinationFormDynamicContentAndTasks language currentDate site config vaccineTy
                                 -- expected due date of first dose.
                                 |> Maybe.withDefault config.firstDoseExpectedFrom
 
+                        -- A dose given previously was given before today, so the
+                        -- range ends yesterday. On the day a dose first becomes due
+                        -- its due date is later than that, and the range starts from
+                        -- the last day it can offer instead.
+                        dateTo =
+                            Date.add Days -1 currentDate
+
+                        rangeStart =
+                            Date.min dateFrom dateTo
+
                         dateSelectorConfig =
                             { select = config.setVaccinationUpdateDateMsg
                             , close = config.setVaccinationUpdateDateSelectorStateMsg Nothing
-                            , dateFrom = dateFrom
-                            , dateTo = Date.add Days -1 currentDate
-                            , dateDefault = Just dateFrom
+                            , dateFrom = rangeStart
+                            , dateTo = dateTo
+                            , dateDefault = Just rangeStart
                             }
                     in
                     ( [ viewLabel language Translate.SelectDate
--- a/client/src/elm/Measurement/Utils.elm
+++ b/client/src/elm/Measurement/Utils.elm
@@ -1558,22 +1558,20 @@ vaccinationFormDynamicContentAndTasks language currentDate site config vaccineTy
                                 -- expected due date of first dose.
                                 |> Maybe.withDefault config.firstDoseExpectedFrom
 
-                        -- A dose given previously was given before today, so the
-                        -- range ends yesterday. On the day a dose first becomes due
-                        -- its due date is later than that, and the range starts from
-                        -- the last day it can offer instead.
+                        -- The range runs from the day the dose became due to
+                        -- yesterday. A dose that becomes due today was not due on any
+                        -- earlier day, so the range is that one day: earlier is before
+                        -- the dose could be given, and for a first dose of BCG or OPV
+                        -- it would be before the child was born.
                         dateTo =
-                            Date.add Days -1 currentDate
-
-                        rangeStart =
-                            Date.min dateFrom dateTo
+                            Date.max (Date.add Days -1 currentDate) dateFrom
 
                         dateSelectorConfig =
                             { select = config.setVaccinationUpdateDateMsg
                             , close = config.setVaccinationUpdateDateSelectorStateMsg Nothing
-                            , dateFrom = rangeStart
+                            , dateFrom = dateFrom
                             , dateTo = dateTo
-                            , dateDefault = Just rangeStart
+                            , dateDefault = Just dateFrom
                             }
                     in
                     ( [ viewLabel language Translate.SelectDate


client/src/elm/Measurement/Utils.elm
```diff
@@ -1558,11 +1558,19 @@ vaccinationFormDynamicContentAndTasks language currentDate site config vaccineTy
                                 -- expected due date of first dose.
                                 |> Maybe.withDefault config.firstDoseExpectedFrom
 
+                        -- The range runs from the day the dose became due to
+                        -- yesterday. A dose that becomes due today was not due on any
+                        -- earlier day, so the range is that one day: earlier is before
+                        -- the dose could be given, and for a first dose of BCG or OPV
+                        -- it would be before the child was born.
+                        dateTo =
+                            Date.max (Date.add Days -1 currentDate) dateFrom
```
anvmn, line 1567: **Confirmed, and it was worse than the defect it replaced. Fixed in `d5c186bb8`.**

`config.firstDoseExpectedFrom` is the birth date for BCG and for OPV dose 1 (`Measurement/Utils.elm:9512-9518`). So for a newborn seen on their date of birth, `dateFrom == currentDate == birthDate`, and clamping the start to yesterday offered exactly one day: **the day before the child was born**. A nurse recording a BCG given at the hospital that morning could only save it dated before birth, and nothing downstream would reject it.

The range now extends its end instead of pulling its start back:

```elm
dateTo =
    Date.max (Date.add Days -1 currentDate) dateFrom
```

Run against justinmimbs/date 4.1.0:

```
newborn, BCG due at birth   shipped: INVERTED
                            clamp  : 2026-08-11..2026-08-11   <- before birth
                            extend : 2026-08-12..2026-08-12
dose due last month         all three identical
dose due yesterday          all three identical
```

This does mean a dose that becomes due today can be recorded as given previously, dated today — which the issue described as part of the defect. The newborn case shows that framing was wrong: a BCG given at the hospital hours earlier is genuinely a previous administration, and today is the only honest date for it. The unambiguous defect was the dead calendar, and that is gone.

anvmn (commented):
Review findings (medium). The first is a regression this branch introduced and is fixed in `d5c186bb8`; the second is answered by the same change.

anvmn:
The second finding — that opening the picker silently commits the default, so a date the nurse never chose reaches the record — is answered by the same change rather than separately.

With the range ending at the due date when a dose becomes due today, the default that gets committed is that due date: the only day the dose could have been given, and the day the nurse would have picked. It is no longer a date placed outside the range by intent.

The commit-on-open behaviour itself is unchanged and stays out of scope. It applies to every vaccine, not to this case, and it is the reason a default reaches the record at all — worth its own look, but not something to change while fixing a date range.

client/src/elm/Measurement/Utils.elm
anvmn, line 1567: **Real, and pre-existing rather than introduced here.**

Both mechanisms check out: `doseGivenToday` (`:1422`) matches any dose in this encounter dated today and pins `nextDose` through `Maybe.Extra.or`, and `doseAllowedForDeletion` (`:1339`) filters out `date == currentDate`, so that entry cannot be removed from the history list.

What neither depends on is this branch. On develop the config reads:

```elm
, dateFrom = dateFrom          -- the due date
, dateTo = Date.add Days -1 currentDate
, dateDefault = Just dateFrom  -- = TODAY when the dose is due today
```

and `SetVaccinationUpdateDateSelectorState` commits `dateDefault` the moment the picker opens, without passing through the widget. So on develop, a nurse who answers "administered previously" and taps the date field has **already** saved a dose dated today through this path — the dead calendar only removes their ability to correct it.

This branch changes which days the calendar offers. It does not change the date that gets committed in the due-today case: it was today before and it is today now.

The real lever is the third finding — `dateDefault` being committed on open. That is worth its own change, and it is what would stop the previously-path from producing a today-dated dose at all. It applies to every vaccine, not to the due-today case, so it does not belong in a fix to a date range.

anvmn (commented):
Second review pass. Both medium findings describe real behaviour, and both already happen on develop — the mechanism is the default, not the range. Detail inline.

PR #2079, opened 2026-08-13
Wellbeing: stop the notification reappearing when it is ignored

### LINKED ISSUE #2078: Wellbeing: the notification reappears immediately after it is ignored
A nurse enrolled in the resilience programme, returning to the main menu with unread messages, gets the notification dialog. Pressing Ignore brings it straight back, and it takes one press per hour since they logged in before it goes away.

The dialog shows while the scheduled time has passed and there are unread messages (`client/src/elm/Pages/PinCode/View.elm:433-441`). Ignore reschedules from the time already stored rather than from now (`Pages/PinCode/Update.elm:44-46`):

```elm
nextNotification =
    Maybe.map (Time.Extra.add Time.Extra.Hour 1 Time.utc)
        model.nextNotification
```

That time was set once, at login (`App/Update.elm:1395`). So after more than an hour, adding an hour to it still lands in the past, the condition still holds, and the dialog renders again.

| hours since login | Ignore presses before it goes away |
| --- | --- |
| 0.5 | 1 |
| 2 | 3 |
| 5 | 6 |
| 8 | 9 |

Nothing is lost or mis-saved; the dialog simply cannot be dismissed in one press.

The resilience programme is enrolled per nurse rather than per site: 20 nurses on ihangane, none on vhw or tip-somalia, with 87 survey records between August 2024 and November 2025 and none in 2026. The mechanism stays armed for any nurse an administrator enrols.

anvmn:
Fixes #2078

Ignoring the notification schedules the next one an hour from now:

```elm
nextNotification =
    Just <| Time.Extra.add Time.Extra.Hour 1 Time.utc currentTime
```

It previously added an hour to the time already stored, which was set at login — so once more than an hour had passed, the new time was still in the past and the dialog showed again immediately, taking one press per hour elapsed to clear.

`Pages.PinCode.Update.update` takes the current time for this, and `App.Update` passes `model.currentTime` when it calls it.

### DIFF
--- a/client/src/elm/App/Update.elm
+++ b/client/src/elm/App/Update.elm
@@ -1054,7 +1054,7 @@ update msg model =
                 (\configured ->
                     let
                         ( subModel, subCmd, outMsg ) =
-                            Pages.PinCode.Update.update subMsg configured.pinCodePage
+                            Pages.PinCode.Update.update model.currentTime subMsg configured.pinCodePage
 
                         ( extraMsgs, extraCmds ) =
                             Maybe.map
--- a/client/src/elm/Pages/PinCode/Update.elm
+++ b/client/src/elm/Pages/PinCode/Update.elm
@@ -6,8 +6,8 @@ import Time
 import Time.Extra
 
 
-update : Msg -> Model -> ( Model, Cmd Msg, Maybe OutMsg )
-update msg model =
+update : Time.Posix -> Msg -> Model -> ( Model, Cmd Msg, Maybe OutMsg )
+update currentTime msg model =
     case msg of
         HandleLoginClicked ->
             ( { model | code = "" }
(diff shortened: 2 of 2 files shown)


anvmn:
Review (medium) found **no defects**. Recording what it established, since some of it is worth having on the record:

- The argument order matches `Time.Extra.add : Interval -> Int -> Zone -> Posix -> Posix` and the existing precedent at `Pages/PinCode/View.elm:386`.
- Update and view schedule and compare against the same clock: every render path feeds the view `model.currentTime`, which is now also what the update schedules from.
- `currentTime` is never a stale zero — `Task.perform Tick Time.now` at init plus `Time.every 10000 Tick` keeps it within ten seconds, irrelevant against an hour-scale threshold.
- `Pages.PinCode.Update.update` has exactly one caller, updated here.
- The one semantic change — `nextNotification` becoming `Just` where the old `Maybe.map` would have left `Nothing` — is unreachable: `HandleNotificationResponse` is emitted only by the dialog's two buttons, which render only when it is already `Just`.

It also noted a pre-existing behaviour left alone here: logging out and back in within the hour re-arms the notification immediately, since `SetLoggedIn` dispatches `SetNextNotification` on both login and logout.

PR #2081, opened 2026-08-13
Patient search: stop the search losing what was typed

### LINKED ISSUE #2080: Patient search: switching the search mode wipes what was typed
Searching patients by National ID and then emptying the field leaves the results of the previous search displayed, with an empty box above them.

`SetInput` for `ModeSearchByNationalId` returns the model untouched when the text does not parse as a number, and the empty string does not parse (`client/src/elm/Components/PatientsSearchForm/Update.elm:70-72`):

```elm
if isJust asNumber then
    ( { model | input = input }, Cmd.none )
        |> sequence update [ MsgDebouncer <| provideInput <| SetSearch input ]

else
    ( model, Cmd.none )
```

So selecting the text and deleting it changes nothing: `model.search` still holds the previous number, and the list still shows what it found. Since the model is unchanged the view is identical, nothing is patched, and the box stays visually empty while the results remain.

The widget is shared by the People, Individual Encounter Participants and Family Encounter Participants pages, and by acute-illness contact tracing.

### LINKED ISSUE #2082: Patient search: a rejected national ID leaves the field showing what the search ignored
In the patient search, tapping "Name" while already searching by name — or "National ID" while already searching by ID — empties the field and discards the query.

The label carries its handler whether or not the option is selected (`client/src/elm/Pages/Utils.elm:598`):

```elm
label [ onClick <| setMsg value ] [ text <| translate language transId ]
```

and `SetMode` resets regardless of which mode it is given (`Components/PatientsSearchForm/Update.elm:31`):

```elm
( { model | mode = mode, input = "", search = Nothing }, Cmd.none )
```

The same handler leaves the debouncer alone. It holds the last input it was given, so a query typed within half a second of a genuine mode switch arrives afterwards and is searched for in the new mode — a name can be looked up as a national ID, or a query that was just cleared can come back.

The widget is shared by the People, Individual Encounter Participants and Family Encounter Participants pages, and by acute-illness contact tracing.

anvmn:
Fixes #2080
Fixes #2082

Three changes to the shared patient-search widget.

**An emptied National ID field searches for nothing**, so `search` is cleared with it and the results of the previous search leave the screen.

**Tapping the label of the mode already in use** asks for that mode again, and what has been typed stays where it is.

**Switching mode gives the debouncer the empty input.** It holds the last input it was given, so a query typed a moment before a switch would otherwise arrive after it and be searched for in the mode it was not typed for. The emptied-field case does the same, so a number typed just before the field was cleared cannot bring its results back half a second later. This is what the search by name already does for every input.

## Not in this PR

A rejected non-digit stays visible in the National ID field: returning the model unchanged leaves the view identical, so nothing is patched and the character the model refused remains on screen.

### DIFF
--- a/client/src/elm/Components/PatientsSearchForm/Update.elm
+++ b/client/src/elm/Components/PatientsSearchForm/Update.elm
@@ -57,11 +57,14 @@ update msg model =
                         |> sequence update [ MsgDebouncer <| provideInput <| SetSearch input ]
 
                 ModeSearchByNationalId ->
-                    let
-                        asNumber =
-                            String.toInt input
-                    in
-                    if isJust asNumber then
+                    if String.isEmpty input then
+                        -- An emptied field searches for nothing, so the results
+                        -- of what it held go with it.
+                        ( { model | input = "", search = Nothing }
+                        , Cmd.none
+                        )
+
+                    else if isJust (String.toInt input) then
                         ( { model | input = input }
                         , Cmd.none
                         )
--- a/client/src/elm/Components/PatientsSearchForm/Update.elm
+++ b/client/src/elm/Components/PatientsSearchForm/Update.elm
@@ -59,10 +59,14 @@ update msg model =
                 ModeSearchByNationalId ->
                     if String.isEmpty input then
                         -- An emptied field searches for nothing, so the results
-                        -- of what it held go with it.
+                        -- of what it held go with it. The empty input goes to
+                        -- the debouncer too: it holds the last input it was
+                        -- given, and a number given a moment earlier would
+                        -- otherwise arrive half a second later and search again.
                         ( { model | input = "", search = Nothing }
                         , Cmd.none
                         )
+                            |> sequence update [ MsgDebouncer <| provideInput <| SetSearch "" ]
 
                     else if isJust (String.toInt input) then
                         ( { model | input = input }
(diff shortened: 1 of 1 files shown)


client/src/elm/Components/PatientsSearchForm/Update.elm
```diff
@@ -57,11 +57,18 @@ update msg model =
                         |> sequence update [ MsgDebouncer <| provideInput <| SetSearch input ]
 
                 ModeSearchByNationalId ->
-                    let
-                        asNumber =
-                            String.toInt input
-                    in
-                    if isJust asNumber then
+                    if String.isEmpty input then
```
anvmn: **Confirmed and fixed in `90a10aeb8`.** Clearing `search` was not enough, because the debouncer still held the previous keystroke.

Verified at each hop: `debounce 500` is `settleWhenQuietFor (Just 500) manual` with the `lastInput` accumulator (Gizra/elm-debouncer 2.0.0), so whatever was provided last still emits after the quiet period. Emptying the field returned without providing anything, so the number from the keystroke before arrived half a second later, set `search` again, and brought the results back under an empty field — the symptom this PR exists to remove, delayed by 500ms.

The empty input now goes to the debouncer as well, which is exactly what the search by name already does for every input including the empty one. `SetSearch ""` trims to empty and resolves to `Nothing`, so the late emit now clears rather than restores.

I used `provideInput` rather than `Debouncer.cancel` to match the name branch; both stop the resurrection, and matching the sibling keeps one behaviour in the component rather than two.

anvmn (commented):
Review finding taken — the fix was half a fix. Detail inline.

PR #2088, opened 2026-08-16
Child scoreboard: stop one undated ANC visit from blanking a district's scoreboard

### LINKED ISSUE #2087: Child scoreboard: one ANC visit with no scheduled date blanks the scoreboard for a whole district
`hedley_ncda_calculate_aggregated_data_for_person()` collects the dates of a pregnancy's ANC visits into the person's stored scoreboard data:

```php
$data['ncda']['pane1']['row1'][] = hedley_ncda_get_date_filed_value($encounter, 'field_scheduled_date');
```

`hedley_ncda_get_date_filed_value()` returns `FALSE` when the field is empty, so an encounter with no scheduled date contributes `false` to a list that holds date strings, and `json_encode()` writes it into `field_ncda_data`.

The aggregated scoreboard decodes that list with `decodeMonthlyValues`, which is `list decodeYYYYMMDD`. A `false` element fails that decoder, which fails the whole `PatientData`, which fails the entire `decodeSyncResponse` batch it travels in. The scoreboard then never loads for anyone in that district — one person's record takes down the report for all of them.

Every other date this function collects is guarded and skipped when absent, and the EDD read a few lines above the ANC loop checks `!empty` before storing. This one call is the exception.

anvmn:
Fixes #2087

An ANC encounter with no scheduled date is skipped rather than contributing `false` to the person's stored list of visit dates, so the list holds only dates and the district's aggregated scoreboard decodes.

An encounter with no type is read without raising a notice. It still counts as an ANC visit, which is how the client reads it too: the encounter syncs with a null type, and the client's decoder falls back to a nurse encounter.

Both skips follow the form the other dates collected in this file already use.

No repair accompanies this change: no stored record holds a `false` today, on any site. A record written before this change would keep it until that person's data is recalculated, which happens only when a measurement save, a new village, or a person edit triggers it — not on a schedule.

### DIFF
--- a/server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
+++ b/server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
@@ -699,7 +699,14 @@ function hedley_ncda_calculate_aggregated_data_for_person($person, $birth_date_f
     if (!empty($encounters_ids)) {
       $encounters = node_load_multiple($encounters_ids);
       foreach ($encounters as $encounter) {
-        $encounter_type = $encounter->field_prenatal_encounter_type[LANGUAGE_NONE][0]['value'];
+        // Encounters recorded before the type field was populated have no
+        // type. They count as ANC visits, which is how the client reads them
+        // too.
+        $encounter_type = NULL;
+        if (!empty($encounter->field_prenatal_encounter_type[LANGUAGE_NONE][0]['value'])) {
+          $encounter_type = $encounter->field_prenatal_encounter_type[LANGUAGE_NONE][0]['value'];
+        }
+
         if (in_array($encounter_type, ['nurse-postpartum', 'chw-postpartum'])) {
           continue;
         }
--- a/server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
+++ b/server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
@@ -704,7 +704,12 @@ function hedley_ncda_calculate_aggregated_data_for_person($person, $birth_date_f
           continue;
         }
 
-        $data['ncda']['pane1']['row1'][] = hedley_ncda_get_date_filed_value($encounter, 'field_scheduled_date');
+        $scheduled_date = hedley_ncda_get_date_filed_value($encounter, 'field_scheduled_date');
+        if (!$scheduled_date) {
+          continue;
+        }
+
+        $data['ncda']['pane1']['row1'][] = $scheduled_date;
       }
     }
   }


server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
```diff
@@ -704,7 +704,12 @@ function hedley_ncda_calculate_aggregated_data_for_person($person, $birth_date_f
           continue;
         }
 
-        $data['ncda']['pane1']['row1'][] = hedley_ncda_get_date_filed_value($encounter, 'field_scheduled_date');
+        $scheduled_date = hedley_ncda_get_date_filed_value($encounter, 'field_scheduled_date');
```
anvmn, line 714: **Review finding (medium) — confirmed, and the PR description is wrong about it.**

The guard fixes the producer, but nothing re-runs it for people whose stored blob already holds a `false`. Recalculation is only triggered by a measurement save, a new village, a created person, or an edited person (`hedley_ncda_trigger_recalculation_*`). A child with a poisoned blob and no further activity keeps it indefinitely, so the district scoreboard stays blank after deploy.

`hedley_ncda` has **no `.install` file at all**, so there is no existing place for an update hook — adding one means creating that file.

The description's claim that records "are corrected when that person's data is next recalculated" is true but misleading, because "next" may be never. That needs correcting either way.

Whether a backfill is warranted depends on whether any poisoned rows exist, which is one query:

```
SELECT COUNT(*) FROM field_data_field_ncda_data
WHERE field_ncda_data_value LIKE '%[false%' OR field_ncda_data_value LIKE '%,false%'
```

(Those patterns match array elements only; a legitimate boolean row renders as `"row1":false`.) The instrument for a repair already exists — `hedley_ncda/scripts/generate-data-for-all.php` — though note B-099: that script's keyset pagination lacks `propertyOrderBy('nid')` and can silently skip children.

anvmn:
**Review finding (low) — confirmed, and stronger than reported.** `hedley_ncda.module:702`, two lines above this hunk and outside the diff, so noted here.

```php
$encounter_type = $encounter->field_prenatal_encounter_type[LANGUAGE_NONE][0]['value'];
```

The review called this the same "missing field" class as the bug being fixed. It is, and the field instance settles it: `field_prenatal_encounter_type` on `prenatal_encounter` is **`required => 0`** — genuinely optional — whereas `field_scheduled_date`, the field this PR guards, is `required => 1`. So the unguarded dereference is on the *less* protected of the two.

An encounter with no type raises a notice, yields `NULL`, fails the postpartum test, and is counted as a regular ANC visit.

Not taken here, because the fix is a product question rather than a mechanical one: skipping an untyped encounter and counting it as ANC are both defensible, and they produce different numbers on the scoreboard. Silencing only the notice while keeping today's counts is a third option. Worth its own decision rather than riding along in a guard PR.

server/hedley/modules/custom/hedley_ncda/hedley_ncda.module
anvmn: Checked on live. **No poisoned records exist, on any site — no backfill is needed.**

- ihangane: **0** rows with a `false` inside an array, out of 121,638 stored blobs
- vhw and tip-somalia: the table is empty (NCDA is not enabled there)

The zero discriminates rather than merely being absent: 118,384 of those ihangane blobs contain the literal `false` (the legitimate boolean rows) and 98,093 contain date arrays, so both ingredients are abundant in the data and the patterns match what they should. Zero arrays contain a `false`.

So the guard is purely preventive, and `generate-data-for-all.php` is not needed here — which also sidesteps B-099's skip-children defect in that script.

The misleading sentence is out of the description: it no longer implies self-healing, and states that recalculation is event-driven rather than scheduled.

anvmn:
Taken in `78b8e2c70`, narrowed to the mechanical half after tracing what the client does with an untyped encounter.

**The classification question turned out to be already answered.** The encounter still syncs — `HedleyRestfulIndividualEncounter` joins the field with `$require = FALSE`, and `hedley_general_join_field_to_query()` maps that to `leftJoin`, so the row survives with a null type rather than being dropped. The client then reads it as:

```elm
|> required "prenatal_encounter_type" (decodeWithFallback NurseEncounter decodePrenatalEncounterType)
```

`required` is satisfied because the key is present (null, not missing), and `decodeWithFallback` is `oneOf [ decoder, succeed fallback ]`, so a null falls through to `NurseEncounter`. The client deliberately treats an untyped encounter as a nurse ANC encounter — the same classification this file was reaching accidentally. So the guard preserves the count and only stops the notice.

Worth recording: that LEFT join is what keeps this benign. Were the key omitted rather than nulled, `required` would fail and the encounter would become a poison record that fails its whole download batch.

**It is not hypothetical.** On ihangane.live, 7,654 of 68,876 prenatal encounters have no type — all published, all created in 2020 (2,459) and 2021 (5,195), none since. A closed legacy cohort, still synced and still reaching this line whenever one of those people's data is recalculated.

`??` was not used, despite being the shorter form: it appears nowhere in the custom modules, and the `!empty()` shape matches how this file reads every other single-value field.

PR #2090, opened 2026-08-16
Dashboard stats: stop stale month figures being read as this month's

### LINKED ISSUE #2089: Dashboard stats: last year's figures are read as this year's when the statistics are a month old
The dashboard's monthly figures are numbered by the month they were computed in, and the dashboard places them by the month being viewed. When those are different months, whole columns are attributed to the wrong month.

`hedley_stats_get_case_management_for_period()` buckets each measurement under `date('n', ...)` — a bare month number 1-12, plus 13 for anything older than a year, decided when the job ran. The client then splits that year at `Date.month currentDate`, the month the page is open (`Pages/Dashboard/Utils.elm`, `Pages/Dashboard/View.elm`).

So statistics computed in July hold August-December of last year in buckets 8-12 and January-July of this year in buckets 1-7. Opened in August, bucket 8 — last August — is presented as this August. The case management table's "three previous months" is selected the same way, so it shows three months that are not the last three of the data.

The cached sections never expire: `cache_set()` is called with no expiry and the entries are replaced only when new content triggers a recalculation. A health center with no recent activity keeps its figures, and its dashboard, indefinitely.

On ihangane, 12 of 22 case management cache entries and 6 of 11 children beneficiaries entries were computed in the previous month, the oldest 40 days ago.

The "Last updated" line does not reveal any of this. It reports `time()` at the moment the client fetched the statistics rather than the moment they were computed, so it always reads as current. The moment they were computed is recorded — `HEDLEY_STATS_CACHE_SET` stores a timestamp — but `HEDLEY_STATS_CACHE_GET` returns only the value and nothing reads that timestamp.

anvmn:
Fixes #2089

Statistics carry the date they were computed on, and the figures that are numbered by month are looked up by its month: the beneficiary counts and their labels, the new-case ordering, the malnourished cards on the FBF statistics page, the case management table's three months, and the two nutrition charts. A health center whose statistics were computed in an earlier month reads those against that month, so a figure and its label describe the same month.

Records carrying their own dates keep being filtered by today. Filtering those from the date of computation would hide the records made after it, and the sections they come from have their own caches and can be newer.

The date is stored alongside the statistics, so it survives the app being restarted, and it is read in the same local time the month buckets are numbered in.

`HEDLEY_STATS_CACHE_GET_TIMESTAMP` reads the generation time that `HEDLEY_STATS_CACHE_SET` already records.

Statistics without the date are read against today, as they were before it was sent, so a client and a server on different versions behave as they do today. The month buckets keep their existing numbering, which clients parse into integers: renumbering them would silently collapse every month into one on any client that had not updated.

Where a page shows both — the FBF beneficiaries table divides bucketed figures by a count of dated records — the two describe different months while a cache is a month old, and no choice of month reconciles them.

"Last updated" is unchanged. It reports when the statistics were served, and the sections behind it are refreshed independently, so no single moment describes them all.

### DIFF
--- a/client/src/elm/Pages/Dashboard/Utils.elm
+++ b/client/src/elm/Pages/Dashboard/Utils.elm
@@ -1232,7 +1232,7 @@ generateNutritionPageData currentDate stats programTypeFilter selectedVillageFil
     , totalEncounters = generateTotalEncounters currentPeriodStats.totalEncounters programTypeFilter selectedVillageFilter
     , totalsGraphData = totalsGraphData
     , newCasesGraphData = newCasesGraphData
-    , statsMonth = statsMonth
+    , statsGeneratedDate = stats.statsGeneratedDate
     }
 
 
--- a/client/src/elm/Pages/Dashboard/Utils.elm
+++ b/client/src/elm/Pages/Dashboard/Utils.elm
@@ -1,4 +1,4 @@
-module Pages.Dashboard.Utils exposing (applyGenderFilter, caseManagementMergeDuplicates, countAcuteIllnessAssessments, countAcuteIllnessCasesByPossibleDiagnosises, countAcuteIllnessCasesByTreatmentApproach, countAcuteIllnessDiagnosedCases, countComplicatedGISentToHC, countComplicatedMalariaSentToHC, countCurrentlyPregnantForSelectedMonth, countCurrentlyPregnantWithDangerSignsForSelectedMonth, countDeliveriesAtLocationForSelectedMonth, countDiagnosedWithCovidCallsTo114, countDiagnosedWithCovidManagedAtHome, countDiagnosedWithCovidSentToHC, countDiagnosedWithGI, countDiagnosedWithMalaria, countHospitalReferralsForSelectedMonth, countNewbornForSelectedMonth, countNewlyIdentifieHypertensionCasesForSelectedMonth, countNewlyIdentifiedDiabetesCasesForSelectedMonth, countNewlyIdentifiedPregananciesForSelectedMonth, countPregnanciesDueWithin4MonthsForSelectedMonth, countPregnanciesWith4VisitsOrMoreForSelectedMonth, countResolvedGICasesForSelectedMonth, countResolvedMalariaCasesForSelectedMonth, countTotalNumberOfPatientsWithDiabetes, countTotalNumberOfPatientsWithGestationalDiabetes, countTotalNumberOfPatientsWithHypertension, countUncomplicatedGIManagedByChw, countUncomplicatedMalariaAndPregnantSentToHC, countUncomplicatedMalariaManagedByChw, countUncomplicatedMalariaSentToHC, filterNewlyDiagnosesCasesForSelectedMonth, filterNewlyDiagnosesMalnutritionForSelectedMonth, filterProgramTypeFromString, filterProgramTypeToString, filterStatsByGender, filterStatsWithinPeriod, generateAssembledData, generatePatientsWithHIV, generateVaccinationProgressDict, getAcuteIllnessFollowUpsBreakdownByDiagnosis, getEncountersForSelectedMonth, getFollowUpsTotals, isAcuteIllnessNurseEncounter, isNurseEncounter, resolveStatsMonth, withinOrAfterSelectedMonth, withinOrBeforeSelectedMonth, withinSelectedMonth)
+module Pages.Dashboard.Utils exposing (applyGenderFilter, caseManagementMergeDuplicates, countAcuteIllnessAssessments, countAcuteIllnessCasesByPossibleDiagnosises, countAcuteIllnessCasesByTreatmentApproach, countAcuteIllnessDiagnosedCases, countComplicatedGISentToHC, countComplicatedMalariaSentToHC, countCurrentlyPregnantForSelectedMonth, countCurrentlyPregnantWithDangerSignsForSelectedMonth, countDeliveriesAtLocationForSelectedMonth, countDiagnosedWithCovidCallsTo114, countDiagnosedWithCovidManagedAtHome, countDiagnosedWithCovidSentToHC, countDiagnosedWithGI, countDiagnosedWithMalaria, countHospitalReferralsForSelectedMonth, countNewbornForSelectedMonth, countNewlyIdentifieHypertensionCasesForSelectedMonth, countNewlyIdentifiedDiabetesCasesForSelectedMonth, countNewlyIdentifiedPregananciesForSelectedMonth, countPregnanciesDueWithin4MonthsForSelectedMonth, countPregnanciesWith4VisitsOrMoreForSelectedMonth, countResolvedGICasesForSelectedMonth, countResolvedMalariaCasesForSelectedMonth, countTotalNumberOfPatientsWithDiabetes, countTotalNumberOfPatientsWithGestationalDiabetes, countTotalNumberOfPatientsWithHypertension, countUncomplicatedGIManagedByChw, countUncomplicatedMalariaAndPregnantSentToHC, countUncomplicatedMalariaManagedByChw, countUncomplicatedMalariaSentToHC, filterNewlyDiagnosesCasesForSelectedMonth, filterNewlyDiagnosesMalnutritionForSelectedMonth, filterProgramTypeFromString, filterProgramTypeToString, filterStatsByGender, filterStatsWithinPeriod, generateAssembledData, generatePatientsWithHIV, generateVaccinationProgressDict, getAcuteIllnessFollowUpsBreakdownByDiagnosis, getEncountersForSelectedMonth, getFollowUpsTotals, isAcuteIllnessNurseEncounter, isNurseEncounter, resolveStatsDate, withinOrAfterSelectedMonth, withinOrBeforeSelectedMonth, withinSelectedMonth)
 
 import AssocList as Dict exposing (Dict)
 import Backend.AcuteIllnessEncounter.Types exposing (AcuteIllnessDiagnosis(..), AcuteIllnessEncounterType(..))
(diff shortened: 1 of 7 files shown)


client/src/elm/Pages/Dashboard/Utils.elm
```diff
@@ -1226,18 +1230,29 @@ generateNutritionPageData currentDate stats programTypeFilter selectedVillageFil
...
 generateTotalBeneficiariesMonthlyDuringPastYear :
     NominalDate
     -> DashboardStats
     -> Dict Int Int
 generateTotalBeneficiariesMonthlyDuringPastYear currentDate stats =
     let
         currentMonth =
-            Date.month currentDate
-                |> Date.monthToNumber
+            resolveStatsMonth currentDate stats.statsGeneratedMonth
```
anvmn: **Confirmed — high, and this makes the function worse than before the change.** The month *labels* now come from the stats month, but the values below still walk back from `currentDate`: `Date.add Months (-1 * index) currentDate`. Reversed index 0 is labelled with the stats month while holding the count as of today.

Before the change both label and value derived from `currentDate` and agreed. This dict is also the denominator in `applyTotalBeneficiariesDenomination` for both graphs, so a stats-anchored numerator would be divided by a today-anchored denominator.

Fixing it needs the generation **date**, not just its month number — walking months back from `currentDate` lands on the wrong months as soon as the two differ.

client/src/elm/Pages/Dashboard/View.elm
```diff
@@ -331,8 +331,7 @@ viewStatsPage language currentDate stats model =
     else
         let
             currentMonth =
-                Date.month currentDate
-                    |> Date.monthToNumber
+                resolveStatsMonth currentDate stats.statsGeneratedMonth
```
anvmn: **Confirmed — medium.** `displayedMonth` now comes from the stats month, but `currentPeriodStats` and `monthBeforeStats` are still `filterStatsWithinPeriod currentDate model.period stats`. The malnourished cards would report one month while the misc cards and the beneficiaries table report another, inside a single "This month" row.

Same remedy: an anchor date that both the month selection and the period filter use.

server/hedley/modules/custom/hedley_stats/hedley_stats.module
```diff
@@ -831,10 +832,19 @@ function hedley_stats_pull_stats_for_health_center($health_center_id, $health_ce
...
+  // management entry dates all of them.
+  $generated = hedley_stats_handle_cache(HEDLEY_STATS_CACHE_GET_TIMESTAMP, HEDLEY_STATS_SYNC_CASE_MANAGEMENT, $health_center_id, HEDLEY_STATS_PERIOD_ONE_YEAR);
+  if (empty($generated)) {
+    $generated = time();
+  }
+  $generated_utc = $generated - date('Z');
   // UTC date and time on which statistics were generated.
-  $timestamp = date('d-m-Y, H:i', $now - date('Z'));
-  $stats['timestamp'] = $timestamp;
+  $stats['timestamp'] = date('d-m-Y, H:i', $generated_utc);
```
anvmn: **Confirmed — medium (finding 4) and low/medium (finding 5).**

**4.** Sourcing the displayed timestamp from the case-management cache alone is wrong for the other pages. `hedley_stats_clear_caches_for_health_center()` returns early for prenatal, acute illness, NCD, PMTCT, SPV, child scoreboard and group education (lines 521-549), so those sections refresh without touching the case-management entry. A health center with daily prenatal activity and no nutrition activity would show a months-old "Last updated" beside current figures — a new inaccuracy in place of the old one.

**5.** The month is computed in UTC (`date('n', $generated - date('Z'))`) while the buckets it indexes are numbered in server-local time (`date('n', $date_as_timestamp)`). Around a month boundary those differ by one, which rotates the whole year split. In D7 `date('Z')` also reflects the current request's timezone, so the same cached stats could yield different months for different users.

anvmn:
**Finding 1 — confirmed, high. `Backend/Dashboard/Encoder.elm:60`, outside the diff, so noted here.**

`encodeDashboardStatsRaw` emits `timestamp` and `stats_cache_hash` but not the new field, and it is on the persistence path: `SyncManager/Utils.elm:1384` encodes downloaded stats with it before they are written to IndexedDB. On the next app start they are read back through `decodeDashboardStatsRaw`, where the field is `optional … Nothing`.

So the generation month survives only in the tab that downloaded the stats. After any reload it is gone and `resolveStatsMonth` falls back to today — the bug this PR exists to fix, restored. And since stats are refetched only when `stats_cache_hash` changes, a health center whose statistics are not recomputed would never get the field back.

This is the same shape as the photo `fileId` encoder rule already recorded for this codebase: a new field on a synced entity has to be added to the encoder, not only the decoder.

---

All five findings hold. Two of them mean this branch does not deliver its fix, so it should not merge as it stands. Reworking as:

- ship a generation **date** (`YYYY-MM-DD`) rather than a month number, in server-local time to match how the buckets are numbered — which also settles finding 5 by construction;
- add it to the encoder;
- anchor the beneficiaries date walk and `filterStatsWithinPeriod` on it, so labels, values and period filters all move together;
- drop the "Last updated" change and file it separately, since one timestamp cannot honestly describe seven independently-refreshed sections.

anvmn:
All five findings addressed in `9178c3ebf`.

**1 — encoder.** `stats_generated_date` is emitted by `encodeDashboardStatsRaw`, so it survives being stored and read back. Covered by a test folded into `Pages/Dashboard/Test.elm`, which encodes and decodes the record and asserts the date comes back. **The test was run against the defective encoder first**: with the line removed it fails with `Ok Nothing`, which is exactly the silent loss described. Compilation and the rest of the suite stayed green throughout, which is why nothing caught this before.

**2 and 3 — label, value and period now share one anchor.** The field is a date rather than a month number, which is what the month walk needs: `Date.add Months (-1 * index) statsDate` for the beneficiary counts, the numbering derived from the same date, and `filterStatsWithinPeriod statsDate` in `viewStatsPage`. Walking back from `currentDate` while numbering from the stats month was the mismatch.

**4 — "Last updated" reverted.** Tying it to the case-management entry would have been a new inaccuracy for the pages whose sections refresh independently, so it reports serve time as before. The label is worth fixing, but it needs a per-section answer rather than one timestamp, so it belongs in its own change.

**5 — local time.** The date is now `date('Y-m-d', $generated)` with no UTC adjustment, matching `date('n', $date_as_timestamp)` in the bucket numbering. That also removes the per-request `date('Z')` variation.

Verified on the committed state in a clean clone: 548 modules, 2,975 tests, elm-format, elm-review, `php -l`, phpcs Drupal and DrupalPractice.

anvmn:
All three addressed in `855e3583a`, two of them by narrowing rather than by patching.

**A and B together — the FBF statistics page is out of this change.** `viewStatsPage` is back to `currentDate` for both the month selection and the period filters, so `viewBeneficiariesTable` and its age filters are consistent again and no record is dropped. Finding B is what settled it: `filterStatsWithinPeriod` ends its window at `Date.add Days 1 <anchor>`, so anchoring it on the generation date silently hides family-planning and session-attendance records recorded after that date. Those sections have their own caches and can be fresher. Losing records is worse than the month a card is labelled with, and the labelling mismatch on that page predates this branch.

The net diff to `View.elm` is now the import, `viewCaseManagementPage`, and the two charts. `viewStatsPage` is untouched.

**C — the charts take their month at render.** `NutritionPageData` carries `statsGeneratedDate : Maybe NominalDate` rather than a resolved month, and `chartsMonth currentDate data` resolves it in the view. Statistics with no generation date follow the clock as they did before, so a device left open across a month boundary rolls over.

What remains anchored on the generation date is only the figures that are month-numbered buckets: the beneficiaries denominator with its label and value moving together, the new-cases ordering, the case management table's three-month filter, and the two charts.

Worth stating plainly, because no anchoring can fix it: when statistics are a month stale, the FBF "This month" cards cannot be right, since the current month's figures are not in the cache at all. That is a freshness-reporting problem, and the "Last updated" label that would make it visible reports serve time. Filed separately.

Verified on the committed state in a clean clone: 548 modules, 2,975 tests, elm-format, elm-review, `php -l`, phpcs Drupal and DrupalPractice.

anvmn:
Both Copilot findings were right.

**The description was stale, and that is the more useful catch.** It still said the period filters move with the month numbering. They did in an earlier revision, and were reverted in `855e3583a` because `filterStatsWithinPeriod` ends its window at `Date.add Days 1 <anchor>` — anchoring it on the generation date hides family-planning and session-attendance records recorded after that date, and those sections have their own caches and can be fresher. Losing records is worse than the month a card is labelled with. The description now says what the branch does, and says the FBF page is deliberately unchanged.

**The timestamp guard is taken in `c26357835`.** `isset()` rather than `!empty($cached)`, so a cache entry without the key answers "no date" instead of raising a notice. Worth noting it is unreachable today: `HEDLEY_STATS_CACHE_SET` has written that key since 2022, and the call site already falls back when the value is empty. The guard removes the notice path rather than a wrong result.

anvmn:
The suppressed comment was right, and it is taken in `d40292a4a`.

`mapMalnorishedByMonth` compares `month /= mappedMonth` against `caseNutrition.nutrition.stunting`, which is keyed by the month the statistics were computed in — a bucket lookup, not a date filter. Indexed by the month being viewed, a July cache opened in August returns bucket 8, last August, and presents it as this month.

**The description was wrong too.** It said that page's figures come from dated records rather than month buckets. That is true of `filterStatsWithinPeriod` and false of the malnourished cards: the page uses both, and the description claimed only one. Corrected.

Reverting the page wholesale in an earlier revision was the mistake. The two mechanisms need opposite treatment, and I applied one decision to both:

- bucket lookups are read against the month of computation, or they name the wrong month;
- records carrying their own dates stay filtered by today, or the ones recorded after the cache was written disappear.

`viewStatsPage` now does both. `filterStatsWithinPeriod` is untouched, and `currentMonth` is gone, since the month being viewed no longer decides anything there.

One thing no anchoring fixes, now stated in the description: the beneficiaries table divides bucketed figures by a count of dated records, so while a cache is a month old those two describe different months.

PR #2093, opened 2026-08-16
Dashboard stats: stop a second measurement at one encounter from breaking the figures

### LINKED ISSUE #2092: Dashboard stats: a second measurement at one encounter is misread, and can stop the dashboard loading
Several statistics collect an encounter's measurements with `GROUP_CONCAT` and then read the result as if it were a single value. An encounter can hold more than one of each, and then the reads are wrong in two different ways.

**A fever goes uncounted.** `hedley_stats_get_acute_illness_data()`:

```php
$fever_by_signs = !empty($row->fever_period) && $row->fever_period > 0;
$fever_by_vitals = !empty($row->body_temperature) && (float) $row->body_temperature >= 37.5;
```

`(float) "37.2,38.5"` is `37.2`, so a fever recorded in a later reading is not counted, and `"0,5" > 0` is `false`, so a fever period recorded after one of zero is not counted either. Both feed the COVID-suspect and complicated-infection figures.

**The dashboard stops loading.** Three fields are passed on as the concatenated list itself: the prenatal MUAC, and the HIV test result and execution note on an NCD encounter. The dashboard decodes those as a float and as fixed sets of values, and a value it cannot parse fails the whole batch it travels in — so the dashboard would stop loading for everyone at that health center. Every other single measurement in the file goes through `hedley_stats_extract_first_value()`; these three do not.

Neither has fired yet. Across all three sites no prenatal encounter holds two different MUAC values and no NCD encounter holds two HIV tests, and the two acute illness encounters that hold more than one temperature hold the same value repeated, which the cast happens to read correctly.

anvmn:
Fixes #2092

The statistics ask for the highest temperature and the longest fever period recorded at an acute illness encounter, so a fever counts wherever in the encounter it was recorded.

The prenatal MUAC, and the HIV test result and execution note on an NCD encounter, are read through `hedley_stats_extract_first_value()`, the same way every other single measurement in this file already is, so the dashboard receives one value rather than a list. A list fails the dashboard's decoder and takes the whole health center's dashboard with it.

Every `GROUP_CONCAT` in the file is now either read as a list by its consumer or reduced to one value before it is sent.

The statistics test records two of each measurement, so each of these reads is asked which one decides: two temperatures with only the second feverish, a second encounter reporting a fever only in the later of two fever periods and no temperature reaching the threshold, two MUAC readings at a prenatal encounter, and two HIV tests at an NCD encounter. `HedleyWebTestBase` gained a prenatal nutrition and an NCD HIV test entity, which it could not create before, and its acute illness symptoms accept a fever period.

Also here, since it is a one-line convention rather than a change worth its own branch: a title in `CLAUDE.md` now starts with the feature area, so the merge log says where a defect was as well as what it was.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -138,7 +138,7 @@ Avoid code duplication. Extract shared logic into common functions rather than c
 
 The same principle applies to issue and PR text: describe the current state, not how it came to be.
 
-- The title names the defect from the reader's side, not the mechanism chosen to fix it. Someone scanning the merge log should learn what was wrong. "Stop a person edit from erasing recorded GPS coordinates", not "Carry stored coordinates in PatchPerson".
+- The title starts with the feature area, then names the defect from the reader's side, not the mechanism chosen to fix it: `<Feature area>: <what was wrong>`. Someone scanning the merge log should learn both where the problem was and what it was. "Dashboard stats: stop stale month figures being read as this month's", not "Anchor the month split on stats.timestamp", and not "Stop stale month figures being read as this month's".
 - An issue describes what is wrong and how to see it. A PR describes what the change makes true.
 - **No verification sections.** Do not list how the change was tested, which commands were run, or what they printed. CI reports that, and it is noise in the permanent record.
 - No process narrative — no review rounds, no "the first attempt did X", no account of what was tried and abandoned.
--- a/server/hedley/HedleyWebTestBase.inc
+++ b/server/hedley/HedleyWebTestBase.inc
@@ -538,7 +538,7 @@ class HedleyWebTestBase extends DrupalWebTestCase {
   /**
    * Create an acute illness vitals entity.
    */
-  public function createAcuteIllnessVitals($encounter_id, $person_id, $timestamp = REQUEST_TIME) {
+  public function createAcuteIllnessVitals($encounter_id, $person_id, $timestamp = REQUEST_TIME, $temperature = 37.2) {
     $node = $this->drupalCreateNode(['type' => 'acute_illness_vitals']);
 
     $wrapper = entity_metadata_wrapper('node', $node);
@@ -549,7 +549,7 @@ class HedleyWebTestBase extends DrupalWebTestCase {
     $wrapper->field_dia->set(70);
     $wrapper->field_heart_rate->set(80);
     $wrapper->field_respiratory_rate->set(20);
-    $wrapper->field_body_temperature->set(37.2);
+    $wrapper->field_body_temperature->set($temperature);
     $wrapper->save();
 
     return $node->nid;
--- a/server/hedley/HedleyWebTestBase.inc
+++ b/server/hedley/HedleyWebTestBase.inc
@@ -497,7 +497,7 @@ class HedleyWebTestBase extends DrupalWebTestCase {
   /**
    * Create an acute illness symptom measurement entity of given type.
    */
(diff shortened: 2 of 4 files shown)


anvmn:
Both findings verified. The medium one is taken in `8aef20746`.

**HIV test result and execution note — taken.** Confirmed identical to the prenatal MUAC case: `optional "hiv_test_result" (nullable decodeTestResult) Nothing` with a strict string-to-union decoder, fed a raw `GROUP_CONCAT`. Two `ncd_hiv_test` measurements on one encounter would fail the batch and stop the NCD dashboard for the whole health center. Both now go through `hedley_stats_extract_first_value()`.

You were right that leaving them would have left the class half-closed, and the PR is named for that class. I re-swept every `GROUP_CONCAT` alias in the file rather than close it on a spot check: `administration_dates`, `comorbidities`, `contact_114`, `contact_hc`, `danger_signs`, `education_topics`, `encounter_warnings`, `isolation`, `medical_conditions`, `ncd_diagnoses`, `nutrition_signs`, `participating_patients`, `prenatal_diagnoses`, `recommendation_114`, `recommendation_hc`, `send_to_hc` are all `explode`d by their consumers; `stunting`, `underweight`, `wasting`, `zscore_age`, `zscore_length` and the well-child and nutrition MUACs already used the helper. The five this PR touches were the only ones left. Live check: no NCD encounter on any site holds two HIV tests today, so this is latent like the rest.

**MUAC ordering — not taken, and worth stating why.** The point is correct: `GROUP_CONCAT` guarantees no ordering, so "first" is whichever the engine returns. But nine other call sites in this file take the first value the same way, and changing one of them to `MIN()` would make this read differ from its neighbours for no gain that a reader could see. If the non-determinism is worth closing it should be closed for all ten together, with a decision about which reading is clinically right — that is its own change, and it is recorded in the backlog rather than done quietly here.

anvmn:
The coverage gap is real, and larger than it looks — which is why it is not being closed here.

The test covers the aggregate change: two temperatures at one encounter, only the second feverish, asserting the fever is seen. `fever_period` takes the identical `MAX()` change on the same kind of column, so the mechanism is exercised.

The three scalar conversions are a different matter. Testing them needs duplicate measurements, and the suite creates none of them at all: the prenatal fixture creates no MUAC, and there is no NCD HIV test fixture anywhere. Covering them means writing two new measurement helpers in `HedleyWebTestBase`, which is adding coverage for untested areas rather than testing this change. Each of the three is a one-line call to `hedley_stats_extract_first_value()`, the helper nine other reads in the same file already use.

Recorded as its own piece of work rather than dropped.

anvmn:
Taken in full in `640ab44da` — every read this branch changes is now exercised with two measurements behind it.

- **Temperature**: two readings, only the second feverish.
- **Fever period**: a second acute illness encounter where no temperature reaches the threshold and the fever is reported only in the later of two periods, so this path is asserted independently of the temperature.
- **Prenatal MUAC**: two differing readings at one encounter, asserted to arrive as one of them rather than both.
- **HIV test result and execution note**: two tests at one NCD encounter, asserted the same way.

`HedleyWebTestBase` could not create a prenatal nutrition entity or an NCD HIV test at all, so both were added; acute illness symptoms now accept a fever period.

Every one of these fails before the change: `(float) "37.2,38.5"` is `37.2`, `"0,5" > 0` is `false`, and the MUAC and HIV assertions receive `"23.5,24.5"` and `"positive,negative"`.

phpcs reports the same 11 pre-existing findings in `HedleyWebTestBase.inc` before and after, none on the added lines.

PR #2097, opened 2026-08-17
Immunisation: stop a child holding extra doses reading as behind for good

### LINKED ISSUE #2096: Immunisation: a child holding extra doses reads as behind for good
A course of vaccine is treated as over only when the doses recorded reach exactly the last dose the site expects. A record holding more than that falls past the test, so a further dose is worked out from it: the next dose after the last one recorded, due an interval after the date it was given.

That date has already passed, so the child reads as behind. The same thing happens the following month, and every month after, and nothing a health worker can record will clear it. IPV is the sharpest case, since its interval is zero days: the dose that does not exist is due the day the last real one was given.

On ihangane, children whose recorded doses exceed what the site expects: DTP 12, PCV13 11, OPV 10, IPV 3, MR 2 — around 38 children, and the number only grows.

All three places that work this out agree on it:

- `nextVaccinationDataForVaccine` in the client, `client/src/elm/Measurement/Utils.elm`
- the same function in the scoreboard, `server/elm/src/Pages/Scoreboard/Utils.elm`
- `hedley_reports_resolve_expected_immunisations`, `hedley_reports.module`

So the child reads as behind on the encounter page, on the dashboard's immunisation figures, and in the Completion report — where being behind also raises health education and a referral to the health centre.

anvmn:
Fixes #2096

A course of vaccine is over once the doses recorded reach the last dose the site expects. A child holding more than that has no dose left to receive, so none is worked out for them and they no longer read as behind.

A child short of the last dose still has one due, unchanged.

The three places that work this out say the same thing: the client's encounter pages, the scoreboard, and the reports, whose immunisation figures also raise health education and a referral. Saying it in one place alone would put them back out of step with each other.

### DIFF
--- a/client/src/elm/Measurement/Test.elm
+++ b/client/src/elm/Measurement/Test.elm
@@ -36,6 +36,7 @@ import Measurement.Utils
         , liverFunctionResultFormWithDefault
         , ncdaFormWithDefault
         , ncdaMeasurementsOutOfRange
+        , nextVaccinationDataForVaccine
         , outOfRangeAsEntered
         , setNCDAStep
         , showNCDAMeasurementOutOfRange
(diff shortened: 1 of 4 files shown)


anvmn:
Both taken in `3e1b21108`.

**The unknown-immunisation change was mine, and worse than "loudly wrong to silently dropped" sounds.** `hedley_reports_get_last_dose_for_vaccine()` answers `FALSE` for a type it has no case for. `"dose-3" == FALSE` was false, so the old code fell through and went on asking for the dose after the last one recorded. Reading the answer as a number made it `0`, so `N >= 0` held and the immunisation was skipped for every child, with a warning per row.

Ran the three versions side by side rather than reasoning about it:

| recorded | expected | before the branch | as I pushed it | now |
|---|---|---|---|---|
| dose-3 | dose-3 | complete | complete | complete |
| dose-6 | dose-3 | DUE | complete | complete |
| dose-2 | dose-3 | DUE | DUE | DUE |
| dose-2 | `false` | DUE | **complete** | DUE |

The guard is `!empty($last_dose_for_immunisation) && …`, so an immunisation the site says nothing about falls through exactly as it did before, and the comment says so.

**The duplication is the rule I added to `CLAUDE.md` this morning**, and I broke it the same day, four lines apart. `$last_dose_as_number` is now worked out once, above the guard, and both places use it.

Thank you for checking the two `== $last_dose_for_immunisation` sites I left alone — the prenatal ones cap at `dose-5`, the highest dose that exists, so they cannot be exceeded and equality is right there. I had not stated that, and it deserved stating.

anvmn:
Taken in `3c083c04e`. `Measurement/Test.elm` now asks the DTP course what it wants next, three times:

- **a dose still owing** — two doses given, expects the third four weeks after the last;
- **the last dose given** — three of three, expects none;
- **more doses than the site expects** — four recorded, expects none.

The third fails against the old comparison. Reverting the guard to equality and running the suite: `✗ no dose is due when more doses are recorded than the three expected`, 2,976 passed, 1 failed. Restored, all 2,977 pass. Without that check the test would only have asserted today's behaviour.

DTP was the vaccine to use: it expects three doses at four-week intervals and takes neither of the special paths in that function — OPV's six-week rule for the second dose, and Rwanda's nine-month rule for the second IPV — so the expected dates are the plain interval and the test says what it means.

Also in this branch since the review: the condition now names what it compares, rather than calling `vaccineDoseToComparable` twice in one line.

```elm
dosesGiven =
    vaccineDoseToComparable lastDoseAdministered

dosesTheSiteExpects =
    vaccineDoseToComparable (getLastDoseForVaccine site initialOpvAdministered vaccineType)
in
if dosesGiven >= dosesTheSiteExpects then
```

